PR #2, opened 2019-07-03
Improve incorrect results representation

alefedor:
Now instead of
> get(1)
> get(2)
>
> 3
> 4

we get
> get(1): 3
> get(2): 4

upon incorrect result

### DIFF
--- a/lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
+++ b/lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
@@ -26,7 +26,6 @@ import org.jetbrains.kotlinx.lincheck.execution.ExecutionResult
 import org.jetbrains.kotlinx.lincheck.execution.ExecutionScenario
 import java.io.PrintStream
 
-
 class Reporter @JvmOverloads constructor(val logLevel: LoggingLevel, val out: PrintStream = System.out) {
     fun logIteration(iteration: Int, maxIterations: Int, scenario: ExecutionScenario) = synchronized(this) {
         if (logLevel > LoggingLevel.INFO) return
@@ -46,16 +45,13 @@ class Reporter @JvmOverloads constructor(val logLevel: LoggingLevel, val out: Pr
 
     fun logIncorrectResults(scenario: ExecutionScenario, results: ExecutionResult) = synchronized(this) {
         out.println("= Invalid execution results: =")
-        if (logLevel > LoggingLevel.INFO) { // scenario was not logged before
-            logExecutionScenario(scenario)
-            out.println()
-        }
-        out.println("Execution results (init part):")
-        out.println(results.initResults)
-        out.println("Execution results (parallel part):")
-        out.println(printInColumns(results.parallelResults))
-        out.println("Execution results (post part):")
-        out.println(results.postResults)
+        out.println("Init part:")
+        out.println(uniteActorsAndResults(scenario.initExecution, results.initResults))
+        out.println("Parallel part:")
+        val parallelExecutionData = uniteParallelActorsAndResults(scenario.parallelExecution, results.parallelResults)
+        out.println(printInColumns(parallelExecutionData))
+        out.println("Post part:")
+        out.println(uniteActorsAndResults(scenario.postExecution, results.postResults))
     }
 
     inline fun log(logLevel: LoggingLevel, crossinline msg: () -> String) {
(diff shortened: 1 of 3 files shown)


ndkoval:
When we allow using exceptions, output looks like this:

```
Execution results (parallel part):
| put(1, 4): exception=java.lang.NullPointerException | get(4): null                                        |
| get(3): 0                                           | put(3, 6): exception=java.lang.NullPointerException |
| get(4): null                                        | put(1, 1): 4                                        |
| put(1, 4): 1                                        | get(3): null                                        |
| get(3): 0                                           | put(3, 0): exception=java.lang.NullPointerException |
```

Can we make it less verbose? E.g., using simple names for exceptions and replacing `exception=` with `ex=`?

ndkoval:
It would also be cool to align results, see below:

```
Execution results (parallel part):
| put(1, 4): ex=NullPointerException | get(4):    null                    |
| get(3):    0                       | put(3, 6): ex=NullPointerException |
| get(4):    null                    | put(1, 1): 4                       |
| put(1, 4): 1                       | get(3):    null                    |
| get(3):    0                       | put(3, 0): ex=NullPointerException |
```

ndkoval:
Let's also replace "Execution results (parallel part):" with just "Parallel part:" since we combine both scenario and results.

ndkoval:
Please, update the README as well

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +84,46 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
...
+                "(${actors.size} != ${results.size})")
+
+    return actors.mapIndexed { id, actor -> ActorWithResult(actor, results[id]) }
+}
+
+private fun uniteParallelActorsAndResults(
+        actors: List<List<Actor>>,
+        results: List<List<Result>>
+): List<List<ActorWithResult>> {
+    if (actors.size != results.size) // fail fast
```
ndkoval: Use `check(..)` here

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +84,46 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
             .map { rowCells -> rowCells.joinToString(separator = " | ", prefix = "| ", postfix = " |") }
             .joinToString(separator = "\n")
 }
+
+private class ActorWithResult(private val actor: Actor, private val result: Result) {
```
ndkoval: Let's store actor and result representations as strings + number of spaces

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +82,39 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
             .map { rowCells -> rowCells.joinToString(separator = " | ", prefix = "| ", postfix = " |") }
             .joinToString(separator = "\n")
 }
+
+private class ActorWithResult(val actorRepresentation: String, val resultRepresentation: String) {
+    var spaces = 1
```
ndkoval: could you make it final?

alefedor: @ndkoval, IDE says that 'final' is redundant...

ndkoval: I mean the field

alefedor: I mean the field too.
Do you want to prevent possible future inheritors from overriding the field?
If not, it seems to me that it won't have any difference with `final`

alefedor: Or do you ask to replace `var` with `val`?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +81,34 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
...
+
+private fun uniteActorsAndResults(actors: List<Actor>, results: List<Result>): List<ActorWithResult> {
+    require(actors.size == results.size) {
+        "Different number of actors and matching results found (${actors.size} != ${results.size})"
+    }
+
+    val actorRepresentations = actors.map { it.toString() }
+    val resultRepresentations = results.map { it.toString() }
+
+    val maxActorLength = actorRepresentations.map { it.length }.max()
```
ndkoval: put "!!" here

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +81,34 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
...
             .joinToString(separator = "\n")
 }
+
+private class ActorWithResult(val actorRepresentation: String, val spaces: Int, val resultRepresentation: String) {
+    override fun toString(): String = actorRepresentation + ":" + " ".repeat(spaces) + resultRepresentation
+}
+
+private fun uniteActorsAndResults(actors: List<Actor>, results: List<Result>): List<ActorWithResult> {
+    require(actors.size == results.size) {
+        "Different number of actors and matching results found (${actors.size} != ${results.size})"
```
ndkoval: numbers

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -85,3 +81,34 @@ private fun <T> printInColumns(groupedObjects: List<List<T>>): String {
...
+        ActorWithResult(actorRepr, spaces, resultRepresentations[id])
+    }
+}
+
+private fun uniteParallelActorsAndResults(
+        actors: List<List<Actor>>,
+        results: List<List<Result>>
+): List<List<ActorWithResult>> {
+    require(actors.size == results.size) {
+        "Different number of threads and matching results found (${actors.size} != ${results.size})"
```
ndkoval: numbers

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
alefedor: Then this code won't work if we have zero actors (currently it works).

PR #4, opened 2019-07-04
Spinlock synchronization in parallel threads runner

alefedor:


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.java
```diff
@@ -91,6 +95,8 @@ public ExecutionResult run() throws InterruptedException {
     public void onStart(int iThread) {
         super.onStart(iThread);
         ((TestThread) Thread.currentThread()).iThread = iThread;
+        uninitializedThreads.decrementAndGet();
+        while (uninitializedThreads.get() != 0);
```
ndkoval: add comments here

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/stress/StressStrategy.java
```diff
@@ -60,14 +60,7 @@ public StressStrategy(Class<?> testClass, ExecutionScenario scenario,
             }
         }
         // Create runner
-        Phaser phaser = new Phaser(testCfg.threads);
-        runner = new ParallelThreadsRunner(scenario, this, testClass, null) {
-            @Override
-            public void onStart(int iThread) {
-                super.onStart(iThread);
-                phaser.arriveAndAwaitAdvance();
```
ndkoval: Have you removed phaser completely?

alefedor: That's right

PR #5, opened 2019-07-29
Managed strategies

alefedor:
Strategies with managing of thread switches.
The strategies emulate single core processors where only one thread can be executed at a time.
The switches occur only before interesting events (e.g. shared variables reads/writes or monitor locks).   

The incorrect interleaving can be reported with information about all thread switches.

There are two main managed strategies:
1. RandomSwitchStrategy.
Just switches with a certain probability at every interesting codelocation.
2. RandomSearchStrategy
First runs the current thread until its end or a moment when it can not continue its execution collection information about interesting codelocations and then chooses one where it will switch on next run.


libtest/src/main/java/tests/custom/set/LockFreeSet.kt
```diff
@@ -0,0 +1,106 @@
...
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+package tests.custom.set
+
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicReference
+
+class LockFreeSet {
```
ndkoval: I would name it `OrderedLinkedList` or `LockFreeLinkedListSet`. Guess, it is also incorrect.

libtest/src/main/java/tests/custom/set/LockFreeSet.kt
```diff
@@ -0,0 +1,106 @@
...
+                snapshot.add(node.key)
+        }
+
+        return snapshot
+    }
+
+    private inner class Node(val key: Int, next: Node?, initialMark: Boolean) {
+        // marked if is deleted
+        val next: AtomicReference<Node?> = AtomicReference(next)
+        val isDeleted = AtomicBoolean(initialMark)
```
ndkoval: Is it correct? BTW, it is better to use atomicfu here.

libtest/src/main/java/tests/custom/set/ReentrantLockSet.kt
```diff
@@ -0,0 +1,49 @@
...
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+package tests.custom.set
+
+import java.util.concurrent.locks.ReentrantLock
+
+class ReentrantLockSet {
```
ndkoval: Please, rename it

libtest/src/main/java/tests/custom/set/ReentrantLockSet.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    fun remove(key: Int): Boolean = activeSynchronized {
+        set.remove(key)
+    }
+
+    fun contains(key: Int): Boolean = activeSynchronized {
+        set.contains(key)
+    }
+
+    private fun activeSynchronized(code: () -> Boolean): Boolean {
```
ndkoval: Use `lock.withLock` instead

libtest/src/main/java/tests/custom/set/SuspendMutexSet.kt
```diff
@@ -0,0 +1,42 @@
...
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+package tests.custom.set
+
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+class SuspendMutexSet {
```
ndkoval: For such data structures, it is better to introduce a special mode for verification, which does not take suspensions into account

libtest/src/main/java/tests/custom/set/SynchronizedMethodSet.kt
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Lesser Public License for more details.
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>.
+ * #L%
+ */
+package tests.custom.set
+
+class SynchronizedMethodSet {
```
ndkoval: Why do you need several sets with global locks?

libtest/src/main/java/tests/custom/transfer/AccountsWrong1.java
```diff
@@ -71,4 +72,17 @@ public String toString() {
                 "data=" + data +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
ndkoval: Please, implement `VerifierState` (you can implement for the base `Accounts` class)

libtest/src/main/java/tests/custom/transfer/AccountsWrong2.java
```diff
@@ -71,4 +72,17 @@ public String toString() {
                 "data=" + data +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
ndkoval: Please, implement VerifierState (you can implement for the base Accounts class)

libtest/src/main/java/tests/custom/transfer/AccountsWrong3.java
```diff
@@ -71,4 +72,17 @@ public String toString() {
                 "data=" + data +
                 '}';
     }
+
+    @Override
+    public boolean equals(Object o) {
```
ndkoval: Please, implement VerifierState (you can implement for the base Accounts class)

libtest/src/main/java/tests/custom/transfer/AccountsWrong4.java
```diff
@@ -63,4 +64,17 @@ public String toString() {
                 "data=" + data +
                 '}';
     }
+
+    @Override
```
ndkoval: Please, implement VerifierState (you can implement for the base Accounts class)

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Actor.kt
```diff
@@ -36,8 +35,31 @@ data class Actor(
     val arguments: List<Any?>,
     val handledExceptions: List<Class<out Throwable>>
 ) {
+    val wasInvoked: Boolean
+        get() = wasInvokedInner
+
+    private var wasInvokedInner = false
```
ndkoval: Actors should specify scenarios and are immutable by design. Please, do not break this.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Actor.kt
```diff
@@ -36,8 +35,31 @@ data class Actor(
...
+            throw IllegalStateException("Same actor was invoked twice during the same invocation")
+        }
+
+        wasInvokedInner = true
+    }
+
+    /**
+     * Resets actor
+     */
+    fun reset() {
```
ndkoval: Having `reset` methods is bug-prone.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/CTestConfiguration.java
```diff
@@ -48,6 +50,7 @@
     public static final int DEFAULT_ACTORS_PER_THREAD = 5;
     public static final int DEFAULT_ACTORS_BEFORE = 5;
     public static final int DEFAULT_ACTORS_AFTER = 5;
+    public static final int DEFAULT_MAX_REPETITIONS = 20;
```
ndkoval: What is "repetition"?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/CTestConfiguration.java
```diff
@@ -60,6 +63,7 @@
     public final int actorsAfter;
     public final Class<? extends ExecutionGenerator> generatorClass;
     public final Class<? extends Verifier> verifierClass;
+    public boolean hasTestClassSuspendableActors;
```
ndkoval: This is not a part of configuration (especially if it is not final)

libtest/src/main/java/tests/custom/set/SynchronizedMethodSet.kt
alefedor: To test the support of all ways of locking

libtest/src/main/java/tests/custom/set/ReentrantLockSet.kt
alefedor: How? There are no other differences with similar tests other than the fact that it uses ReentrantLock

libtest/src/main/java/tests/custom/transfer/AccountsWrong1.java
alefedor: Currently libtest main does not depend on lincheck submodule, so either this restriction can be removed or this classes can't implement ```VerifierState```

libtest/src/test/java/org/jetbrains/kotlinx/lincheck/test/AbstractLincheckTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        RandomSearchOptions()
+                .guarantee(if (it) ConcurrentGuarantee.OBSTRUCTION_FREE else ConcurrentGuarantee.NONE)
+                .iterations(20)
+                .invocationsPerIteration(2000)
+    }),
+    RANDOM_SWITCH_OPTIONS({
+        RandomSwitchOptions()
+                .guarantee(if (it) ConcurrentGuarantee.OBSTRUCTION_FREE else ConcurrentGuarantee.NONE)
+                .iterations(20)
+                .invocationsPerIteration(2000)
```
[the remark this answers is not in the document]
alefedor: How do you want to return this information?

lincheck/pom.xml
```diff
@@ -13,15 +13,25 @@
     <dependencies>
         <dependency>
             <groupId>org.ow2.asm</groupId>
-            <artifactId>asm-all</artifactId>
-            <version>5.2</version>
+            <artifactId>asm</artifactId>
```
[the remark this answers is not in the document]
alefedor: Supports all Javas up to Java 13.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/CTestConfiguration.java
alefedor: Visiting the same code location

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Actor.kt
alefedor: I agree, but it still is not easy not to break other invariants and restrictions.
A alternative solution may be to make TestThreadExecution know about Strategy, add onActorStart(iThread) method to Strategy (even to StressStrategy) and call it before every actor.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/ThreadEvent.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.kotlinx.lincheck
+
+/**
+ * ThreadEvent stores information about events occuring during managed execution so that we could
+ * write the execution by Reporter
+ */
+sealed class ThreadEvent(val iThread: Int, val iActor: Int)
+
+class SwitchEvent(iThread: Int, actorId: Int, val info: StackTraceElement, val reason: String = "") : ThreadEvent(iThread, actorId)
```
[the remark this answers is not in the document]
alefedor: Just some info about switch that will be used in the report. For instance, the reason can be that the thread is waiting on the monitor, waiting for the lock or just get stuck in the active lock.

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/RendezvousChannelStressTest.kt
```diff
@@ -33,7 +33,7 @@ import org.junit.Test
 
 
 @Param(name = "value", gen = IntGen::class, conf = "1:5")
-@StressCTest(verifier = LinearizabilityVerifier::class, actorsAfter = 0)
+@StressCTest(verifier = LinearizabilityVerifier::class, actorsAfter = 0, actorsBefore = 0)
```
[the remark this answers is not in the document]
alefedor: Without it, all scenarios have 5 "close" operations in the init part, so the test tests nothing.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
```diff
@@ -34,13 +34,18 @@
...
 public abstract class CachedVerifier implements Verifier {
-    private final Set<ExecutionResult> previousResults = new HashSet<>();
+    private final Map<ExecutionResult, Boolean> previousResults = new HashMap<>();
 
     @Override
     public final boolean verifyResults(ExecutionResult results) {
-        if (!previousResults.add(results))
-            return true;
-        return verifyResultsImpl(results);
+        if (previousResults.containsKey(results)) {
```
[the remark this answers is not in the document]
ndkoval: In the current version, we create new verifiers for each minimization step

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/RendezvousChannelStressTest.kt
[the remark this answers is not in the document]
ndkoval: Let's add `poll` and `offer` operations instead

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/ThreadEvent.kt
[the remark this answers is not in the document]
ndkoval: What do you think about having an enum for reasons?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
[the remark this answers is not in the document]
alefedor: That's not good that we can get different results in functions that seem stateless for an external observer such as ```verify(results)``` when we call them twice

lincheck/pom.xml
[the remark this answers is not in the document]
ndkoval: Cool

libtest/src/test/java/org/jetbrains/kotlinx/lincheck/test/AbstractLincheckTest.kt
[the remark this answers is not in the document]
ndkoval: You can create an internal function in addition to the public `LinChecker.check` one. This function can return a report with some statistics information instead of throwing an exception.

libtest/src/main/java/tests/custom/transfer/AccountsWrong1.java
ndkoval: Let's add this dependency, I don't see a problem.

libtest/src/main/java/tests/custom/set/SynchronizedMethodSet.kt
ndkoval: Probably, it would be better to have a sequential implementation and inherit all these lock-based implementations from it. Please, also put all of them in a single file.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/ThreadEvent.kt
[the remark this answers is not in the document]
alefedor: We can do that

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Actor.kt
ndkoval: You can add `onActorStart(..)` method invocations using the byte-code generator. In this case, the `Strategy` class should have this method (`TestThreadExecution` can know about `Strategy` but not about specific implementations).

alefedor: Would such virtual calls be optimized out by the optimizer for StressStrategy?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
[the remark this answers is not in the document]
ndkoval: We can't

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+
+        boolean isSynchronized = (access & ACC_SYNCHRONIZED) != 0;
+        if (isSynchronized) {
+            access ^= ACC_SYNCHRONIZED; // disable synchronized
+        }
+
         MethodVisitor mv = super.visitMethod(access, mname, desc, signature, exceptions);
         mv = new JSRInlinerAdapter(mv, access, mname, desc, signature, exceptions);
+        mv = new IndirectSharedVariableAccessMethodTransformer(mname, new GeneratorAdapter(mv, access, mname, desc));
+        mv = new SynchronizedLockTransformer(mname, new GeneratorAdapter(mv, access, mname, desc));
```
ndkoval: `SynchronizedLockTransformer` -> `SynchronizedBlockTransformer`

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+
+        if (isSynchronized) {
+            // synchronized method is replaced with synchronized lock
+            mv = new SynchronizedLockAddingTransformer(mname, new GeneratorAdapter(mv, access, mname, desc), className, access, classVersion);
```
ndkoval: Synchronized block

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+        }
+
+        if ("<clinit>".equals(mname)) {
+            mv = new ClassInitializationTransformer(mname, new GeneratorAdapter(mv, access, mname, desc));
+        }
+
+        mv = new WaitNotifyTransformer(mname, new GeneratorAdapter(mv, access, mname, desc));
```
ndkoval: Do you need to apply all these transformers to <clinit>?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
         mv = new SharedVariableAccessMethodTransformer(mname, new GeneratorAdapter(mv, access, mname, desc));
-        // TODO add other transformers
         mv = new TryCatchBlockSorter(mv, access, mname, desc, signature, exceptions);
         return mv;
     }
 
-    class SharedVariableAccessMethodTransformer extends ManagedStrategyMethodVisitor {
-        public SharedVariableAccessMethodTransformer(String methodName, GeneratorAdapter mv) {
+    /**
+     * Changes package of some transformed classes, because they cannot stay in the same package
```
ndkoval: The remapper name specifies these "some" classes, change the comment correspondingly, please.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+     */
+    private static class JavaUtilRemapper extends Remapper {
+        @Override
+        public String map(String name) {
+            // TODO: remove this exception check when exceptions in transformable strategies will be supported
```
ndkoval: apply this TODO :)

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+
+        private boolean isAtomicFieldUpdater(String owner) {
+            return owner.endsWith("FieldUpdater") && owner.contains("Atomic");
+        }
+    }
+
+    /**
+     * Adds invocations of ManagedStrategy methods before direct reads and writes of shared variables
+     */
+    private class SharedVariableAccessMethodTransformer extends ManagedStrategyMethodVisitor {
```
ndkoval: add the logic from `IndirectSharedVariableAccessMethodTransformer` into this method transformer

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+                    mv.pop();
+                    mv.visitLabel(opEnd);
+                    break;
+                case Opcodes.MONITOREXIT:
+                    opEnd = mv.newLabel();
+                    Label skipMonitorExit = mv.newLabel();
+                    mv.dup();
+                    invokeBeforeLockRelease();
+                    mv.push(false);
+                    mv.ifCmp(Type.BOOLEAN_TYPE, EQ, skipMonitorExit);
```
ndkoval: there should be a special instruction for `if (false)`

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+            Label opEnd;
+
+            switch(opcode) {
+                case Opcodes.MONITORENTER:
+                    opEnd = mv.newLabel();
+                    Label skipMonitorEnter = mv.newLabel();
+                    mv.dup();
+                    invokeBeforeLockAcquire();
+                    mv.push(false);
+                    mv.ifCmp(Type.BOOLEAN_TYPE, EQ, skipMonitorEnter);
```
ndkoval: there should be a special instruction for `if (false)`

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+    /**
+     * Replace "method(...) {...}" with "method(...) {synchronized(this) {...} }"
+     */
+    private class SynchronizedLockAddingTransformer extends ManagedStrategyMethodVisitor {
+        private final String className;
+        private final boolean isStatic;
+        private final int classVersion;
+
+        private final Label tryLabel = new Label();
+        private final Label catchLabel = new Label();
```
ndkoval: finally?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+                case RETURN:
+                    loadSynchronizedMethodMonitorOwner();
+                    mv.monitorExit();
+                    break;
+                default:
+                    // do nothing
+            }
+
+            mv.visitInsn(opcode);
+
```
ndkoval: remove empty lines

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -48,80 +46,659 @@
...
+     * Adds invocations of ManagedStrategy methods before wait and after notify calls
+     */
+    private class WaitNotifyTransformer extends ManagedStrategyMethodVisitor {
+        WaitNotifyTransformer(String methodName, GeneratorAdapter mv) {
+            super(methodName, mv);
+        }
+
+        @Override
+        public void visitMethodInsn(int opcode, String owner, String name, String desc, boolean itf) {
+            Label afterWait = mv.newLabel();
```
ndkoval: it can be `notify`

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -184,10 +861,106 @@ void loadNewCodeLocation() {
...
+            } catch(NoSuchFieldException e) {
+                // the field is not in this class
+                // need to go to a super class
+            }
+            clazz = clazz.getSuperclass();
+        } while(clazz != null);
+        throw new NoSuchFieldException();
+    }
+
+    public static class UnsafeLoader {
```
ndkoval: Move to `Utils.kt` and rename to `UnsafeHolder`

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+            // ignore
+        }
+
+        override fun threadEvents(): List<ThreadEvent> = emptyList()
+    }
+
+    /**
+     * Implementation of ThreadEventLogger
+     */
+    protected inner class SimpleEventLogger : ThreadEventLogger {
```
ndkoval: SimpleEventLogger -> ExecutionEventCollector

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+        fun finishThread(iThread: Int)
+        fun passCodeLocation(iThread: Int, codeLocation: Int)
+
+        fun threadEvents(): List<ThreadEvent>
+    }
+
+    /**
+     * Just ignores every event
+     */
+    protected object DummyEventLogger : ThreadEventLogger {
```
ndkoval: Why do you need it?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+
+        fun reset() {
+            lastIThread = Int.MIN_VALUE
+        }
+    }
+
+    /**
+     * Logs thread events such as thread switches and passed codeLocations
+     */
+    protected interface ThreadEventLogger {
```
ndkoval: Why do you need it?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+            notifyAll(monitor)
+        }
+
+        fun notifyAll(monitor: Any) {
+            for (iThread in needsNotification.indices)
+                if (acquiringMonitor[iThread] === monitor)
+                    needsNotification[iThread] = false
+        }
+
+        // TODO: add tests on wait-notify using custom scenarios
```
ndkoval: You can use custom ExecutionGenerator implementations

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.java
```diff
@@ -46,18 +52,22 @@
...
+    protected final Verifier verifier;
 
     protected Strategy(ExecutionScenario scenario, Verifier verifier, Reporter reporter) {
         this.scenario = scenario;
         this.verifier = verifier;
         this.reporter = reporter;
     }
 
-    protected void verifyResults(ExecutionResult results) {
+    protected void verifyResults(ExecutionResult results, List<ThreadEvent> threadEvents) {
```
ndkoval: threadEvents -> interleaving

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.java
```diff
@@ -70,6 +80,13 @@ public boolean needsTransformation() {
         return false;
     }
 
+    /**
+     * Check if Strategy allows resuming the coroutine.
+     */
+    public boolean canResumeCoroutine(int iThread) { return true; }
```
ndkoval: why `true`?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.java
```diff
@@ -46,18 +52,22 @@
...
         this.reporter = reporter;
     }
 
-    protected void verifyResults(ExecutionResult results) {
+    protected void verifyResults(ExecutionResult results, List<ThreadEvent> threadEvents) {
         if (!verifier.verifyResults(results)) {
             StringBuilder msgBuilder = new StringBuilder("Invalid interleaving found:\n");
             appendIncorrectResults(msgBuilder, scenario, results);
+            if (threadEvents != null) {
+                msgBuilder.append(System.lineSeparator());
```
ndkoval: you use `\n` two lines above :)
Either use `System.lineSeparator()` or `\n` everywhere

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomsearch/RandomSearchStrategy.kt
```diff
@@ -0,0 +1,157 @@
...
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.random.Random
+
+/**
+ * RandomSearchStrategy at first studies codeLocations at the choosen thread
+ * until a blocking event or a thread finish and
+ * then chooses one uniform randomly
+ */
+class RandomSearchStrategy(
```
ndkoval: Can we brute force interleavings without repetitions? What does this strategy guarantee?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomsearch/RandomSearchStrategy.kt
```diff
@@ -0,0 +1,157 @@
...
+        testClass: Class<*>,
+        scenario: ExecutionScenario,
+        verifier: Verifier,
+        testCfg: RandomSearchCTestConfiguration,
+        reporter: Reporter
+) : ManagedStrategyBase(testClass, scenario, verifier, reporter, testCfg.maxRepetitions, testCfg.checkObstructionFreedom) {
+    // an increasing id of operation performed within this execution
+    private val executionPosition = AtomicInteger(0)
```
ndkoval: you don't need _atomic_ registers since the strategy is sequential by design

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    override fun onActorStart(iThread: Int) {
+        currentActorId[iThread]++
+    }
+
+    /**
+     * Detects loop when visiting a codeLocation too often
+     */
+    protected class LoopDetector(val maxRepetitions: Int) {
```
ndkoval: LiveLockDetector

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+    // detector of loops (i.e. active locks)
+    protected val loopDetector = LoopDetector(maxRepetitions)
+    // logger of all events in the execution such as thread switches
+    protected var eventLogger: ThreadEventLogger = DummyEventLogger
+    // tracker of acquisitions and releases of monitors
+    protected val monitorTracker = MonitorTracker(nThreads)
+    // random used for execution
+    protected var executionRandom = PseudoRandom()
+    // copy of random used for execution for reproducability of results
+    protected var executionRandomCopy = executionRandom
```
ndkoval: Can't you just introduce a 'lastValue()' method? AFAIR, you already keep these values.

ndkoval:
Please, rebase on the current develop

ndkoval:
Let's remove `jagent` from dependencies. You use only `ClassInfoVisitor` class, and it is possible to get all the required information by overriding `visit(...)` method in your class visitor; see the link below for details.
https://github.com/devexperts/jagent/blob/master/jagent-impl/src/main/java/com/devexperts/jagent/ClassInfoVisitor.java

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/RendezvousChannelStressTest.kt
```diff
@@ -47,6 +47,12 @@ class RendezvousChannelStressTest : VerifierState() {
     @Operation(handleExceptionsAsResult = [ClosedReceiveChannelException::class])
     suspend fun receiveOrNull() = ch.receiveOrNull()?.plus(100)
 
+    @Operation(handleExceptionsAsResult = [ClosedSendChannelException::class])
```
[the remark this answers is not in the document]
alefedor: It seems to me that you can not omit them here

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/util/PseudoRandom.kt
```diff
@@ -0,0 +1,102 @@
...
+                }
+
+                numbers.add(value)
+            }
+        }
+
+        return directionNumbers[dimensionId][id]
+    }
+
+    private fun lowestZeroBit(a: Int): Int {
```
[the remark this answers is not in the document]
alefedor: Then we need to use experimental ```UInt```

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyBase.kt
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    /**
+     * Returns all data to the initial state before invocation
+     */
+    protected open fun initializeInvocation() {
+        finished.forEach { it.set(false) }
+        isSuspended.forEach { it.set(false) }
+        // save previous random state
+        executionRandomCopy = executionRandom.copy()
```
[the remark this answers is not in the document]
alefedor: Because we need to be able to reproduce the execution in order to collect information about it

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomswitch/RandomSwitchCTest.java
```diff
@@ -135,4 +146,5 @@
     @interface RandomSwitchCTests {
```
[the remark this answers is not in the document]
alefedor: 2. Seem not work for java classes

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomswitch/RandomSwitchOptions.java
```diff
@@ -1,49 +1,71 @@
...
+
 import static org.jetbrains.kotlinx.lincheck.UtilsKt.chooseSequentialSpecification;
 
 /**
- * Options for {@link RandomSwitchStrategy random-switch} strategy.
+ * Options for {@link RandomSwitchStrategy random search} strategy.
  */
 public class RandomSwitchOptions extends Options<RandomSwitchOptions, RandomSwitchCTestConfiguration> {
-    protected int invocationsPerIteration = RandomSwitchCTestConfiguration.DEFAULT_INVOCATIONS;
+    protected boolean checkObstructionFreedom = DEFAULT_CHECK_OBSTRUCTION_FREEDOM;
```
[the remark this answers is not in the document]
alefedor: I can do it with some probability

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomswitch/RandomSwitchStrategy.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+
+    override fun shouldSwitch(iThread: Int): Boolean {
+        // TODO: can reduce number of random calls using geometric distribution
+        return executionRandom.nextDouble() < switchProbability
+    }
+
+    override fun initializeInvocation() {
+        executionRandom.endLastPoint() // a point corresponds to an invocation
```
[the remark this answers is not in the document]
alefedor: Isn't it a uniform random?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyHolder.java
```diff
@@ -28,6 +28,7 @@
  */
 public class ManagedStrategyHolder {
     public static ManagedStrategy strategy;
+    public static LocalObjectManager objectManager;
```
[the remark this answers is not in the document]
alefedor: Its quite related to the ```ManagedStrategy```

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -63,13 +72,24 @@ public TransformationClassLoader(Strategy strategy, Runner runner) {
...
+        if (className.startsWith(TRANSFORMED_POINTED_PACKAGE)) {
+            return false;
+        }
         return className == null ||
-            (className.startsWith("org.jetbrains.kotlinx.lincheck.") &&
+            (
+                className.startsWith("org.jetbrains.kotlinx.lincheck.") &&
                 !className.startsWith("org.jetbrains.kotlinx.lincheck.test.") &&
-                !className.equals("ManagedStrategyHolder")) ||
+                !className.startsWith("org.jetbrains.kotlinx.lincheck.tests.") &&
```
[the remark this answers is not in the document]
alefedor: We have "tests" in libtest and "test" in lincheck module

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -63,13 +72,24 @@ public TransformationClassLoader(Strategy strategy, Runner runner) {
...
+                !className.equals("org.jetbrains.kotlinx.lincheck.strategy.ManagedStrategyHolder")
```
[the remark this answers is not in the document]
alefedor: Otherwise we will have the same static class for all class loaders

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategy.java
```diff
@@ -22,28 +22,29 @@
...
 /**
- * This is an abstract class for all managed strategies.
- * This abstraction helps to choose a proper {@link Runner},
- * to transform byte-code in order to insert required for managing instructions,
- * and to hide class loading problems from the strategy algorithm.
+ * This is an abstract class for all strategies with transformers.
+ * This abstraction helps to transform byte-code in order to insert required for
+ * managing instructions and to hide class loading problems from the strategy algorithm.
  */
 public abstract class ManagedStrategy extends Strategy {
```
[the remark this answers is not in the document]
alefedor: There is too much logic in ```ManagedStrategyBase``` while ``ManagedStrategy``` primarily contains interface description

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategy.java
```diff
@@ -58,41 +59,48 @@ public void onStart(int iThread) {
...
             }
         };
         ManagedStrategyHolder.setStrategy(runner.classLoader, this);
     }
 
     @Override
     public ClassVisitor createTransformer(ClassVisitor cv, ClassInfo classInfo) {
-        return transformer = new ManagedStrategyTransformer(cv, classInfo);
+        List<StackTraceElement> previousCodeLocations;
+        if (transformer == null) {
```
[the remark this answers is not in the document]
alefedor: A transformer is created for each transformed class

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.java
alefedor: For example, in ```StressStrategy``` it should be true. I don't see a reason to have it false by default.

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategy.java
```diff
@@ -142,24 +149,33 @@ public void beforeSharedVariableWrite(int iThread, int codeLocation) {}
...
     /**
      *
      * @param iThread the number of the executed thread according to the {@link ExecutionScenario scenario}.
      * @param codeLocation the byte-code location identifier of this operation.
      * @param monitor
+     * @return whether lock should be actually released
      */
-    public void afterLockRelease(int iThread, int codeLocation, Object monitor) {}
+    public boolean beforeLockRelease(int iThread, int codeLocation, Object monitor) {
+        return true;
```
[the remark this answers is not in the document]
alefedor: I can return false only if I haven't actually acquired the lock

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
alefedor: To what?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -184,10 +861,106 @@ void loadNewCodeLocation() {
...
+        }
+
         @Override
         public void visitLineNumber(int line, Label start) {
             this.lineNumber = line;
             super.visitLineNumber(line, start);
         }
     }
+
+    private boolean isFinalField(String owner, String name) {
```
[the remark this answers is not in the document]
alefedor: We can not, because it is also used in `SharedVariableAccessMethodTransformer`

ndkoval:
@alefedor Please, rebase your code onto *develop* and add `suspend` functions support (see the linearizability verifier) as well as `HBClock`-related feature (in the model checking mode you always know happens-before relations between all the scenario operations).

PR #6, opened 2019-07-31
Check correctness of state equivalency relation

SokolovaMaria:


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/LinChecker.java
```diff
@@ -110,6 +110,20 @@ private void checkImpl(CTestConfiguration testCfg) throws AssertionError, Except
         }
     }
 
+    private void checkStateEquivalenceImpl(Class<?> testClass) {
+        if (!testClass.isAnnotationPresent(QuantitativeRelaxationVerifierConf.class)) {
+            Object i1 = createTestInstance(testClass);
+            Object i2 = createTestInstance(testClass);
+            if (i1.hashCode() != i2.hashCode() || !i1.equals(i2)) {
+                throw new IllegalStateException(
+                    "equals() and hashCode() methods for a test instance are not defined or defined incorrectly.\n" +
```
ndkoval: for a test instance -> for this test

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/LinChecker.java
```diff
@@ -110,6 +110,20 @@ private void checkImpl(CTestConfiguration testCfg) throws AssertionError, Except
...
+                    "To define equals() and hashCode() methods the test class should extend `VerifierState` class and override `extractState()` function.\n" +
```
ndkoval: Please, use ... to define ...

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/LinChecker.java
```diff
@@ -110,6 +110,20 @@ private void checkImpl(CTestConfiguration testCfg) throws AssertionError, Except
...
+                    "This check may be suppressed by setting the option `requireStateEquivalenceImplementationCheck = false`."
```
ndkoval: by setting the `req..` option to `false`.

PR #7, opened 2019-07-31
Method loading fixed

SokolovaMaria:


ndkoval:
Please, add a test for this bug

PR #8, opened 2019-08-01
Custom scenario DSL

alefedor:
A feature to generate custom scenarios with DSL builder and to run tests on them.
DSL scenario generation:
```kotlin
scenario {
  initial {
    actor(SPMCQueue::offer, 1)
    actor(SPMCQueue::offer, 2)
  }
  parallel {
    thread {
      elements.forEach { actor(SPMCQueue::offer, it) }
    }
    thread {
      repeat(5) {
        actor(SPMCQueue::poll)
      }
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Table of contents
     + [Operation groups](#operation-groups)
   * [Parameter generators](#parameter-generators)
     + [Binding parameter and generator names](#binding-parameter-and-generator-names)
+  * [Custom scenarios](#custom-scenarios)
   * [Sequential specification](#sequential-specification)
   * [Validation functions](#validation-functions)
   * [Parameter and result types](#parameter-and-result-types)
@@ -146,6 +147,39 @@ Unfortunately, this feature is disabled in **javac** compiler by default. Use `-
 ```
 
 However, some IDEs (such as IntelliJ IDEA) do not understand build system configuration as well as possible and running a test from these IDEs will not work. In order to solve this issue you can add `-parameters` option for **javac** compiler in your IDE configuration.
+
+## Custom scenarios
+Sometimes, it is important to be confident that the testing algorithm works under some corner-case situations.
+For this purpose, **lincheck** provides a possibility to specify *custom* scenarios via a special Kotlin DSL in a way similar to the example below.
+After the scenario is defined, it can be added to the configuration via `Options.addCustomScenario(ExecutionScenario)`
+(see [Configuration via options](#configuration-via-options) for details of how to configure Lincheck via `Options`).
+In this case, **lincheck**  examines custom scenarios followed by checking the generated ones.
+
+Custom scenario generation in Kotlin can be done as follows:
+```kotlin
+val scenario = scenario {
+  initial { // initialize the queue with two elements
+    actor(SPMCQueue::offer, 1)
+    actor(SPMCQueue::offer, 2)
+  }
+  parallel {
+    thread { // one producer 
+      // add elements one-by-one
+      elements.forEach { actor(SPMCQueue::offer, it) }
+    }
+    repeat(2) { // two consumers
+      thread {
+        repeat(3) { // add three poll-s 
+          actor(SPMCQueue::poll)
+        }
+      }
+    }
+  }
+}
+
+// Add this custom scenario to the test configuration
+options.addCustomScenario(scenario)
+```
 
 ## Sequential specification
 By default, **lincheck** sequentially uses the testing data structure to define the correct specification.
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestConfiguration.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestConfiguration.kt
@@ -48,7 +48,8 @@ abstract class CTestConfiguration(
     val requireStateEquivalenceImplCheck: Boolean,
     val minimizeFailedScenario: Boolean,
     val sequentialSpecification: Class<*>,
-    val timeoutMs: Long
+    val timeoutMs: Long,
+    val customScenarios: List<ExecutionScenario>
 ) {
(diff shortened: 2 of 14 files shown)


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
```diff
@@ -0,0 +1,133 @@
...
+import kotlin.reflect.KFunction
+import kotlin.reflect.jvm.javaMethod
+
+/**
+ * Kotlin DSL for defining custom scenarios.
+ *
+ *
+ * Example:
+ * ```
+ * scenario {
```
SokolovaMaria: In our semantics `Operation` = `Actor` + `Result`, and custom scenarios are defined by actors only. Please, get rid of this `operation` constructor and use `actor` constructor to create actors for a scenario.
So, the dsl should look like this:
```
scenario {
     initial {
      actor(CustomTest::offer, 1)
    }
    ....
 }
```

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
```diff
@@ -0,0 +1,133 @@
...
+    require(method.exceptionTypes.all { Throwable::class.java.isAssignableFrom(it) }) { "Not all declared exceptions are Throwable" }
+    return Actor(
+        method = method,
+        arguments = args.toList(),
+        handledExceptions = (method.exceptionTypes as Array<Class<out Throwable>>).toList()
+    )
+}
+
+@ScenarioDSLMarker
+class DSLThreadScenario : ArrayList<Actor>() {
```
SokolovaMaria: You can get rid of this class, directly using `actor` constructor to create actors for a scenario

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
```diff
@@ -0,0 +1,133 @@
+/*-
```
SokolovaMaria: Please, move it out to `org.jetbrains.kotlinx.lincheck` package

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
SokolovaMaria: And fix the custom scenario dsl example in README

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/CustomScenarioDSL.kt
```diff
@@ -64,51 +64,49 @@ fun verify(
     if (expected) assert(res) else assert(!res)
 }
 
+fun actor(f: KFunction<*>, vararg args: Any?): Actor =
```
SokolovaMaria: Get rid of this delegation - use `actor` function directly

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
```diff
@@ -0,0 +1,133 @@
...
+        return ExecutionScenario(
+                initial,
+                parallel,
+                post
+        )
+    }
+}
+
+@DslMarker
+private annotation class ScenarioDSLMarker
```
SokolovaMaria: Please, add a test, demonstrating usage of this dsl

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
alefedor: I thought that `actor` is an internal name in LinCheck while `operation` is public and understandable because of `@Operation` annotation.
There are 77 mentions of `operation` in readme, but only 4 of `actor` (such as `actorsPerThread` parameter, note that its definition has word `operations`)

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/CustomScenarioDSL.kt
alefedor: Is it ok that then `actor` should be imported separately from this execution DSL?

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
SokolovaMaria: From the API perspective `operation` is just a method without concrete parameters (`@Operation` marks just a method to test) whereas `actor = method + parameters`.
So, it's better to use `actor` constructor for this dsl.

alefedor: Ok

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/CustomScenarioDSL.kt
alefedor: If we get rid of this class It will allow to write code such as
```
scenario {
  actor(Queue::add, 1)
  parallel {...}
  actor(Queue::remove, 1)
}
```
`initial` and `post` are mistakenly omitted but a programmer won't notice it until the test is invoked

ndkoval (approved):
I slightly changed the behavior -- custom scenarios are now ADDITIONAL to the generated ones

Issue #9, opened 2019-08-02
Wrong alignment in incorrect results representation


alefedor:
Currently a wrong alignment is added to init and post parts (additional meaningless spaces).

Example of wrong representation: 
```
= Invalid execution results: =
Init part:
[getAmount(4):     0, transfer(3,1,-9): void, getAmount(2):     0, transfer(3,2,-7): void, setAmount(2,7):   void]
Parallel part:
| setAmount(1,1):   void | setAmount(4,-8): void | setAmount(4,6):  void |
| setAmount(4,-4):  void | setAmount(3,9):  void | setAmount(2,-1): void |
| transfer(4,3,10): void | setAmount(4,4):  void | getAmount(3):    12   |
| transfer(3,4,7):  void | setAmount(1,-8): void | getAmount(4):    -7   |
| setAmount(2,4):   void | setAmount(3,1):  void | getAmount(3):    12   |
Post part:
[transfer(4,4,7):  void, getAmount(1):     -8, getAmount(3):     12, getAmount(1):     -8, transfer(1,4,-7): void]
```

There are odd spaces in third and last lines.

PR #10, opened 2019-08-02
Fixed bug with wrong alignment in Reporter

alefedor:
A fix for this: #9


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -86,18 +86,21 @@ private class ActorWithResult(val actorRepresentation: String, val spaces: Int,
     override fun toString(): String = actorRepresentation + ":" + " ".repeat(spaces) + resultRepresentation
 }
 
-private fun uniteActorsAndResults(actors: List<Actor>, results: List<Result>): List<ActorWithResult> {
+private fun uniteActorsAndResults(actors: List<Actor>, results: List<Result>, needsAlignment: Boolean = false): List<ActorWithResult> {
```
ndkoval: You can use this method for the parallel part only

PR #16, opened 2019-09-08
Fixed getting void results after invocation of kotlin Unit methods

SokolovaMaria:


lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/CancellableChannelTest.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    @Operation
+    fun poll() = try {
+        c.poll()
+    } catch (e : NumberedCancellationException) {
+        e.testResult
+    }
+
+    @Operation
+    fun cancel(@Param(gen = IntGen::class) token: Int) = c.cancel(NumberedCancellationException(token))
```
ndkoval: 1. It should fail
2. You already have a test with Kotlin Coroutines channels

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -143,22 +143,22 @@ private fun getRelaxedMethod(testInstance: Any, actor: Actor): Method = relaxedM
...
     res === COROUTINE_SUSPENDED -> NoResult.also { it.wasSuspended = true }
-    res is kotlin.Result<Any?> -> res.toLinCheckResult(wasSuspended)
-    else -> ValueResult(res)
+    res is kotlin.Result<Any?> -> res.toLinCheckResult().also { it.wasSuspended = wasSuspended }
+    else -> ValueResult(res).also { it.wasSuspended = wasSuspended }
 }
 
-private fun kotlin.Result<Any?>.toLinCheckResult(wasSuspended: Boolean = false) =
+private fun kotlin.Result<Any?>.toLinCheckResult() =
     (if (isSuccess) {
```
ndkoval: do we need these outer "()" brackets?

SokolovaMaria: Removed

PR #19, opened 2019-09-19
Adding sequential specification for verifiers

ndkoval:
Instead of using concurrent implementations sequentially a sequential way as a specification, it is sometimes better to write a simple sequential implementation instead.


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/verifier/LTS.kt
```diff
@@ -90,9 +93,9 @@ typealias ResumedTickets = Set<Int>
  */
 
 class LTS(
-    private val testClass: Class<*>,
-    private val isQuantitativelyRelaxed: Boolean,
-    private val relaxationFactor: Int
+        private val sequentialSpecification: Class<*>,
```
SokolovaMaria: Extra tabs again

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/strategy/randomswitch/RandomSwitchCTest.java
```diff
@@ -115,6 +117,17 @@
      */
     boolean minimizeFailedScenario() default true;
 
+    /**
+     * The specified class defines the sequential behavior of the testing data structure;
```
SokolovaMaria: test**ed** data structure

PR #30, opened 2019-10-24
Updated bytecode transformation dependencies

alefedor:
Changes:
1) Updated version of ASM to 7.1. Now LinCheck can transform classes up to Java 13.
2) Got rid of dependency from Devexperts Bintray  repository.

### DIFF
--- a/lincheck/pom.xml
+++ b/lincheck/pom.xml
@@ -13,13 +13,13 @@
     <dependencies>
         <dependency>
             <groupId>org.ow2.asm</groupId>
-            <artifactId>asm-all</artifactId>
-            <version>5.2</version>
+            <artifactId>asm-commons</artifactId>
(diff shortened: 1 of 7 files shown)


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassWriter.java
```diff
@@ -0,0 +1,40 @@
...
+ * #L%
+ */
+package org.jetbrains.kotlinx.lincheck;
+
+import org.objectweb.asm.ClassWriter;
+
+/**
+ * ClassWriter for classes transformed by LinCheck
+ */
+public class TransformationClassWriter extends ClassWriter {
```
ndkoval: You don't need this class.

PR #31, opened 2019-12-24
Make the heuristic for fixing starvation on internal spin loops adaptive

ndkoval:


lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -115,6 +119,15 @@ open class ParallelThreadsRunner(
...
         completedOrSuspendedThreads.set(0)
         completions.forEach { it.forEach { it.resWithCont.set(null) } }
+        uninitializedThreads.set(scenario.threads)
+        // update `spinningTimeBeforeYield` adaptively
+        if (yieldInvokedInOnStart) {
+            spinningTimeBeforeYield = (spinningTimeBeforeYield + 1) / 2
+            yieldInvokedInOnStart = false
+        } else {
+            spinningTimeBeforeYield *= 2
+            if (spinningTimeBeforeYield > 10_000_000) spinningTimeBeforeYield = 10_000_000
```
SokolovaMaria: Make `10_000_000` a constant value

PR #33, opened 2020-03-31
Linearizability support additionally to sequential consistency

ndkoval:
Although *lincheck* is called ***lin**check*, it was able to test only for sequential consistency (or weaker guarantees) so far. This pull request adds linearizability support by adding additional synchronization between threads from time to time and taking it into account during the verification phase.


lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/ClocksTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    // the execution is sequential consistent but not linearizable
+    @Test
+    fun test() = try {
+        LinChecker.check(this::class.java)
+        throw AssertionError("Should fail since the execution with clocks is " +
+                             "sequential consistent but not linearizable")
+    } catch (e: AssertionError) {
+//        check(e.message!!.contains("Invalid interleaving found")) {
```
ndkoval: Please uncomment

lincheck/src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -149,14 +149,18 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
                                     boolean waitsEnabled, boolean scenarioContainsSuspendableActors)
     {
         int access = ACC_PUBLIC;
-        Method m = new Method("call", RESULT_ARRAY_TYPE, NO_ARGS);
+        Method m = new Method("run", Type.VOID_TYPE, NO_ARGS);
```
SokolovaMaria, line 152: Replace all `Type.*` usages with static imports of types

lincheck/src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/ClocksTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import org.jetbrains.kotlinx.lincheck.execution.*
+import org.jetbrains.kotlinx.lincheck.strategy.stress.*
+import org.junit.*
+import kotlin.reflect.jvm.*
+
+
+@StressCTest(generator = ClocksTestScenarioGenerator::class, iterations = 1,
+             actorsBefore = 0, threads = 2, actorsPerThread = 1, actorsAfter = 0,
+             sequentialSpecification = ClocksTestSequential::class,
+             requireStateEquivalenceImplCheck = false, minimizeFailedScenario = false)
```
SokolovaMaria, line 35: We were going to add some error message referring to check the `minimizeFailedScenario` parameter in case of the hanging execution like here

ndkoval: We do not have such issues in practice (only in our internal tests, and only once so fat) so that such a message seems redundant. Let's just keep this in mind.

PR #40, opened 2020-07-03
Model checking strategy

alefedor:
This pull request adds *model checking strategy* - a managed strategy that goes through all possible interleavings in increasing order of the number of thread context switches. 

The main advanvantage of *model checking strategy* over *stress strategy* is that it can report the information of the thread switches needed to get an incorrect result.

The example of model checking strategy output can be:
```
= Invalid execution results =
Parallel part:
| put(4, -4): null | put(-8, -8): null |
Post part:
[get(-8): null]
= The execution that led to this result =
= Parallel part execution: =
| put(4, -4)                                                  |                      |
|   table.READ: null at HashMap.putVal(HashMap.java:628)      |                      |
|   table.READ: null at HashMap.resize(HashMap.java:678)      |                      |
|   switch                                                    |                      |
|                                                             | put(-8, -8): null    |
|                                                             |   thread is finished |
|   threshold.READ: 12 at HashMap.resize(HashMap.java:680)    |                      |
|   threshold.WRITE(9) at HashMap.resize(HashMap.java:702)    |                      |
|   table.WRITE(Node[]@1) at HashMap.resize(HashMap.java:705) |                      |
|   READ: null at HashMap.putVal(HashMap.java:630)            |                      |
|   WRITE(Node@1) at HashMap.putVal(HashMap.java:631)         |                      |
|   modCount.READ: 1 at HashMap.putVal(HashMap.java:661)      |                      |
|   modCount.WRITE(2) at HashMap.putVal(HashMap.java:661)     |                      |
|   size.READ: 1 at HashMap.putVal(HashMap.java:662)          |                      |
|   size.WRITE(2) at HashMap.putVal(HashMap.java:662)         |                      |
|   threshold.READ: 9 at HashMap.putVal(HashMap.java:662)     |                      |
|   result: null                                              |                      |
|   thread is finished                                        |                      |
```


src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -45,30 +50,56 @@
...
+    private static final Method BEFORE_WAIT_METHOD = new Method("beforeWait", Type.BOOLEAN_TYPE, new Type[]{Type.INT_TYPE, Type.INT_TYPE, OBJECT_TYPE, Type.BOOLEAN_TYPE});
+    private static final Method AFTER_NOTIFY_METHOD = new Method("afterNotify", Type.VOID_TYPE, new Type[]{Type.INT_TYPE, Type.INT_TYPE, OBJECT_TYPE, Type.BOOLEAN_TYPE});
+    private static final Method BEFORE_PARK_METHOD = new Method("beforePark", Type.BOOLEAN_TYPE, new Type[]{Type.INT_TYPE, Type.INT_TYPE, Type.BOOLEAN_TYPE});
+    private static final Method AFTER_UNPARK_METHOD = new Method("afterUnpark", Type.VOID_TYPE, new Type[]{Type.INT_TYPE, Type.INT_TYPE, OBJECT_TYPE});
+    private static final Method ENTER_IGNORED_SECTION_METHOD = new Method("enterIgnoredSection", Type.VOID_TYPE, new Type[]{Type.INT_TYPE});
+    private static final Method LEAVE_IGNORED_SECTION_METHOD = new Method("leaveIgnoredSection", Type.VOID_TYPE, new Type[]{Type.INT_TYPE});
+
+    private static final Method NEW_LOCAL_OBJECT_METHOD = new Method("newLocalObject", Type.VOID_TYPE, new Type[]{OBJECT_TYPE});
```
ndkoval, line 74: It is better to reference methods/fields directly, but you have to switch to Kotlin in this case... 

Example:
```
    val m = Method.getMethod(ManagedStrategy::currentThreadNumber.javaMethod)
```

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -10,31 +10,36 @@
  * it under the terms of the GNU Lesser General Public License as
```
ndkoval, line 10: It is hard to accept, but it seems that the same code in Kotlin would look much better and safer for further refactorings.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -80,58 +111,667 @@ public void visitSource(String source, String debug) {
...
+        return false;
+    }
+
+    /**
+     * Changes package of transformed classes from java/util package, excluding some
+     */
+    private static class JavaUtilRemapper extends Remapper {
+        @Override
+        public String map(String name) {
+            boolean isException = name.endsWith("Exception");
```
ndkoval, line 167: Let's check whether it is inherited from `Throwable` instead (seems bug-prone now)

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStrategyTransformer.java
```diff
@@ -80,58 +111,667 @@ public void visitSource(String source, String debug) {
...
+            boolean isTrustedAtomicPrimitive = TrustedAtomicPrimitives.INSTANCE.isTrustedPrimitive(name);
+            // function package is not transformed, because AFU uses it and thus there will be transformation problems
+            if (name.startsWith("java/util/") && !isTrustedAtomicPrimitive && !isException)
+                name = TRANSFORMED_PACKAGE_INTERNAL_NAME + name;
+            return name;
+        }
+    }
+
+    /**
+     * Generates body of a native method VMSupportsCS8
```
ndkoval, line 177: Please add a detailed comment on why we do need this visitor, it does not seem obvious.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/TrustedAtomicPrimitives.kt
```diff
@@ -0,0 +1,61 @@
...
+package org.jetbrains.kotlinx.lincheck.strategy
+
+/**
+ * Some atomic primitives are common and can be analyzed from a higher level of abstraction
+ * or can not be transformed (i.e, Unsafe or AFU).
+ * We don't transform them and improve the output of lincheck for them.
+ * For example, in the execution instead of a code location in AtomicLong.get() method
+ * we could just print the code location where the method is called.
+ */
+internal object TrustedAtomicPrimitives {
```
ndkoval, line 31: Cool feature!

ndkoval: Let's expand this feature on other trusted blocks. E.g., you use `ConcurrentHashMap` in your implementation and you believe that its operations are atomic (e.g., `put`, `get`, `remove` but not `size` are linearizable). Thus, you can exclude the `ConcurrentHashMap` from the search and save resources for more detailed analysis. 

To be short, we need contracts here.

src/main/java/org/jetbrains/kotlinx/lincheck/runner/InvocationResult.kt
```diff
@@ -53,4 +53,8 @@ class ValidationFailureInvocationResult(
     val scenario: ExecutionScenario,
```
ndkoval, line 53: Let's add a comment for `ValidationFailureInvocationResult`

src/main/java/org/jetbrains/kotlinx/lincheck/runner/InvocationResult.kt
```diff
@@ -53,4 +53,8 @@ class ValidationFailureInvocationResult(
...
     val functionName: String,
     val exception: Throwable
+) : InvocationResult()
+
+class ObstructionFreedomViolationInvocationResult(
```
ndkoval, line 58: Documentation

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -161,7 +163,7 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
         mv.loadThis();
         mv.getField(TEST_THREAD_EXECUTION_TYPE, "results", RESULT_ARRAY_TYPE);
         mv.storeLocal(resLocal);
-        // Call runner's onStart(iThread) method
+        // Call runner's onStart(threadId) method
```
ndkoval, line 166: Is `threadId` is consistent with thread identifiers in verifies?

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -194,16 +196,26 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
...
-                start = mv.newLabel();
-                end = mv.newLabel();
                 handler = mv.newLabel();
-                handlerEnd = mv.newLabel();
                 for (Class<? extends Throwable> ec : actor.getHandledExceptions())
                     mv.visitTryCatchBlock(start, end, handler, getType(ec).getInternalName());
                 mv.visitLabel(start);
             }
+            // Catch those exceptions that were not catched yet
+            Label remainingExceptionHandler = mv.newLabel();
```
ndkoval, line 210: `remainingExceptionHandler` -> `unexpectedExceptionHandler`

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -194,16 +196,26 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
...
             if (actor.getHandlesExceptions()) {
-                start = mv.newLabel();
-                end = mv.newLabel();
                 handler = mv.newLabel();
-                handlerEnd = mv.newLabel();
                 for (Class<? extends Throwable> ec : actor.getHandledExceptions())
                     mv.visitTryCatchBlock(start, end, handler, getType(ec).getInternalName());
                 mv.visitLabel(start);
             }
+            // Catch those exceptions that were not catched yet
```
ndkoval, line 209: Guess, the problem you are fixing here appears in both stress and managed strategies. Let's move this code with a test to a separate pull request.

ndkoval: Not obvious how to solve this problem for stress strategies...

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -55,6 +55,8 @@
     private static final Type RUNNER_TYPE = getType(Runner.class);
     private static final Method RUNNER_ON_START_METHOD = new Method("onStart", VOID_TYPE, new Type[]{INT_TYPE});
     private static final Method RUNNER_ON_FINISH_METHOD = new Method("onFinish", VOID_TYPE, new Type[]{INT_TYPE});
```
ndkoval, line 57: `Method.getMethod(...)` in Kotlin :)

src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -210,4 +212,32 @@ fun storeCancellableContinuation(cont: CancellableContinuation<*>) {
     } else {
         storedLastCancellableCont = cont
     }
-}
\ No newline at end of file
+}
+
+object UnsafeHolder {
```
ndkoval, line 217: Does it work on Java 9+?

src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -210,4 +212,32 @@ fun storeCancellableContinuation(cont: CancellableContinuation<*>) {
...
+            } catch (e: Exception) {
+                throw RuntimeException(e)
+            }
+        }
+
+        return theUnsafe!!
+    }
+}
+
+fun collectThreadDump() = Thread.getAllStackTraces()
```
ndkoval, line 237: Let's move the filtering logic in `Reporter` into this method (it is better when all the thread dump transformations are in a single place)

src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
ndkoval, line 217: Do we really need `Unsafe`? Let's add a comment on it!

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/LocalObjectManager.kt
```diff
@@ -0,0 +1,61 @@
...
+ * #L%
+ */
+package org.jetbrains.kotlinx.lincheck.strategy
+
+import java.util.*
+
+/**
+ * Tracks the creation of local objects and leaks of their references to non-local objects.
+ */
+internal class LocalObjectManager {
```
ndkoval, line 29: Let's add an internal option to switch on/off this and run an experiment with and without this feature.

ndkoval:
When you use suspension functions, there are several switch points related to the internal coroutines logic that should be ignored.

E.g., with this operation:
```
    @Operation(cancellableOnSuspension = false)
    suspend fun inc(): Int = try {
        m.acquireWrite()
        c++
    } finally {
        m.releaseWrite()
    }
```

The beginning of a possible trace is the following:
```
inc() *   
      pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:0) 
      pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:350)
      pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:351)
      pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:351)
      ...      
```

ndkoval:
Let's change such lines
`switch at: ReadWriteMutex.releaseWrite(SegmentQueueSynchronizerLCStressTests.kt:178) `
to
`switch to thread $id`

ndkoval:
Interleaving descriptions are too wide, let's remove whitespace with "pass:" in the beginning, i.e., this line
`      pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:0)`
changes to
`  ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:0)`

ndkoval:
Sometimes, the whole functions are invoked, so that we are not interested in details and it would be better to show in interleavings that the function is invoked; follow the example below. 

Here is an operation:
```
    @Operation(cancellableOnSuspension = false)
    suspend fun inc(): Int = try {
        m.acquireWrite()
        c++
    } finally {
        m.releaseWrite()
    }
```

Here is a part of the interleaving:
```
 inc() *                                                                                              
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1245)            
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1257)          
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:69) 
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:75) 
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1277)   
       pass: ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1286)      
       pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:352)  
       pass: ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:352)   
```

It should be replaced with the following one:
```
 inc() *                                                                                              
       "acquireWrite" at ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:351)
       READ at ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:352)  
       WRITE at ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:352) 
```

Moreover, it is better to add atomic operation types (READ, WRITE, CAS_succ, CAS_fail) to the beginning of events

ndkoval:
When you debug your code, you need to draw the state of your data structure after each interleaving point to understand all the details and problems. Instead of doing it manually, we can automize the process and draw these states after each event automatically if a special function with `@StateRepresentation` annotation that returns the state representation (returns `String`) exists in your test.

ndkoval:
We have a lot of useless events on suspension:

```
inc() *                                                                                                   
   ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:0)             
   ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:350)           
   ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:351)           
   ReadWriteMutexCounterLCStressTest.inc(SegmentQueueSynchronizerLCStressTests.kt:351)           
   ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1245)                    
   ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:1257)                    
   ReadWriteMutex.acquireWrite(SegmentQueueSynchronizerLCStressTests.kt:69)                      
   CancellableContinuationImpl.<init>(CancellableContinuationImpl.kt:58)                         
   CancellableContinuationImpl.<init>(CancellableContinuationImpl.kt:69)                         
   CancellableContinuationImpl.<init>(CancellableContinuationImpl.kt:71)                         
   SegmentQueueSynchronizer.suspend(SegmentQueueSynchronizer.kt:195)                             
   SegmentQueueSynchronizer.suspend(SegmentQueueSynchronizer.kt:196)                             
   Segment.getRemoved(ConcurrentLinkedList.kt:206)                                               
   SegmentQueueSynchronizer.suspend(SegmentQueueSynchronizer.kt:633)                             
   SegmentQueueSynchronizer.suspend(SegmentQueueSynchronizer.kt:642)                             
   SegmentQueueSynchronizer.suspend(SegmentQueueSynchronizer.kt:642)                             
   CancellableContinuationImpl.invokeOnCancellation(CancellableContinuationImpl.kt:500)          
   CancellableContinuationImpl.invokeOnCancellation(CancellableContinuationImpl.kt:269)          
   CancellableContinuationImpl.getState$kotlinx_coroutines_core(CancellableContinuationImpl.kt:76) 
   CancellableContinuationImpl.getParentHandle(CancellableContinuationImpl.kt:73)                
   CancellableContinuationImpl.trySuspend(CancellableContinuationImpl.kt:482)                    
   CancellableContinuationImpl.trySuspend(CancellableContinuationImpl.kt:210)           
```

alefedor:
Improve representation of default `toString` implementation.
Instead of `org.jetbrains.kotlinx.Node@6a03ad55` print just `Node@1`.
Instead of `[Lorg.jetbrains.kotlinx.Node;@6a03ad55` – `Node[]@1`

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -49,12 +52,18 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
         this.iterations = iterations
     }
 
+    /**
+     * Run each test scenario `invocations` times.
```
ndkoval, line 56: ... the specified number of times

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -49,12 +52,18 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
...
+     */
+    fun invocationsPerIteration(invocations: Int): OPT = applyAndCast {
```
ndkoval, line 58: Why should we make this strategy-related configuration common?

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -49,12 +52,18 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
...
+        this.invocationsPerIteration = invocations
+    }
+
     /**
      * Use the specified number of threads for the parallel part of an execution.
      *
-     *
      * Note, that the the actual number of threads can be less due to some restrictions
-     * like [Operation.runOnce].
+     * like [org.jetbrains.kotlinx.lincheck.annotations.Operation.runOnce].
```
ndkoval, line 66: Let's use imports

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -65,9 +74,8 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
     /**
      * Generate the specified number of operations for each thread of the parallel part of an execution.
      *
-     *
      * Note, that the the actual number of operations can be less due to some restrictions
-     * like [Operation.runOnce].
+     * like [org.jetbrains.kotlinx.lincheck.annotations.Operation.runOnce].
```
ndkoval, line 78: Let's use imports

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -78,9 +86,8 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
     /**
      * Generate the specified number of operation for the initial sequential part of an execution.
      *
-     *
      * Note, that the the actual number of operations can be less due to some restrictions
-     * like [Operation.runOnce].
+     * like [org.jetbrains.kotlinx.lincheck.annotations.Operation.runOnce].
```
ndkoval, line 90: Let's use imports

src/main/java/org/jetbrains/kotlinx/lincheck/Options.kt
```diff
@@ -91,9 +98,8 @@ abstract class Options<OPT : Options<OPT, CTEST>, CTEST : CTestConfiguration> {
     /**
      * Generate the specified number of operation for the last sequential part of an execution.
      *
-     *
      * Note, that the the actual number of operations can be less due to some restrictions
-     * like [Operation.runOnce].
+     * like [org.jetbrains.kotlinx.lincheck.annotations.Operation.runOnce].
```
ndkoval, line 102: Let's use imports

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -126,8 +134,70 @@ static boolean doNotTransform(String className) {
...
+     */
+    private String originalInternalName(String className) {
+        if (className.startsWith(TRANSFORMED_PACKAGE_INTERNAL_NAME))
+            return className.substring(TRANSFORMED_PACKAGE_INTERNAL_NAME.length());
+        return className;
+    }
+}
+
+/**
+ * Visitor for retrieving information of class version needed for choosing between COMPUTE_FRAME and COMPUTE_MAXS
```
ndkoval, line 187: Please explain here why do we need to make this choice

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -126,8 +134,70 @@ static boolean doNotTransform(String className) {
...
+    private final Remapper remapper;
+
+    public TransformationClassWriter(int classVersion, Remapper remapper) {
+        super(classVersion > V1_6 ? COMPUTE_FRAMES : COMPUTE_MAXS);
+        this.remapper = remapper;
+    }
+
+    /**
+     * ASM uses Class.forName for given types, however it can lead to cyclic dependencies when loading transformed classes.
+     * Thus, we call original method for not-transformed class names and then just fix it if needed.
```
ndkoval, line 166: **the** original method

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -126,8 +134,70 @@ static boolean doNotTransform(String className) {
...
+class TransformationClassWriter extends ClassWriter {
+    private final Remapper remapper;
+
+    public TransformationClassWriter(int classVersion, Remapper remapper) {
+        super(classVersion > V1_6 ? COMPUTE_FRAMES : COMPUTE_MAXS);
+        this.remapper = remapper;
+    }
+
+    /**
+     * ASM uses Class.forName for given types, however it can lead to cyclic dependencies when loading transformed classes.
```
ndkoval, line 165: put `Class.forName` in ``

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -126,8 +134,70 @@ static boolean doNotTransform(String className) {
         return cw.toByteArray();
     }
 
+    /**
+     * Returns name of class the moment before it was transformed
```
ndkoval, line 138: Do not understand the comment

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -22,62 +22,68 @@
...
  * class transformations and hines them from others.
  */
 public class TransformationClassLoader extends ExecutionClassLoader {
     private final List<Function<ClassVisitor, ClassVisitor>> classTransformers;
-    // Cache for classloading and frames computing during the transformation
+    // Cache for classloading and frames computing during the transformation.
     private final Map<String, Class<?>> cache = new ConcurrentHashMap<>();
+    private final Remapper remapper;
+
+    public static final String TRANSFORMED_PACKAGE_INTERNAL_NAME = "org/jetbrains/kotlinx/lincheck/tran$f*rmed/";
```
ndkoval, line 51: nice package name :)

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
```diff
@@ -24,37 +24,50 @@ package org.jetbrains.kotlinx.lincheck.strategy
...
-    is DeadlockInvocationResult -> DeadlockWithDumpFailure(scenario, threadDump)
-    is UnexpectedExceptionInvocationResult -> UnexpectedExceptionFailure(scenario, exception)
-    is ValidationFailureInvocationResult -> ValidationFailure(this.scenario, functionName, exception)
+internal class ObstructionFreedomViolationFailure(
+    scenario: ExecutionScenario,
+    val reason: String,
+    interleaving: List<InterleavingEvent>? = null
+) : LincheckFailure(scenario, interleaving)
+
+internal fun InvocationResult.toLincheckFailure(scenario: ExecutionScenario, execution: List<InterleavingEvent>? = null) = when (this) {
```
ndkoval, line 67: execution -> interleaving

src/test/java/org/jetbrains/kotlinx/lincheck/test/runner/ParallelThreadsRunnerExceptionTest.kt
```diff
@@ -108,7 +108,7 @@ class ParallelThreadsRunnerExceptionTest {
                 }
             }
         }
-        val runner = ParallelThreadsRunner(mockStrategy(scenario), testClass, emptyList(), null, DEFAULT_TIMEOUT_MS)
+        val runner = ParallelThreadsRunner(mockStrategy(scenario), testClass, emptyList(), null, null, DEFAULT_TIMEOUT_MS)
```
ndkoval, line 111: let's add parameter names

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/ExecutionReportingTest.kt
```diff
@@ -0,0 +1,94 @@
...
+package org.jetbrains.kotlinx.lincheck.test.strategy.modelchecking
+
+import org.jetbrains.kotlinx.lincheck.*
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.strategy.managed.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
+import org.jetbrains.kotlinx.lincheck.verifier.*
+import org.junit.*
+
+class ExecutionReportingTest : VerifierState() {
```
ndkoval, line 31: execution -> interleaving

ndkoval, line 31: Please explain what it tests and how

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/MethodReportingTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+import org.jetbrains.kotlinx.lincheck.*
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.strategy.managed.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
+import org.jetbrains.kotlinx.lincheck.verifier.*
+import org.junit.*
+import java.lang.StringBuilder
+
+class MethodReportingTest : VerifierState() {
```
ndkoval, line 32: Please explain what it tests and how

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/StateRepresentationTest.kt
```diff
@@ -0,0 +1,97 @@
...
+    @Operation
+    fun operation(): Int {
+        ++a
+        return inc()
+    }
+
+    private fun inc(): Int = ++a
+
+    override fun extractState(): Any = a
+
```
ndkoval, line 81: remove this empty line

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/StateRepresentationTest.kt
```diff
@@ -0,0 +1,97 @@
...
+import org.jetbrains.kotlinx.lincheck.strategy.managed.forClasses
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingOptions
+import org.jetbrains.kotlinx.lincheck.verifier.VerifierState
+import org.junit.Test
+import java.lang.IllegalStateException
+import java.lang.StringBuilder
+
+open class StateReportingTest : VerifierState() {
+    @Volatile
+    var a = 0
```
ndkoval, line 37: private

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/StateRepresentationTest.kt
```diff
@@ -0,0 +1,97 @@
...
+        check("STATE: 1" in log)
+        check("STATE: 2" in log)
+    }
+}
+
+class StateRepresentationInParentClassTest : StateReportingTest()
+
+class TwoStateRepresentationFunctionsTest : VerifierState() {
+    @Volatile
+    var a = 0
```
ndkoval, line 70: private

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/StateRepresentationTest.kt
```diff
@@ -0,0 +1,97 @@
...
+
+
+    @StateRepresentation
+    fun stateRepresentation1() = a.toString()
+
+    @StateRepresentation
+    fun stateRepresentation2() = a.toString()
+
+    @Test(expected = IllegalStateException::class)
+    fun test() {
```
ndkoval, line 90: = ModelCheckingOptions(). ... .check(..)

src/test/java/org/jetbrains/kotlinx/lincheck/test/verifier/linearizability/LockBasedSetTests.kt
```diff
@@ -40,6 +41,10 @@ abstract class AbstractSetTest(private val set: Set) : AbstractLincheckTest() {
     operator fun contains(@Param(name = "key") key: Int): Boolean = set.contains(key)
 
     override fun extractState(): Any = (1..5).map { set.contains(it) }
+
+    override fun <O : Options<O, *>> O.customize() {
+        iterations(5)
```
ndkoval, line 46: Why do we need this? Let's roll back

ndkoval (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/ManagedStateHolder.java
```diff
@@ -0,0 +1,71 @@
...
+    public static LocalObjectManager objectManager;
+    public static Random random;
+    private static final long initialSeed = 1337;
+
+    /**
+     * Sets the specified strategy and its initial state in the specified class loader.
+     */
+    public static void setState(ClassLoader loader, ManagedStrategy strategy) {
+        try {
+            Class<?> clazz = loader.loadClass(ManagedStateHolder.class.getCanonicalName());
```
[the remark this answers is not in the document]
alefedor: Its seems to me no, because using transformed code for `ManagedStateHolder` initialization is bug-prone as `ManagedStateHolder.strategy` is widely used by the transformer.

src/main/java/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
alefedor: I believe, it was an optimization for Java 6 or older classes due to the fact that these classes may not have valid stack frames. I consider removing this code

ndkoval (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -276,6 +295,23 @@ open class ParallelThreadsRunner(
...
     }
+
+    override fun getStateRepresentation(): String? {
+        stateRepresentation?.let { return getMethod(testInstance, it).invoke(testInstance) as String }
+        return null
+    }
+
+    override fun transformTestClass() {
+        super.transformTestClass()
+        initializeTestThreadExecutions()
```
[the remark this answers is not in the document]
alefedor: Because `TestThreadExecutions` depend on the class loader.
This method is invoked either after class loader initialization or after class loader change.
Better names for the method may be `onClassLoaderChange` or `retransformWithLoader`.
However, it seems to me that test thread executions are strongly connected to the test class, so it is ok to reinitialize them in method named `transformTestClass`

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -247,24 +245,47 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
...
                 if (scenarioContainsSuspendableActors) {
                     storeExceptionResultFromSuspendableThrowable(mv, resLocal, iLocal, iThread, i);
                 } else {
                     storeExceptionResultFromThrowable(mv, resLocal, iLocal);
                 }
-                mv.visitLabel(handlerEnd);
             }
+            // End of try-catch block for all other exceptions
+            mv.goTo(skipHandlers);
+
```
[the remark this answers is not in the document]
alefedor: I wanted an empty line here to make exception handlers visually distinguishable from the usual sequential code.
They can be reached only via non-trivial code paths (with goto), so it is deceptive when they just are after sequential code.
The same thing about all other handlers, I want to easily recognize their bounds too.

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -247,24 +245,47 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
             }
             // Store result to array
             mv.arrayStore(RESULT_TYPE);
-            // End of try-catch block
+
```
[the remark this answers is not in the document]
alefedor: see the comment about another empty line above

src/main/java/org/jetbrains/kotlinx/lincheck/runner/TestThreadExecutionGenerator.java
```diff
@@ -247,24 +245,47 @@ private static void generateRun(ClassVisitor cv, Type testType, int iThread, Lis
...
+            // End of try-catch block for handled exceptions
+            mv.visitLabel(actorCatchBlockEnd);
+            Label skipHandlers = mv.newLabel();
+            mv.goTo(skipHandlers);
+
```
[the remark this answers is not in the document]
alefedor: see the comment about another empty line above

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -39,5 +40,28 @@ abstract class Strategy protected constructor(
         throw UnsupportedOperationException("$javaClass strategy does not transform classes")
     }
 
+    /**
+     * Returns remapper that is used for transformation.
+     * null if there is no need in renamings.
+     */
+    open fun createRemapper(): Remapper? = null
```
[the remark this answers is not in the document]
alefedor: I think that it is better to do **no** remapping for the stress strategy.
The reason is that remapping is bug-prone and I wish to have more reliability for stress testing, which do not require any remapping.

I am pretty sure that in java standard library there exist a pair of classes, one of which is from `java.util` (and thus, is remapped) and the other is not with cyclic dependencies between them.
The only correct choices in this case are either to transform them both or to transform neither of the classes.
However, on the one hand there are classes that we can not or wish not to transform, while on the other hand we want to transform as many classes in `java.util` as possible in the model checking mode.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/InterleavingEvent.kt
```diff
@@ -0,0 +1,66 @@
...
+    ACTIVE_LOCK("active lock detected"),
+    SUSPENDED("coroutine is suspended"),
+    STRATEGY_SWITCH("");
+
+    override fun toString() = reason
+}
+
+/**
+ * Info about a [methodName] method call.
+ * [identifier] helps to distinguish two different calls of the same method.
```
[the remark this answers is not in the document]
alefedor: Improved comments

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectManager.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+package org.jetbrains.kotlinx.lincheck.strategy.managed
+
+import java.util.*
+
+/**
+ * Object manager holds information about some object characteristics.
+ * Tracks the creation of local objects and leaks of their references to non-local objects.
+ */
+internal class ObjectManager {
```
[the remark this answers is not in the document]
alefedor: It was LocalObjectManager, but it became just ObjectManager after it learned to store and load names for AFUs.
The name of an AFU is the name of the field it changes.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectManager.kt
```diff
@@ -0,0 +1,75 @@
...
+        if (ownerObjects != null) {
+            // actually save the dependency
+            ownerObjects.add(dependent)
+        } else {
+            // a link to the dependent leaked to a non-local object
+            deleteLocalObject(dependent)
+        }
+    }
+
+    fun registerObjectName(o: Any, name: String) {
```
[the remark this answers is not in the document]
alefedor: I do not want to have several ObjectManagers for different logic since we will need to have a field for each one in `ManagedStateHolder` and getter methods in `ManagedStrategyTransformer`. Nevertheless, the refactoring will make sense, if the logic becomes more difficult or more information about objects is needed.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectManager.kt
```diff
@@ -0,0 +1,75 @@
...
+import java.util.*
+
+/**
+ * Object manager holds information about some object characteristics.
+ * Tracks the creation of local objects and leaks of their references to non-local objects.
+ */
+internal class ObjectManager {
+    // for each local object store all objects that depend on it (e.g, are referenced by it)
+    private val localObjects = IdentityHashMap<Any, MutableList<Any>>()
+    // some objects can have a name associated to them
```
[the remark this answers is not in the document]
alefedor: In general case, it can be sometimes convenient to have labels for some objects, so that to use them in method call representation `objectLabel.call(parameter): result`.
Yet only AFUs have such labels (their names are the names of the fields they are associated with), but I feel it is a good idea to make the generalization here.

src/test/java/org/jetbrains/kotlinx/lincheck/test/strategy/modelchecking/ObstructionFreedomViolationTest.kt
```diff
@@ -0,0 +1,51 @@
...
+package org.jetbrains.kotlinx.lincheck.test.strategy.modelchecking
+
+import org.jetbrains.kotlinx.lincheck.*
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.strategy.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
+import org.jetbrains.kotlinx.lincheck.verifier.*
+import org.junit.*
+
+class ObstructionFreedomViolationTest : VerifierState() {
```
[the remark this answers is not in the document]
alefedor: That is strange, because I get
```
= Obstruction-freedom is required but a lock has been found =
Execution scenario (init part):
[incAndGet(), incAndGet(), get(), incAndGet(), incAndGet()]
Execution scenario (parallel part):
| get()       | incAndGet() |
| get()       | incAndGet() |
| incAndGet() | get()       |
| get()       | get()       |
| incAndGet() | incAndGet() |
Execution scenario (post part):
[incAndGet(), incAndGet(), get(), incAndGet(), get()]
= The following interleaving leads to the error =
Parallel part interleaving:
|                                                                                               | incAndGet()                                                                                         |
|                                                                                               |   MONITOR ENTER at ObstructionFreedomViolationTest.incAndGet(ObstructionFreedomViolationTest.kt:36)  |
|                                                                                               |   c.READ: 4 at ObstructionFreedomViolationTest.incAndGet(ObstructionFreedomViolationTest.kt:36)     |
|                                                                                               |   c.WRITE(5) at ObstructionFreedomViolationTest.incAndGet(ObstructionFreedomViolationTest.kt:36)    |
|                                                                                               |   switch                                                                                            |
| get()                                                                                         |                                                                                                     |
|   MONITOR ENTER at ObstructionFreedomViolationTest.get(ObstructionFreedomViolationTest.kt:39) |                                                                                                     |
|                                                                                               |   MONITOR EXIT at ObstructionFreedomViolationTest.incAndGet(ObstructionFreedomViolationTest.kt:36)  |

```

src/main/java/org/jetbrains/kotlinx/lincheck/execution/ExecutionResult.kt
```diff
@@ -35,22 +35,40 @@ data class ExecutionResult(
      * @see ExecutionScenario.initExecution
      */
     val initResults: List<Result>,
+    /**
+     * State representation at the end of the init part.
+     */
+    val afterInitState: String?,
```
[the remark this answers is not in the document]
alefedor: How do you want it to look like?
Currently the representation is
```
Init part:
[inc(), inc(), inc()]
STATE: 3
```
but with state representation in each quiescent point, should it look like
```
Init part:
inc()
STATE: 1
inc()
STATE: 2
inc()
STATE: 3
```
?

I am not sure whether it is better.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStateHolder.java
```diff
@@ -0,0 +1,73 @@
...
+import java.lang.reflect.*;
+import java.util.Random;
+
+import static org.jetbrains.kotlinx.lincheck.TransformationClassLoader.TRANSFORMED_PACKAGE_NAME;
+
+/**
+ * This utility class helps to hold the current strategy and its state. In order to run several tests in parallel,
+ * every iteration has to use its own class loader, thus this holder is not shared between them.
+ */
+public class ManagedStateHolder {
```
ndkoval, line 34: Let's use Kotlin here

ndkoval (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -276,6 +295,23 @@ open class ParallelThreadsRunner(
...
+
+    override fun constructStateRepresentation(): String? =
+        stateRepresentationMethod?.let{ getMethod(testInstance, it) }?.invoke(testInstance) as String?
+
+    override fun transformTestClass() {
+        super.transformTestClass()
+        initializeTestThreadExecutions()
+    }
+
+    private fun initializeTestThreadExecutions() {
```
ndkoval, line 307: Please, move the initialization back to the constructor, you do not need to re-create them.

ndkoval: However, it could be better to keep a method `createTestThreadExecutions` so that the code becomes cleaner

src/main/java/org/jetbrains/kotlinx/lincheck/runner/Runner.java
```diff
@@ -41,19 +42,52 @@
...
+    protected final Strategy strategy;
+    public ExecutionClassLoader classLoader;
 
     protected final AtomicInteger completedOrSuspendedThreads = new AtomicInteger(0);
 
-    protected Runner(Strategy strategy, Class<?> testClass, List<Method> validationFunctions) {
+    protected Runner(Strategy strategy, Class<?> testClass, List<Method> validationFunctions, Method stateRepresentationFunction) {
+        this.testClass = testClass;
+        this.strategy = strategy;
+        this.scenario = strategy.getScenario();
```
ndkoval, line 57: Let's use Kotlin here, such constructors make me cry :)
BTW, `val scenario get() = strategy.scenario`.

ndkoval (changes requested):

ndkoval (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/execution/ExecutionResult.kt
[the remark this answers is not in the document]
ndkoval: Hmm.. Probably, it is really better to keep the current state and improve trace representation by integrating with the IntelliJ IDEA debugger.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -35,9 +36,27 @@ abstract class Strategy protected constructor(
...
+
     abstract fun run(): LincheckFailure?
+
+    /**
+     * Check if Strategy allows to resume the coroutine.
+     */
+    open fun canResumeCoroutine(iThread: Int): Boolean {
+        return true
```
ndkoval, line 55: Why do you return `true` by default? Shouldn't this function be `abstract`?

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -35,9 +36,27 @@ abstract class Strategy protected constructor(
     val scenario: ExecutionScenario
 ) {
     open fun needsTransformation() = false
-    open fun createTransformer(cv: ClassVisitor?): ClassVisitor {
+    open fun createTransformer(cv: ClassVisitor): ClassVisitor {
```
ndkoval, line 39: It would be super-cool if we can somehow split the strategy logic itself with the JVM-related issues (such as code transformation) -- that would help a lot to make Lincheck multiplatform.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,863 @@
...
+    // logger of all events in the execution such as thread switches
+    private var eventCollector: InterleavingEventCollector? = null // null when `constructStateRepresentation` is false
+    // stack with info about method invocations in current stack trace for each thread
+    private val callStackTrace = Array(nThreads) { mutableListOf<CallStackTraceElement>() }
+    // an increasing id all method invocations
+    private var methodIdentifier = 0
+    // stack with info about suspended method invocations for each thread
+    private val suspendedMethodStack = Array(nThreads) { mutableListOf<Int>() }
+    // store previous created transformer to make code location ids different for different classes
+    private var previousTransformer: ManagedStrategyTransformer? = null
```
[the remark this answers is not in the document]
ndkoval: The next location id can be provided by the strategy itself (just create a special `IdProvider` interface and that's it)

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+        }
+        else -> result.toLincheckFailure(scenario, collectTrace(result))
+    }
+
+    /**
+     * Reruns previous invocation to log all its execution events.
+     */
+    private fun collectTrace(previousResults: InvocationResult): List<InterleavingEvent>? {
+        val detectedByStrategy = suddenInvocationResult != null
+        val canCollectInterleavingEvents = when {
```
ndkoval, line 194: canCollectTrace

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+            if (verifier.verifyResults(scenario, result.results)) null
+            else IncorrectResultsFailure(scenario, result.results, collectTrace(result))
+        }
+        else -> result.toLincheckFailure(scenario, collectTrace(result))
+    }
+
+    /**
+     * Reruns previous invocation to log all its execution events.
+     */
+    private fun collectTrace(previousResults: InvocationResult): List<InterleavingEvent>? {
```
ndkoval, line 192: previousResults -> failingResult

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+            else -> false
+        }
+
+        if (!canCollectInterleavingEvents) {
+            // interleaving events can be collected almost always,
+            // except for strange cases such as Runner timeout or LinChecker exceptions.
+            return null
+        }
+        // re-transform class constructing trace
+        constructTraceRepresentation = true
```
ndkoval, line 207: rename to `collectTrace`

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+        runner = createRunner()
```
ndkoval, line 208: Please explain here why we need a new runner

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+        initializeManagedState()
+        runner.initialize()
+        val loggedResults = runInvocation()
+        val sameResultTypes = loggedResults.javaClass == previousResults.javaClass
```
ndkoval, line 212: shouldn't you compare the results themselves, not only the types?

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+        // cannot check whether the results are exactly the same because of re-transformation
+        // so just check that types are the same
+        check(sameResultTypes) {
+            StringBuilder().apply {
+                appendln("Non-determinism found. Probably caused by non-deterministic code (WeakHashMap, Object.hashCode, etc).")
+                appendln("Reporting scenario without execution trace.")
+                appendln(loggedResults.asLincheckFailureWithoutTrace().toString())
+            }.toString()
+        }
+        return eventCollector!!.interleavingEvents
```
ndkoval, line 222: eventCollector -> traceCollector?

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+        if (this is CompletedInvocationResult)
+            return IncorrectResultsFailure(scenario, results, null)
+        return toLincheckFailure(scenario, null)
+    }
+
+    /**
+     * Logs thread events such as thread switches and passed code locations.
+     */
+    private inner class InterleavingEventCollector {
+        private val _interleavingEvents = mutableListOf<InterleavingEvent>()
```
ndkoval, line 660: interleavingEvents -> trace

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,864 @@
...
+    // code points for trace construction
+    private val interleavingPoints_USELESS_AND_SHOULD_BE_REMOVED: MutableList<InterleavingPoint> = ArrayList()
+    // logger of all events in the execution such as thread switches
+    private var eventCollector: InterleavingEventCollector? = null // null when `constructStateRepresentation` is false
+    // stack with info about method invocations in current stack trace for each thread
+    private val callStackTrace = Array(nThreads) { mutableListOf<CallStackTraceElement>() }
+    // an increasing id all method invocations
+    private var methodIdentifier = 0
+    // stack with info about suspended method invocations for each thread
+    private val suspendedMethodStack = Array(nThreads) { mutableListOf<Int>() }
```
ndkoval, line 97: let's explain in details why we need this information and how we use it

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
```diff
@@ -0,0 +1,271 @@
...
+package org.jetbrains.kotlinx.lincheck.strategy.managed
+
+import org.jetbrains.kotlinx.lincheck.*
+import org.jetbrains.kotlinx.lincheck.execution.ExecutionResult
+import org.jetbrains.kotlinx.lincheck.execution.ExecutionScenario
+import org.jetbrains.kotlinx.lincheck.execution.parallelResults
+import org.jetbrains.kotlinx.lincheck.printInColumnsCustom
+import kotlin.math.min
+
+private const val INTERLEAVING_INDENTATION = "  "
```
ndkoval, line 31: According to our last call, interleaving is a list of events that represents a certain execution, we use it for an interleaving tree, for example.
At the same time, the detailed and readable interleaving is called _trace_. Let's, therefore, rename all interleavingXXX-s to traceXXX-s here, since we work with a trace in this part of Lincheck.

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Construct an interleaving graph based on provided [interleavingEvents].
+ * Returns the node corresponding to the first interleaving event.
+ * An interleaving graph consists of two types of edges:
+ * `next` edges form a single-directed list in which the order of events is the same as in [interleavingEvents].
+ * `internalEvents` edges form a directed forest.
+ * The root of each tree is an actor.
+ * Its children are events and method invocations.
+ */
+private fun constructInterleavingGraph(scenario: ExecutionScenario, results: ExecutionResult?, interleavingEvents: List<InterleavingEvent>): InterleavingNode? {
```
ndkoval, line 81: Why is it called `graph`? It confuses a bit.

ndkoval (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
alefedor: Because graph theory is a great way to describe complex dependencies.
An interleaving with method calls is not a common object, but a graph is, and it is easier to visualize it as a graph

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -0,0 +1,863 @@
...
+    }
+
+    private fun createRunner(): Runner =
+        ManagedStrategyRunner(this, testClass, validationFunctions, stateRepresentationFunction, testCfg.timeoutMs, UseClocks.ALWAYS)
+
+    private fun initializeManagedState() {
+        ManagedStrategyStateHolder.setState(runner.classLoader, this)
+    }
+
+    override fun createTransformer(cv: ClassVisitor): ClassVisitor = ManagedStrategyTransformer(
```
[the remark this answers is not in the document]
alefedor: Thread safety is provided by `TransformationClassLoader.loadClass`. I can not think of a way to use this transformer other than for byte-code transformation in `TransformationClassLoader`

ndkoval (approved):

PR #46, opened 2020-08-17
Supported serializable classes as actor parameters and results

alefedor:


src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -126,12 +130,26 @@ private val methodsCache = WeakHashMap<Class<*>, WeakHashMap<Method, WeakReferen
...
     }
 }
 
+/**
+ * Finds a method corresponding to [name] and [parameterTypes] ignoring difference in loaders for [parameterTypes].
+ */
+private fun Class<out Any>.getMethod(name: String, parameterTypes: Array<Class<out Any>>): Method {
+    this.methods.forEach {
+        if (it.name == name && it.parameterTypes.size == parameterTypes.size) {
+            val sameParameters = parameterTypes.indices.all  { i -> it.parameterTypes[i].name == parameterTypes[i].name }
```
ndkoval, line 145:
```
= methods.findAny { m ->
  m.name == name && m.parameterTypes.map { it.name } == parameterTypes
} ?: error("Cannot find methods $m(....) in $name")
```

src/main/java/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -21,17 +21,20 @@
  */
 package org.jetbrains.kotlinx.lincheck
 
-import kotlinx.coroutines.*
+import kotlinx.coroutines.CancellableContinuation
 import org.jetbrains.kotlinx.lincheck.CancellableContinuationHolder.storedLastCancellableCont
-import org.jetbrains.kotlinx.lincheck.execution.*
-import org.jetbrains.kotlinx.lincheck.runner.*
-import org.jetbrains.kotlinx.lincheck.verifier.*
+import org.jetbrains.kotlinx.lincheck.execution.ExecutionResult
```
ndkoval, line 26: Please, fix your code style so that wildcards are always used

ndkoval (approved):

PR #47, opened 2020-09-02
Add a mechanism to provide the thread number as an operation parameter

ndkoval:


src/main/java/org/jetbrains/kotlinx/lincheck/execution/ActorGenerator.kt
```diff
@@ -0,0 +1,54 @@
...
+    cancellableOnSuspension: Boolean,
+    val allowExtraSuspension: Boolean
+) {
+    private val cancellableOnSuspension = cancellableOnSuspension && isSuspendable
+
+    fun generate(threadId: Int): Actor {
+        val parameters = parameterGenerators
+            .map { it.generate() }
+            .map { if (it === THREAD_ID_TOKEN) threadId else it }
+        val cancelOnSuspension = cancellableOnSuspension and nextBoolean()
```
alefedor, line 47: Shouldn't scenario generation be deterministic?

src/main/java/org/jetbrains/kotlinx/lincheck/execution/RandomExecutionGenerator.java
```diff
@@ -101,7 +101,7 @@ public ExecutionScenario nextExecution() {
                 leftActorGenerators.addAll(threadGen.nonParallelActorGenerators);
             for (int i = 0; i < testConfiguration.actorsAfter && !leftActorGenerators.isEmpty(); i++) {
                 ActorGenerator agen = getActorGenFromGroup(leftActorGenerators, random.nextInt(leftActorGenerators.size()));
-                postExecution.add(agen.generate());
+                postExecution.add(agen.generate(testConfiguration.threads));
```
alefedor, line 104: `testConfiguration.threads` -> `testConfiguration.threads + 1`

alefedor (changes requested):

PR #48, opened 2020-09-30
Cancelled operations can take effect and should not be removed from the history

ndkoval:


src/main/java/org/jetbrains/kotlinx/lincheck/verifier/LTS.kt
```diff
@@ -209,6 +224,8 @@ class LTS(sequentialSpecification: Class<*>) {
                 createLincheckResult(finalRes, wasSuspended = true)
             }
             CANCELLATION -> {
+                if (continuationsMap[Operation(this.actor, this.ticket, REQUEST)] == null)
```
SokolovaMaria, line 227: Remove this breakpoint

SokolovaMaria (approved):

PR #54, opened 2020-11-02
Move from Maven to Gradle

Krock21rus:


Krock21rus:
Вроде получилось добиться почти одинаковых jar файлов на выходе

mvn package vs ./gradlew jar

Screenshot from 2020-11-03 01-30-38
Screenshot from 2020-11-03 01-30-09

build.gradle.kts
```diff
@@ -109,4 +110,28 @@ publishing {
             artifact(tasks.getByName("javadocJar"))
```
Krock21rus, line 110: We need to add this line in new publications to bintray, because they are in infra.publishing

ndkoval (approved):
Great work!

PR #55, opened 2020-11-04
The obstruction-freedom check should ignore executions with uncompleted blocking operations

ndkoval:
Sometimes some operations on the testing data structures are blocking, but it would be great to still be able to test the rest ones for obstruction-freedom.

### DIFF
--- a/src/test/java/org/jetbrains/kotlinx/lincheck/test/BlockingOperationTest.kt
+++ b/src/test/java/org/jetbrains/kotlinx/lincheck/test/BlockingOperationTest.kt
@@ -1,3 +1,25 @@
+/*-
+ * #%L
+ * Lincheck
+ * %%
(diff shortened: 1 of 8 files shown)


src/test/java/org/jetbrains/kotlinx/lincheck/test/BlockingOperationTest.kt
```diff
@@ -0,0 +1,36 @@
+package org.jetbrains.kotlinx.lincheck.test
```
alefedor, line 1: Probably need license here

alefedor:
Everything is fine except for a bug at line 224 in Utils.kt.
There an `Actor` is created, but as we use default values and many arguments are Booleans, it did not break upon a new actor parameter addition.
So we need to use named arguments there to avoid the same bug in future and to add `actor.blocking` passing

alefedor (approved):

PR #56, opened 2020-11-05
The last event that caused obstruction freedom violation should be logged

alefedor:
It was unclear why obstruction freedom violation happened, because the event that caused it was not reported


src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -488,8 +492,8 @@ abstract class ManagedStrategy(
      */
     internal fun beforeWait(iThread: Int, codeLocation: Int, tracePoint: WaitTracePoint?, monitor: Any, withTimeout: Boolean): Boolean {
         if (!isTestThread(iThread)) return true
-        failIfObstructionFreedomIsRequired { "Obstruction-freedom is required but a waiting on monitor block has been found" }
         newSwitchPoint(iThread, codeLocation, tracePoint)
+        failIfObstructionFreedomIsRequired { "Obstruction-freedom is required but a waiting on monitor block has been found" }
```
ndkoval, line 496: on **a** monitor

ndkoval (approved):

PR #57, opened 2020-11-06
Add prompt cancellation support

ndkoval:


src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedCTestConfiguration.kt
```diff
@@ -49,6 +50,9 @@ abstract class ManagedCTestConfiguration(
             // Non-determinism should not be present in managed executions, but luckily the classes
             // can be just ignored, so that no thread context switches are added inside their methods.
             forClasses("kotlinx.coroutines.internal.StackTraceRecoveryKt").allMethods().ignore(),
+            // Ignore the DispatchedTask methods, they can be called by Lincheck.
+            @Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
+            forClasses(DispatchedTask::class.qualifiedName!!).allMethods().ignore(),
```
alefedor, line 55: That won't work.
`ManagedStrategyGuarantees` are supported only for methods that are called from the transformed code, because we add strategy invocation before and after the method's invocation.
See `ManagedStrategyTransformer.CancellabilitySupportMethodTransformer` to see how `CancellableContinuationImpl.cancel` из handled. We either need to handle `DispatchedTask` in the same way, or add a way to call `strategy.enterIgnoredSection` and `strategy.leaveIgnoredSection` directly in the `cancelByLincheck` method.

Note that if exceptions are possible, then try-catch block should be added in the transformed code

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -260,6 +260,7 @@ internal enum class SwitchReason(private val reason: String) {
  */
 internal class CallStackTraceElement(val call: MethodCallTracePoint, val identifier: Int)
 
+@Synchronized
```
alefedor, line 263: `objectNumeration` can be cleared concurrently in the `appendTrace` in `TraceReporter.kt`.
It is a bug and its simplest solution is to add `synchronized(objectNumeration)` for the whole `appendTrace` method.
We do not need additional synchronized here, because this method is called only from `appendTrace` (perhaps, we need a note about it?)

alefedor (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategyTransformer.kt
```diff
@@ -953,7 +953,9 @@ internal class ManagedStrategyTransformer(
      * when a continuation is cancelled by lincheck
      */
     private inner class CancellabilitySupportMethodTransformer(methodName: String, mv: GeneratorAdapter) : ManagedStrategyMethodVisitor(methodName, mv) {
-        private val isCancel = className == "kotlinx/coroutines/CancellableContinuationImpl" && methodName == "cancel"
+        private val isCancel =
+            className == "kotlinx/coroutines/CancellableContinuationImpl" && methodName == "cancel" ||
+            className == "kotlinx/coroutines/DispatchedTask" && methodName == "cancelCompletedResult"
```
alefedor, line 958: That still is not what you want.
`DispatchedTask.cancelCompletedResult` is an empty method, and adding ignored section here will not have any effect.
Probably you want to check whether `DispatchedTask` is a superclass of the `className`.

Another way is to remove `CancellabilitySupportMethodTransformer` at all, but add `strategy.enterIgnoredSection` and `strategy.leaveIgnoredSection` invocation in `cancelByLincheck` method

alefedor (changes requested):

src/main/java/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategyTransformer.kt
ndkoval: Guess, I should just check the same method but for `CancellableContinuationImpl`

alefedor (approved):

PR #62, opened 2020-11-18
Improved scenario generation diversity

alefedor:
Previously, different parameter generators with the same `Random` seeds were used for all unnamed parameters,
causing `fun operation(a: Int, b: Int, c: Int)` to always have `a == b == c`

### DIFF
--- a/src/main/java/org/jetbrains/kotlinx/lincheck/CTestStructure.java
+++ b/src/main/java/org/jetbrains/kotlinx/lincheck/CTestStructure.java
@@ -90,6 +90,8 @@ public class CTestStructure {
             }
             namedGens.put(paramAnn.name(), createGenerator(paramAnn));
         }
+        // Create map for default (not named) gens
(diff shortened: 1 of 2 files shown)


src/main/java/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -172,7 +174,7 @@ private static void readTestStructureFromClass(Class<?> clazz, Map<String, Param
             if (name != null)
                 return checkAndGetNamedGenerator(namedGens, name);
             // Parameter generator is not specified, try to create a default one
-            ParameterGenerator<?> defaultGenerator = createDefaultGenerator(p);
+            ParameterGenerator<?> defaultGenerator = createDefaultGenerator(p, defaultGens);
```
ndkoval, line 177: Let's create all default generators in advance and perform `defaultGens [p.getType()]!!` here

ndkoval (changes requested):

ndkoval (approved):

PR #67, opened 2020-12-02
Prompt cancellation should be emulated by cancelling the parent job

ndkoval:
Invoking `cancelCompletedResult` method is incorrect with `tryResume` -- it may lead to both completing the suspended coroutine and invoking all the cancellation handlers. See the unit test below that shows the problem:

```
@Test
fun testTryResumeOnCancellation() = runTest {
    // Start a new coroutine, suspend it, and store
    // the corresponding `CancellableContinuation` instance.
    var cont: CancellableContinuation<Unit>? = null
    launch {
        expect(2)
        suspendCancellableCoroutine<Unit> {
            cont = it
        }
        expectUnreached() // <-- IT IS REACHED ACTUALLY!!
    }
    expect(1)
    yield()
    expect(3)
    // Try to resume the launched coroutine.
    var onCancellationInvoked = false  // we expect it to become `true`
    val token = cont!!.tryResume(Unit, null) {
        onCancellationInvoked = true
    }
    assertNotNull(token) // assert that `tryResume` succeeded
    // Try to cancel the launched coroutine, should fail
    // due to the successful `tryResume` invocation.
    val cancelled = cont!!.cancel()
    assertFalse(cancelled)
    // However, `completeResume` has not been invoked yet,
    // so that the coroutine is not dispatched.
    // Let's then use prompt cancellation here.
    (cont as DispatchedTask<*>).cancelCompletedResult(null, Exception())
    // Complete the resumption. However, the coroutine should be cancelled due to
    // `cancelCompletedResult` invocation, so that `onCancellation` lambda should
    // be invoked and the coroutine execution should be aborted.
    cont!!.completeResume(token)
    assertTrue(onCancellationInvoked) // coroutine is cancelled while dispatching => should be invoked
    yield() // should not take any effect, but fails the test if the coroutine completes the execution
    finish(4)
}
```

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
@@ -198,8 +198,7 @@ fun <T> CancellableContinuation<T>.cancelByLincheck(promptCancellation: Boolean)
         throw it.cause!! // let's throw the original exception, ignoring the internal coroutines details
     }
     if (!cancelled && promptCancellation) {
-        // TODO we can invoke the method directly if the transformation is disabled
-        getMethod(this, cancelCompletedResultMethod).invoke(this, null, cancellationByLincheckException)
+        context[Job]!!.cancel() // we should always put a job into the context for prompt cancellation
         return true
     }
     return cancelled
--- a/src/jvm/test/org/jetbrains/kotlinx/lincheck/test/PromptCancellationTest.kt
+++ b/src/jvm/test/org/jetbrains/kotlinx/lincheck/test/PromptCancellationTest.kt
@@ -20,11 +20,13 @@
 
 package org.jetbrains.kotlinx.lincheck.test
 
+import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import org.jetbrains.kotlinx.lincheck.*
 import org.jetbrains.kotlinx.lincheck.annotations.*
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.paramgen.*
 import org.jetbrains.kotlinx.lincheck.strategy.*
-import org.jetbrains.kotlinx.lincheck.strategy.stress.*
 import kotlin.coroutines.*
 import kotlin.reflect.*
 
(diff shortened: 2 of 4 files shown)


src/jvm/test/org/jetbrains/kotlinx/lincheck/test/PromptCancellationTest.kt
```diff
@@ -33,22 +35,41 @@ abstract class AbstractPromptCancellationTest(
...
         }
+        check(completedOrCancelled.compareAndSet(false, true))
     }
 
     @ExperimentalCoroutinesApi
+    @InternalCoroutinesApi
     @Operation(runOnce = true)
-    fun resumeOp(): Int {
-        cont?.resume(Unit, { returnResult = 42 })
+    fun resumeOp(@Param(gen = IntGen::class, conf = "1:2") mode: Int): Int {
```
alefedor, line 56: This prompt cancellation test does not have cancellation...
The incorrect results I get are 
```
| suspendOp(): S | resumeOp(2): 0 |
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -198,8 +198,7 @@ fun <T> CancellableContinuation<T>.cancelByLincheck(promptCancellation: Boolean)
         throw it.cause!! // let's throw the original exception, ignoring the internal coroutines details
     }
     if (!cancelled && promptCancellation) {
-        // TODO we can invoke the method directly if the transformation is disabled
-        getMethod(this, cancelCompletedResultMethod).invoke(this, null, cancellationByLincheckException)
+        context[Job]!!.cancel() // we should always put a job into the context for prompt cancellation
```
alefedor, line 201: As far as I now `CancellableContinuationImpl.cancel` does not call `cancelCompletedResult`, so I am not sure that this change is correct...

alefedor (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
ndkoval: Here we cancel the parent job (not the continuation!) to cancel the resumed but not dispatched continuation -- that is how prompt cancellation works. I had a conversation with Vsevolod Tolstopyatov on this.

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/PromptCancellationTest.kt
ndkoval: How did you get them?

ndkoval: It does have cancellation -- `suspendOp` is cancellable with prompt cancellation.

alefedor: > How did you get them?

Just removed `IncorrectResultsFailure` from class constructor

alefedor: > It does have cancellation -- suspendOp is cancellable with prompt cancellation.
It does, but it is strange that `IncorrectPromptCancellationTest` does not have prompt cancellation in the incorrectness

alefedor (approved):

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/PromptCancellationTest.kt
ndkoval: that was a bug in the test, I fixed it

PR #74, opened 2021-02-27
Fix incorrect link

okue:
Changes:

- `Default generators` -> `Parameter generators`
- `binding-parameter-and-generator-names` -> `#binding-parameter-and-generator-names`


ndkoval:
Thanks for the fix!

PR #75, opened 2021-03-01
enable GitHub Actions CI

sullis:


sullis:
WDYT? @ndkoval

ndkoval:
@sullis thanks for the PR!

We now use Teamcity for continuous integration, and it is not clear to me why we need to maintain GitHub Actions in addition to it.

PR #79, opened 2021-04-18
Model checking implementation for NVM

zuevmaxim:


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -196,7 +211,7 @@ internal class ModelCheckingStrategy(
      */
     private inner class SwitchChoosingNode : InterleavingTreeNode() {
         override fun nextInterleaving(interleavingBuilder: InterleavingBuilder): Interleaving {
-            val isLeaf = maxNumberOfSwitches == interleavingBuilder.numberOfSwitches
+            val isLeaf = maxNumberOfSwitches == interleavingBuilder.numberOfCrashes + interleavingBuilder.numberOfSwitches
```
alefedor, line 214: Probably, instead of `numberOfCrashes` or `numberOfSwitches` we should be just `numberOfEvents`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -211,27 +226,68 @@ internal class ModelCheckingStrategy(
...
+        private val crashPositions: List<Int>,
         private val threadSwitchChoices: List<Int>,
-        private var lastNotInitializedNode: SwitchChoosingNode?
+        private val switchOrCrashChoices: List<Int>,
+        private var lastNotInitializedNode: InterleavingTreeNode?
     ) {
         private lateinit var interleavingFinishingRandom: Random
         private lateinit var nextThreadToSwitch: Iterator<Int>
+        private lateinit var nextCrashSwitchChoice: Iterator<Int>
+        private var currentSwitchCrashChoice: Int = -1
```
alefedor, line 275: I guess, this logic can be simplified.
You can check `isSwitchPosition` and `isCrashPosition` without comparing event type choices, because every execution position implicitly corresponds to a possible crash or switch. Every time `isSwitchPosition` or `isCrashPosition` is called, the execution position is increased, so there are no collisions.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -256,28 +313,37 @@ internal class ModelCheckingStrategy(
...
          * which helps to distinguish different switch points.
          */
-        fun newExecutionPosition(iThread: Int) {
+        private fun newExecutionPosition(iThread: Int, isSwitch: Boolean) {
+            if (exploreCrashNode == isSwitch) return
             executionPosition++
-            if (executionPosition > switchPositions.lastOrNull() ?: -1) {
+            if (executionPosition > positions.lastOrNull() ?: -1) {
                 // Add a new thread choosing node corresponding to the switch at the current execution position.
                 lastNotInitializedNodeChoices?.add(Choice(ThreadChoosingNode(switchableThreads(iThread)), executionPosition))
```
alefedor, line 329: Is it the intended logic that immediately after every crash there is a thread switch?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -256,28 +313,37 @@ internal class ModelCheckingStrategy(
...
 
         /**
-         * Creates a new execution position that corresponds to the current switch point.
+         * Creates a new execution position that corresponds to the current switch/crash point.
          * Unlike switch points, the execution position is just a gradually increasing counter
          * which helps to distinguish different switch points.
          */
-        fun newExecutionPosition(iThread: Int) {
+        private fun newExecutionPosition(iThread: Int, isSwitch: Boolean) {
+            if (exploreCrashNode == isSwitch) return
```
alefedor, line 325: This seems like a bug. Whether or not nodes in interleaving tree are initialized should not change execution position numbering, otherwise wrong positions will be saved in a node initialization

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -123,15 +125,19 @@ abstract class ManagedStrategy(
...
-        cv = cv,
-        tracePointConstructors = tracePointConstructors,
-        guarantees = testCfg.guarantees,
-        eliminateLocalObjects = testCfg.eliminateLocalObjects,
-        collectStateRepresentation = collectStateRepresentation,
-        constructTraceRepresentation = collectTrace,
-        codeLocationIdProvider = codeLocationIdProvider
-    )
+    override fun createTransformer(cv: ClassVisitor): ClassVisitor {
+        val visitor = CrashEnabledVisitor(cv, testClass, recoverModel.crashes)
```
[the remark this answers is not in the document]
zuevmaxim: I use this class visitor to reuse code that checks if crashes are allowed for this class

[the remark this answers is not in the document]
alefedor: Why not to transform every class that can be transformed?
Internal lincheck classes are already filtered by `TransformationClassLoader`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
zuevmaxim: No, I just want to unify the creation of crash/switch choices. The plan here was that the position array corresponds to lastNotInitializedNodeChoices of the same type(crash or switch).

zuevmaxim: I'll add the logic of immediate switch. I had a hope that Lincheck can deal with it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
zuevmaxim: Then the only point is filtering `@CrashFree` annotated classes. This annotation was introduced for gradual crashes enabling.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
zuevmaxim: I've rearranged the lines, so increment is always executed now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
zuevmaxim: Also, we do not transform testClass as it requires to mark a method with `@Recoverable`, which is not convenient when delegation is used.

zuevmaxim:
It works!

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/verifier/durable/detectable/LogQueueTest.kt
```diff
@@ -193,8 +193,8 @@ internal class LogQueue(threadsNumber: Int) : RecoverableQueue {
     }
 }
 
-internal abstract class LogQueueFailingTest :
-    AbstractNVMLincheckFailingTest(Recover.DETECTABLE_EXECUTION, THREADS_NUMBER, SequentialQueue::class) {
+internal abstract class LogQueueFailingTest //: AbstractNVMLincheckFailingTest(Recover.DETECTABLE_EXECUTION, THREADS_NUMBER, SequentialQueue::class)
```
alefedor, line 196: Is it correct?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -257,27 +307,34 @@ internal class ModelCheckingStrategy(
...
+         * Creates a new execution position that corresponds to the current switch/crash point.
          * Unlike switch points, the execution position is just a gradually increasing counter
          * which helps to distinguish different switch points.
          */
-        fun newExecutionPosition(iThread: Int) {
+        private fun newExecutionPosition(iThread: Int, isSwitch: Boolean) {
             executionPosition++
-            if (executionPosition > switchPositions.lastOrNull() ?: -1) {
+            if (exploreCrashNode == isSwitch) return
+            if (executionPosition > positions.lastOrNull() ?: -1) {
```
alefedor, line 320: Instead of creating `positions`, you can compare here with the last execution event position (maximum of `crashPositions.last()` and `switchPositions.last()`). Probably this logic can be extracted to `lastChosenExecutionPosition()` method.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -257,27 +307,34 @@ internal class ModelCheckingStrategy(
...
-         * Creates a new execution position that corresponds to the current switch point.
+         * Creates a new execution position that corresponds to the current switch/crash point.
          * Unlike switch points, the execution position is just a gradually increasing counter
          * which helps to distinguish different switch points.
          */
-        fun newExecutionPosition(iThread: Int) {
+        private fun newExecutionPosition(iThread: Int, isSwitch: Boolean) {
             executionPosition++
-            if (executionPosition > switchPositions.lastOrNull() ?: -1) {
+            if (exploreCrashNode == isSwitch) return
```
alefedor, line 319: Lets add `enum ExplorationNodeType` and use it, otherwise this boolean logic seems sophisticated

alefedor (changes requested):
Congratulations that you have something working!
Yet, not every test pass.
Besides, there are some problems in the output:
1) Strange
```
Crashes on thread 1:
Crashed inside actor 1:
	.()
```
2) ```: threw CrashErrorProxy```
See https://github.com/Kotlin/kotlinx-lincheck/blob/ce28a0cc79e6e30e94bd0ac0b4f266716394ce9d/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt#L149
3) No `CRASH` messages, because you do not add `CrashTracePoint` to the trace

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/verifier/durable/detectable/LogQueueTest.kt
zuevmaxim: Log queue algorithm is not ready for now, so it is muted.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
zuevmaxim: `positions` does not create a copy of a list, it is just a link. 
I think comparing with maximum does not work. For example, we are exploring a crash node and our crashPositions list is empty while switchPositions is [42]. We are visiting the current crash position 33, so we want to add it while it is smaller than 42.

alefedor: If `crashPositions` is empty and `switchPositions` is not, you are interested only in crash positions that happen after the last event (switch). Adding a crash before the last switch can break the switch choice (e.g., when a thread you chose to switch to becomes no longer available)

zuevmaxim: You are right. I fixed it

src/jvm/main/org/jetbrains/kotlinx/lincheck/nvm/crash.kt
```diff
@@ -139,18 +142,22 @@ object Crash {
...
+private class BusyWaitingBarrier(private val yieldCallback: () -> Unit) {
     private val free = atomic(false)
     private val waitingCount = atomic(0)
 
     inline fun await(action: (Boolean) -> Unit) {
         waitingCount.incrementAndGet()
         // wait for all to access the barrier
-        while (waitingCount.value < Crash.threads && !free.value);
+        while (waitingCount.value < Crash.threads && !free.value) {
+            yieldCallback()
```
alefedor, line 153: Due to this callback won't you cover some odd interleavings like "switch from thread 1 to thread 2 because of a system crash, switch back to thread 1, then switch again from thread 1, because not all threads reached this barrier"?

zuevmaxim: It is true. I've fixed it by force manual switch

alefedor (commented):
Great work!
Only a few issues with incorrect execution appearance.
1) "crash" is difficult to see visually.
Maybe "CRASH" is better?
2) We need to somehow mark recovery method invocation.
For example, add "RECOVERY" after crash message
3) Callstack for recovery seems strange.
From `CounterFailingTest3`:
```
Parallel part trace:
| increment(1): void                                                                         |                                                                                    |                                                                                            |
| get(1)                                                                                     |                                                                                    |                                                                                            |
|   get(1): 5 at CounterFailingTest.get(CounterTest.kt:99)                                   |                                                                                    |                                                                                            |
|     iterator(): Itr@1 at NRLFailingCounter1.get(CounterTest.kt:247)                        |                                                                                    |                                                                                            |
|     hasNext(): true at NRLFailingCounter1.get(CounterTest.kt:247)                          |                                                                                    |                                                                                            |
|     next(): NRLReadWriteObject@1 at NRLFailingCounter1.get(CounterTest.kt:247)             |                                                                                    |                                                                                            |
|     read(): 4 at NRLFailingCounter1.get(CounterTest.kt:127)                                |                                                                                    |                                                                                            |
.....
|     next() at NRLFailingCounter1.get(CounterTest.kt:247)                                   |                                                                                    |                                                                                            |
|       checkForComodification() at ArrayList$Itr.next(ArrayList.java:997)                   |                                                                                    |                                                                                            |
|       size.READ: 5 at ArrayList$Itr.next(ArrayList.java:999)                               |                                                                                    |                                                                                            |
|       elementData.READ: Object[]@1 at ArrayList$Itr.next(ArrayList.java:1001)              |                                                                                    |                                                                                            |
|       READ: NRLReadWriteObject@5 at ArrayList$Itr.next(ArrayList.java:1005)                |                                                                                    |                                                                                            |
|       crash                                                                                |                                                                                    |                                                                                            |
|       iterator(): Itr@2 at NRLFailingCounter1.get(CounterTest.kt:247)                      |                                                                                    |                                                                                            |
|       hasNext(): true at NRLFailingCounter1.get(CounterTest.kt:247)                        |                                                                                    |                                                                                            |
|       next(): NRLReadWriteObject@1 at NRLFailingCounter1.get(CounterTest.kt:247)           |                                                                                    |                                                                                            |
|       read(): 4 at NRLFailingCounter1.get(CounterTest.kt:127)                              |                                                                                    |                                                                                            |
```
Here it seems that the recovery is called in `ArrayList$Itr.next`, which is not intuitive, because after crash we expect call stack to be cleared and recovery to start from scratch

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -305,6 +331,40 @@ abstract class ManagedStrategy(
...
+        if (iThread != systemCrashInitiator) {
+            currentThread = systemCrashInitiator
+            awaitTurn(iThread)
+        } else {
+            for (t in switchableThreads(iThread)) {
+                currentThread = t
+                awaitTurn(iThread)
+            }
+            Crash.onSystemCrash()
+            systemCrashInitiator = -1
```
alefedor, line 362: Better have a constant like `NO_CRASH_INITIATOR`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -222,7 +245,9 @@ abstract class ManagedStrategy(
         runner.initialize()
         val loggedResults = runInvocation()
         val sameResultTypes = loggedResults.javaClass == failingResult.javaClass
-        val sameResults = loggedResults !is CompletedInvocationResult || failingResult !is CompletedInvocationResult || loggedResults.results == failingResult.results
+        val sameResults = loggedResults !is CompletedInvocationResult ||
+            failingResult !is CompletedInvocationResult ||
+            loggedResults.results.withoutCrashes == failingResult.results.withoutCrashes
```
alefedor, line 250: Should we not check that crashes are the same?

alefedor (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
zuevmaxim: The problem here is that during rerunning the failed interleaving the trace is collected, so crash exceptions are filled with stacktace info, so crashes in these two invocations always differ. I can disable stacktrace collection at all in MC mode as we have all trace collected, then crashes will contain no info => will be equal.

zuevmaxim: Sorry, this is not the reason. I'll fix it.

PR #103, opened 2021-07-04
Extract model checking for crashes

zuevmaxim:


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -809,13 +733,24 @@ abstract class ManagedStrategy(
             _trace += StateRepresentationTracePoint(iThread, currentActorId[iThread], stateRepresentation, callStackTrace)
         }
     }
+
+
+    // == NVM related ==
+
+    internal fun beforeCrashPoint(iThread: Int) = newCrashPoint(iThread)
+    internal fun beforeNVMOperation(iThread: Int, codeLocation: Int, tracePoint: MethodCallTracePoint?) =
```
alefedor, line 741: Could you please add comments for these methods?
In some sense they are internal api for transformers, so it is useful to know, for example, whether `NVM operation` is just read/write from a shared location or it can be something else

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingCTestConfiguration.kt
```diff
@@ -41,6 +41,9 @@ class ModelCheckingCTestConfiguration(testClass: Class<*>, iterations: Int, thre
     checkObstructionFreedom, hangingDetectionThreshold, invocationsPerIteration, guarantees, requireStateEquivalenceCheck,
     minimizeFailedScenario, sequentialSpecification, timeoutMs, eliminateLocalObjects, verboseTrace, customScenarios, recoverabilityModel) {
     override fun createStrategy(testClass: Class<*>, scenario: ExecutionScenario, validationFunctions: List<Method>,
-                                stateRepresentationMethod: Method?, verifier: Verifier): Strategy
-        = ModelCheckingStrategy(this, testClass, scenario, validationFunctions, stateRepresentationMethod, verifier, recoverabilityModel)
+                                stateRepresentationMethod: Method?, verifier: Verifier): Strategy = if (recoverabilityModel.crashes) {
```
alefedor, line 44: Probably, this code is better readable when `if` is on the next line and there are no brackets.
Usually you write `val map = if (sorted) TreeMap(...) else HashMap(..)` and not `if (sorted) { TreeMap(...) } else { HashMap(...) }`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -27,9 +27,17 @@ import org.jetbrains.kotlinx.lincheck.strategy.*
 import org.jetbrains.kotlinx.lincheck.strategy.managed.*
 import org.jetbrains.kotlinx.lincheck.verifier.*
 import java.lang.reflect.*
-import kotlin.math.max
 import kotlin.random.*
 
+internal interface Interleaving {
```
alefedor, line 32: Do we need this interface when you almost always use inheritors of `AbstractSwitchesInterleaving`?
If not, lets get rid of it and simplify `AbstractSwitchesInterleaving` name to `SwitchestInterleaving`.
Similarly, `AbstractSwitchesInterleavingBuilder` -> `SwitchesInterleavingBuilder`.
Abstract `InterleavingTreeNode` does not have `Abstract`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/SwitchesAndCrashesModelCheckingStrategy.kt
```diff
@@ -0,0 +1,372 @@
...
+            fun fromNode(node: AbstractModelCheckingStrategy<*, *>.InterleavingTreeNode?) = when (node) {
+                is SwitchChoosingNode -> SWITCH
+                is CrashChoosingNode -> CRASH
+                else -> NONE
+            }
+        }
+    }
+}
+
+private class CrashesManagedStrategyTransformer(
```
alefedor, line 272: This file is too large => lets move to another file

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -367,27 +308,32 @@ internal class ModelCheckingStrategy(
...
-        SWITCH, CRASH, NONE;
-
-        companion object {
-            fun fromNode(node: InterleavingTreeNode?) = when (node) {
-                is SwitchChoosingNode -> SWITCH
-                is CrashChoosingNode -> CRASH
-                else -> NONE
-            }
-        }
+internal class ModelCheckingStrategy(
```
alefedor, line 317: Lets separate these classes (`ModelCheckingStrategy` and `AbstractModelCheckingStrategy`) into different files

alefedor (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
zuevmaxim: We will lose all Git Blame(If I name files ModelCheckingStrategy.kt and AbstractModelCheckingStrategy.kt respectively). Is it OK?

alefedor: I think it's alright, the git blame of the file is not very helpful

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
zuevmaxim: > Abstract InterleavingTreeNode does not have Abstract
I think InterleavingTreeNode is shorter and looks better. Having so many concrete inheritors InterleavingTreeNode looks like something common, so it is clear that it is abstract. No?

zuevmaxim: I cannot get rid of ModelCheckingStrategy.SwitchesInterleaving as I have to specify a type parameter for super class. The same with builder - a concrete result of build can be specified only in inheritor. In this case, I think basic interfaces make understanding type parameters easier.

If you have some ideas how to make a simpler compilable version of this code, could we discuss it together?

alefedor: > I think InterleavingTreeNode is shorter and looks better.
That's my point. I just showed that not all abstract classes have `Abstract` and suggested to shorten these names

> I cannot get rid of ModelCheckingStrategy.SwitchesInterleaving
But I think you can get rid of `AbstractSwitchesInterleaving`

zuevmaxim: Ok, I got it. How about revert names back to simple Interleaving and InterleavingBuilder?

alefedor: Yeah, sounds good

PR #29, opened 2019-10-23
Libtest tests fixed

alefedor:
Fixed a problem with not all tests being executed.
Improved tests.
Added ```AbstractLinCheckTest``` for testing all LinCheck strategies

PR #111, opened 2021-08-21
Iteration skipping

alefedor:
* Iteration number is now always present in the incorrect execution log.
* Iterations can be skipped via `Options.skipIterations(iterations)` to reproduce errors faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ Table of contents
     + [Operation groups](#operation-groups)
   * [Parameter generators](#parameter-generators)
     + [Binding parameter and generator names](#binding-parameter-and-generator-names)
-  * [Custom scenarios](#custom-scenarios)
   * [Sequential specification](#sequential-specification)
   * [Validation functions](#validation-functions)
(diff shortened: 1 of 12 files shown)


ndkoval:
Hi @alefedor! 

Sorry for the long reply; I managed to take a look at the PR only now. 

The suggestion is excellent, but I would slightly replace the API. As you need to reproduce the error by running a specific iteration and only it, I suggest adding an option `runSpecificIteration` (let's think about the better naming later), which runs only this particular iteration, skipping all the previous and further ones. 

I would also embed the iteration number in the error title (e.g., printing `= Invalid execution results = (iteration #29/100)`). However, it does not matter a lot, and we can change this at any moment later.

alefedor:
Hi @ndkoval ,

Yes, it makes sense to run only one specific iteration

ndkoval:
We have decided to implement #139 instead, so I am closing the PR.

@avpotapov00, you may be interested in the tests.

PR #115, opened 2022-01-30
Fix test name for StressOptionsTest in README

dapengzhang0:


ndkoval:
I would prefer not to rename the test name as `StressOptionsTest` means that it tests `StressOptions`.

alefedor:
Yeah, it seems like names in the code example are not consistent.
Either `MyConcurrentTest` -> `StressOptionsTest` or vice versa

Issue #124, opened 2022-11-02
memory leak resulting in hung threads


btwilk:
Version: 2.15
I'm model checking a fairly complex piece of code and found that several iterations complete successfully but eventually lincheck detects hung threads.

Increasing heap size allows more iterations to succeed. This made me suspect a memory leak across iterations. I analyzed some heap dumps and found that most memory was consumed by ModelCheckingStrategy instances which contain huge (100s of MB) trees of potential context switch choices. It seems that each iteration uses a fresh ModelCheckingStrategy instance but something is holding on to references to the instances from completed iterations so that they are not GC'd.

E.g. when I take a snapshot during iteration 3, I see 3 instances of ModelCheckingStrategy in the dump despite there having been full GCs.

btwilk:
I know I inundated you recently with a lot of bug reports, but I wanted to call out that this one is the most debilitating. I'm running with a 48GB heap and still can only do 10-25 iterations depending on the test, and I need to fork JVMs for each test class as the memory leaks pile up across test classes too.

Wondering if you've found the cause of the leak.

alefedor:
Hi @btwilk !

Thank you for your valuable bug reports :)  
Yeah, a memory leak can result in hung threads report.  

I've checked tests for `java.util.concurrent.ConcurrentHashMap`/`ConcurrentDeque`/`ConcurrentSkipList` and `jctools.NonBlockingHashMapLong` but was not able to observe more than one `ModelCheckingStrategy` instance. 

I know there something like a memory leak in `LinearizabilityVerifier` because it caches scenario results and keeps this cache between iterations, however, from your description, this is not your case. Without this leak, the memory consumption graphs seem healthy to me.

So, can you please provide some insight into what kind of tests you are running?  
In particular, I am interested in the following:
- Do operations in your tests have parameters or results of non-primitive types (i.e., actual objects, not `Int`, `Long`, `Bool`, etc)?
- How big are the scenarios you generate? The number of interleavings and valid results grows exponentially with the number of threads and operations, so usually no more than `3` threads and `3` operations in each thread are used in tests (and we do not know an real example when a bug requires more threads/operations)

btwilk:
Answering your questions:
* I currently have non-primitive parameters but looking back at my commit history, I had already reduced the iteration count to 25 with only primitive parameters.
* I happen to have also landed on 3 threads with 3 operations as being a sweet spot.

[...]

PR #128, opened 2022-11-17
Fix Verifier memory leak

alefedor:
Fixed a memory leak between iteration in Lincheck.

The LinCheck Verifier caches scenario results which it already verified to bound computations.
However, this cache was kept between different iterations, essentially resulting in a memory leak.
The cached values from previous iterations can be useful only if the scenario generated in the current iteration is the same as in one of the previous iterations, which is an unlikely event in practice.

The size of the leak can be large for tests of reasonable size. I managed to get a 500Mb leak for a configuration with 3 threads, 2 operations per thread, 1000 invocations per iterations and 500 iterations.

This PR re-creates linearizability verifier for each iteration to avoid this leak

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
@@ -64,15 +64,21 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
 
     private fun CTestConfiguration.checkImpl(): LincheckFailure? {
         val exGen = createExecutionGenerator()
-        val verifier = createVerifier(checkStateEquivalence = true)
         for (i in customScenarios.indices) {
+            val verifier = createVerifier(checkStateEquivalence = i == 0)
             val scenario = customScenarios[i]
             scenario.validate()
             reporter.logIteration(i + 1, customScenarios.size, scenario)
             val failure = scenario.run(this, verifier)
             if (failure != null) return failure
         }
+        var verifier = createVerifier(checkStateEquivalence = true)
         repeat(iterations) { i ->
+            // If verifier is too old or not enough free memory, create a new one to avoid memory leaks.
+            // Otherwise, the memory leak can cause OutOfMemoryError, especially when non-primitive parameters or results are used.
+            // https://github.com/Kotlin/kotlinx-lincheck/issues/124
+            if ((i + 1) % VERIFIER_REFRESH_CYCLE == 0 || lacksFreeMemory())
+                verifier = createVerifier(checkStateEquivalence = false)
             val scenario = exGen.nextExecution()
             scenario.validate()
             reporter.logIteration(i + 1 + customScenarios.size, iterations, scenario)
@@ -85,6 +91,12 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
             }
         }
         return null
+    }
+
+    private fun lacksFreeMemory(): Boolean {
+        val runtime = Runtime.getRuntime()
+        // Check whether less than 10% free memory is left
+        return runtime.freeMemory() < 0.1 * runtime.maxMemory()
     }
 
     // Tries to minimize the specified failing scenario to make the error easier to understand.
@@ -209,6 +221,8 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         fun check(testClass: Class<*>, options: Options<*, *>? = null) {
             LinChecker(testClass, options).check()
         }
+
+        private const val VERIFIER_REFRESH_CYCLE = 100
     }
 }
 
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
@@ -74,10 +74,12 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         }
         var verifier = createVerifier(checkStateEquivalence = true)
         repeat(iterations) { i ->
-            // If verifier is too old or not enough free memory, create a new one to avoid memory leaks.
-            // Otherwise, the memory leak can cause OutOfMemoryError, especially when non-primitive parameters or results are used.
+            // For performance reasons, verifier re-uses LTS from previous iterations.
+            // This behaviour is similar to a memory leak and can potentially cause OutOfMemoryError.
+            // This is why we periodically create a new verifier to still have increased performance
+            // from re-using LTS and limit the size of potential memory leak.
             // https://github.com/Kotlin/kotlinx-lincheck/issues/124
-            if ((i + 1) % VERIFIER_REFRESH_CYCLE == 0 || lacksFreeMemory())
+            if ((i + 1) % VERIFIER_REFRESH_CYCLE == 0)
                 verifier = createVerifier(checkStateEquivalence = false)
             val scenario = exGen.nextExecution()
             scenario.validate()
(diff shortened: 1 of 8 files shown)


ndkoval:
Hi @alefedor, could you please check that this PR does not affect performance and that the source of the memory leak is the cache of the previous results?

In `CachedVerifier`, a *weak* hashmap is used to avoid such a leak; once the ExecutionScenario is garbage collected, the corresponding cache is also removed. 
```
    private final Map<ExecutionScenario, Set<ExecutionResult>> previousResults = new WeakHashMap<>();
```

At the same time, reusing a verifier enables reusing the LTS structure, which improves the performance and produces a memory leak.

alefedor:
Hi @ndkoval !

You are right. I've seen the correct place of the memory leak but forgot about it.
The memory leak is actually not in caching results but here
https://github.com/Kotlin/kotlinx-lincheck/blob/1c3fdc0200769ce7b207251a0c299676c185dc21/src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/LTS.kt#L67

This hashmap strictly increases as the time goes on.
As a result, those 500Mb consisted of `LTS` and `LTS$StateInfo`.

So, the problem is reusing LTS, which, I suspect, is also not very useful when scenarios differ.

ndkoval:
It was useful. Could you please check how your change affects the performance?

alefedor:
@ndkoval 
Good point. It affects performance.

Baseline: stress (36s), model checking (1m17s)
Reset after every iteration: stress (44s), model checking (1m30s)
Reset every 5-th iteration: stress (39s), model checking (1m21s)

It is clear that we can not leave this memory leak. Potentially, It can be especially severe when there are non-primitive transformed parameters or values. They can hold the loader, which holds `ManagedStateHolder`, which holds `ManagedStrategy`, which holds the whole interleaving tree.

 So, do you have good suggestions for how to remove the memory leak without affecting performance much?

I can think of the following heuristics:
- Resetting every k-th iteration instead of every iteration. As you can see, resetting verifier every 5-th iteration works.  
- Removing after every iteration LTS states for large executions as they are unlikely to be reused.
- Removing after every iteration a constant fraction of LTS states. It can be, for example, the least popular states. If there were $S$ states and we filtered $(1-p)S$ of them ( $p<1$ ), then the maximum number of states can reach $\frac S p$, which is in a constant number more than the number of states for one iteration

ndkoval:
We need more thoughtful experiments than one test case to evaluate the performance impact. However, I believe we can create a new verifier once per 100 iterations, and it won't affect most users.

alefedor:
@ndkoval 
Please have a look on the new approach.
The verifier is:
1. Renewed every 100 iterations to bound the maximum memory leak (soft anti-leak tactic).
2. Renewed if less than 10% free memory is remaining to get more free memory (hard anti-leak tactic).

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -87,6 +93,18 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         return null
     }
 
+    private fun lacksFreeMemory(): Boolean {
+        val runtime = Runtime.getRuntime()
+        val maxMemory = runtime.maxMemory()
+        if (maxMemory == Long.MAX_VALUE && !showedUnboundedMaxMemoryWarning) { // unbounded max memory
+            showedUnboundedMaxMemoryWarning = true
+            reporter.logWarning("Unbounded maximum JVM memory detected. This can potentially lead to OutOfMemoryError or " +
```
ndkoval, line 101: This warning does not seem useful (no real use-cases where it helps), let's remove it for code simplicity

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -64,15 +66,19 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
...
+            val verifier = createVerifier(checkStateEquivalence = true)
             val scenario = customScenarios[i]
             scenario.validate()
             reporter.logIteration(i + 1, customScenarios.size, scenario)
             val failure = scenario.run(this, verifier)
             if (failure != null) return failure
         }
+        var verifier = createVerifier(checkStateEquivalence = true)
         repeat(iterations) { i ->
+            // if verifier is too old or not enough free memory, create a new one to avoid memory leaks
```
ndkoval, line 79: Let's add a reference to the issue (#128)

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -181,11 +194,14 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
             if (!checkStateEquivalence) return@also
             val stateEquivalenceCorrect = it.checkStateEquivalenceImplementation()
             if (!stateEquivalenceCorrect) {
-                if (requireStateEquivalenceImplCheck) {
-                    val errorMessage = StringBuilder().appendStateEquivalenceViolationMessage(sequentialSpecification).toString()
-                    error(errorMessage)
-                } else {
-                    reporter.logStateEquivalenceViolation(sequentialSpecification)
+                if (!showedStateEquivalenceError) {
```
ndkoval, line 197: It's simpler to create a new verifier with `checkStateEquivalence = true` only the first time.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -87,6 +94,12 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         return null
     }
 
+    private fun lacksFreeMemory(): Boolean {
+        val runtime = Runtime.getRuntime()
```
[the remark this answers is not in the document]
alefedor: Wait, the warning is no longer here...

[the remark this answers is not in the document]
alefedor: This is not a warning, just a safety mechanism which ensures that `OutOfMemoryError` can not be caused by verifier. Typically, 100 iterations will pass faster than the memory will be exhausted. Yes, GC can be postponed until necessary but Garbage Collectors still try to avoid situations of "almost completely filled memory" since they can lead to long pauses. False positive alarm can lead to reduced performance (for which we do not have experiments), but false negative alarm leads to `OutOfMemoryError` and to Lincheck crash

btwilk:
While I'm happy with a workaround to (https://github.com/Kotlin/kotlinx-lincheck/issues/124), I'm also curious as to whether the root cause for the non-(LTS cache) memory leak involving non-primitive operation parameters can be directly fixed.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -64,15 +64,23 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
...
-        val verifier = createVerifier(checkStateEquivalence = true)
         for (i in customScenarios.indices) {
+            val verifier = createVerifier(checkStateEquivalence = i == 0)
             val scenario = customScenarios[i]
             scenario.validate()
             reporter.logIteration(i + 1, customScenarios.size, scenario)
             val failure = scenario.run(this, verifier)
             if (failure != null) return failure
         }
+        var verifier = createVerifier(checkStateEquivalence = true)
```
ndkoval, line 75: `checkStateEquivalence` can be true multiple times now

ndkoval (changes requested):

ndkoval (approved):

ndkoval:
@alefedor, thanks for investigating the original problem and providing a great-quality fix!

ndkoval:
Fixes #124

PR #143, opened 2023-01-20
Proper blocking operations support

ndkoval:
TBD


ndkoval:
Will be delivered in other PRs

PR #144, opened 2023-02-15
example guide - minor cleanup

rendner:
Hi, I read the [lincheck-guide](https://kotlinlang.org/docs/lincheck-guide.html) and noticed some minor problems. Most of them were missing imports and one non working example (didn't fail as specified).

Please have a look and let me know if any changes are needed.

### DIFF
--- a/docs/topics/operation-arguments.md
+++ b/docs/topics/operation-arguments.md
@@ -93,8 +93,7 @@ For this, configure the generator for a `key: Int` parameter:
    Post part:
    [get(2): [0]]
    ```
-6. To use model checking strategy for Java 9 and later, add the required JVM properties before running `modelCheckingTest` as mentioned in the [previous step](testing-strategies.md#write-a-model-checking-test)
-7. Finally, run `modelCheckingTest()`. It fails with the following output:
+6. Finally, run `modelCheckingTest()`. It fails with the following output:
 
    ```text
    = Invalid execution results =
--- a/docs/topics/operation-arguments.md
+++ b/docs/topics/operation-arguments.md
@@ -46,6 +46,7 @@ For this, configure the generator for a `key: Int` parameter:
    import org.jetbrains.kotlinx.lincheck.check
    import org.jetbrains.kotlinx.lincheck.paramgen.*
    import org.jetbrains.kotlinx.lincheck.strategy.stress.*
+   import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
    import org.junit.*
    
    class MultiMap<K, V> {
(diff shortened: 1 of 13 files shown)


docs/topics/operation-arguments.md
```diff
@@ -65,7 +66,7 @@ For this, configure the generator for a `key: Int` parameter:
        fun get(key: K): List<V> = map[key] ?: emptyList()
    }
    
-   @Param(name = "key", gen = IntGen::class, conf = "1:2")
+   @Param(name = "key", gen = IntGen::class, conf = "1:1")
```
rendner, line 69: I also get an expected error when I use `"1:3"` for `conf`. But the mentioned output for `modelCheckingTest` in the guide looks like `"1:1"` should be used.

Is there a reason why the original `"1:2"` doesn't work? I have an Mac M1, if that matters.

docs/topics/modular-testing.md
```diff
@@ -72,8 +72,7 @@ class MultiMapTest {
     @Test
     fun modularTest() = ModelCheckingOptions()
         .addGuarantee(forClasses(ConcurrentHashMap::class).allMethods().treatAsAtomic())
-        // Note that with the atomicity guarantees set, Lincheck can examine all possible interleavings,
-        // so the test successfully passes when the number of invocations is set to `Int.MAX_VALUE`
+        // Note that with the atomicity guarantees set, Lincheck can examine all possible interleavings.
```
rendner, line 75: Or should the example be modified to use a fixed implementation of the `MultiMap::add` method so that the test always succeeds?

docs/topics/operation-arguments.md
ndkoval: Can you provide more details about the error, please? |

The configuration is correct, "1:1" will always generate the same key.

rendner: >Can you provide more details about the error, please?

I am not sure whether you have requested the output for case `conf = "1:3"`. Therefore, here are the outputs for all three cases.

`stressTest` with `conf = "1:3"` for operation `get` and `add`:
```
= Invalid execution results =
Parallel part:
| add(3, 6): void | add(3, -8): void |
Post part:
[get(3): [-8]]
org.jetbrains.kotlinx.lincheck.LincheckAssertionError: 
= Invalid execution results =
Parallel part:
| add(3, 6): void | add(3, -8): void |
Post part:
[get(3): [-8]]
``` 

`stressTest` with `conf = "1:2"` for operation `get` and `add`:
_No error, the test is successful._


`stressTest` with `conf = "1:1"` for operation `get` and `add`:
```
= Invalid execution results =
Parallel part:
| add(1, 6): void | add(1, -8): void |
Post part:
[get(1): [-8]]
org.jetbrains.kotlinx.lincheck.LincheckAssertionError: 
= Invalid execution results =
Parallel part:
| add(1, 6): void | add(1, -8): void |
Post part:
[get(1): [-8]]
```

rendner: @ndkoval Do you need anything else?

ndkoval:
Hi @rendner! We are now working on a new Lincheck API, and I will return to this PR once it is released. Is it OK for you?

ndkoval:
Hi, @rendner! We have recently updated the guide slightly, removing some unnecessary parts in particular. May I ask you to update the PR, please?

rendner:
I updated the PR. 

I recognized two things which should be double checked by you:
- all links to the code examples in the guide markdown files point to the `guide` branch and not to the `master`
- the previous failing example in the guide works now with lincheck v2.17 (see comment: 499dc0a), but the test in the guide doesn't exist anymore on the `master` branch

ndkoval:
Fixes #188

ndkoval (approved):
@rendner, thank you very much for the fixes! 

I've done a pass over the suggested changes and merged the PR 🎉

PR #145, opened 2023-03-09
Bugfix for wait-notify support in presence of lock re-entrance

eupp:
In current implementation of `MonitorTracker` in model checking strategy, the monitor re-entrance information is lost when thread performs `wait` operation on some object. Thus upon wake-up the model checker just re-acquires a monitor once, but does not restore previous monitor re-entrance counter. This leads to a bug (see the new test added in this PR).

This PR fixes this problem by saving the information about lock re-entrance and then correctly restoring it.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -276,7 +276,7 @@ abstract class ManagedStrategy(
             failIfObstructionFreedomIsRequired {
                 // Log the last event that caused obstruction freedom violation
                 traceCollector?.passCodeLocation(tracePoint)
-                "Obstruction-freedom is required but an active lock has been found"
+                ObstructionFreedomSpinLockViolationMessage
             }
             checkLiveLockHappened(loopDetector.totalOperations)
             isLoop = true
@@ -440,7 +440,7 @@ abstract class ManagedStrategy(
         // Try to acquire the monitor
         while (!monitorTracker.acquireMonitor(iThread, monitor)) {
             failIfObstructionFreedomIsRequired {
-                "Obstruction-freedom is required but a lock has been found"
+                ObstructionFreedomLockViolationMessage
             }
             // Switch to another thread and wait for a moment when the monitor can be acquired
             switchCurrentThread(iThread, SwitchReason.LOCK_WAIT, true)
@@ -495,7 +495,7 @@ abstract class ManagedStrategy(
         if (inIgnoredSection(iThread)) return false
         newSwitchPoint(iThread, codeLocation, tracePoint)
         failIfObstructionFreedomIsRequired {
-            "Obstruction-freedom is required but a waiting on a monitor block has been found"
+            ObstructionFreedomWaitViolationMessage
         }
         if (withTimeout) return false // timeouts occur instantly
         while (monitorTracker.waitOnMonitor(iThread, monitor)) {
@@ -941,3 +941,12 @@ internal object ForcibleExecutionFinishException : RuntimeException() {
 }
 
 private const val COROUTINE_SUSPENSION_CODE_LOCATION = -1 // currently the exact place of coroutine suspension is not known
+
+private const val ObstructionFreedomSpinLockViolationMessage =
+    "Obstruction-freedom is required but an active lock has been found"
+
+private const val ObstructionFreedomLockViolationMessage =
+    "Obstruction-freedom is required but a lock has been found"
+
+private const val ObstructionFreedomWaitViolationMessage =
+    "Obstruction-freedom is required but a waiting on a monitor block has been found"
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -276,7 +276,7 @@ abstract class ManagedStrategy(
             failIfObstructionFreedomIsRequired {
                 // Log the last event that caused obstruction freedom violation
                 traceCollector?.passCodeLocation(tracePoint)
-                ObstructionFreedomSpinLockViolationMessage
+                OBSTRUCTION_FREEDOM_SPINLOCK_VIOLATION_MESSAGE
             }
             checkLiveLockHappened(loopDetector.totalOperations)
             isLoop = true
@@ -440,7 +440,7 @@ abstract class ManagedStrategy(
         // Try to acquire the monitor
         while (!monitorTracker.acquireMonitor(iThread, monitor)) {
             failIfObstructionFreedomIsRequired {
-                ObstructionFreedomLockViolationMessage
+                OBSTRUCTION_FREEDOM_LOCK_VIOLATION_MESSAGE
             }
             // Switch to another thread and wait for a moment when the monitor can be acquired
             switchCurrentThread(iThread, SwitchReason.LOCK_WAIT, true)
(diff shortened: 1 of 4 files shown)


eupp:
@ndkoval could you please also have a look at this.

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/transformation/WaitNotifyLockTest.kt
```diff
@@ -32,10 +32,31 @@ class WaitNotifyLockTest : VerifierState() {
     }
 }
 
-private class WaitNotifyLock {
+@ModelCheckingCTest(iterations = 30)
+class NestedWaitNotifyLockTest : VerifierState() {
```
ndkoval, line 36: Here are several tests now. Please rename the filename correspondingly and add a short comment to each test, explaining it verifies.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -852,16 +855,17 @@ private class MonitorTracker(nThreads: Int) {
...
+        info.timesAcquired--
+        if (info.timesAcquired == 0)
+            acquiredMonitors.remove(monitor)
     }
 
     /**
      * Returns `true` if the corresponding threads is waiting on some monitor.
      */
     fun isWaiting(iThread: Int): Boolean {
-        val monitor = acquiringMonitors[iThread] ?: return false
```
ndkoval, line 864: Why was it "acquiringMonitors"? Why did the code work at all?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -438,12 +438,12 @@ abstract class ManagedStrategy(
         if (inIgnoredSection(iThread)) return false
         newSwitchPoint(iThread, codeLocation, tracePoint)
         // Try to acquire the monitor
-        if (!monitorTracker.acquireMonitor(iThread, monitor)) {
-            failIfObstructionFreedomIsRequired { "Obstruction-freedom is required but a lock has been found" }
+        while (!monitorTracker.acquireMonitor(iThread, monitor)) {
+            failIfObstructionFreedomIsRequired {
+                "Obstruction-freedom is required but a lock has been found"
```
ndkoval, line 443: Please enhance the message and extract it to a constant

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -494,12 +494,14 @@ abstract class ManagedStrategy(
         if (!isTestThread(iThread)) return true
         if (inIgnoredSection(iThread)) return false
         newSwitchPoint(iThread, codeLocation, tracePoint)
-        failIfObstructionFreedomIsRequired { "Obstruction-freedom is required but a waiting on a monitor block has been found" }
+        failIfObstructionFreedomIsRequired {
+            "Obstruction-freedom is required but a waiting on a monitor block has been found"
```
ndkoval, line 498: Please enhance the message and extract it to a constant

ndkoval (changes requested):
The PR looks good to me. I have just a couple of minor suggestions.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Here I just renamed `acquiringMonitors` into `waitingMonitors`.

The `MonitorTracker`  class maintains two mappings (previously names as):
- `acquiredMonitors` mapping each locked object to `MonitorAcquiringInfo`
- `acquiringMonitors` mapping each thread to object it wants to lock (but cannot because it acquired by some another thread).

These two mappings had very similar names, that were easy to confuse. Thus I renamed `acquiringMonitors` into `waitingMonitors`. Other naming suggestions are welcome.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: I extracted the messages into constants. 

How exactly do you wish to improve the error messages themself?

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/transformation/WaitNotifyLockTest.kt
eupp: Done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ndkoval: Thanks for the clarification. Can we have more than one monitor in `waitingMonitors`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ndkoval: Just make it simpler to read, something like "The algorithm should be non-blocking, but an active lock is detected"

ndkoval: Note that we do not check for other progress guarantees than obstruction-freedom, so I would suggest using "non-blocking" instead of "obstruction-freedom" everywhere for simplicity.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: No according to the current contract, because while thread is waiting to acquire some lock, it is blocked and cannot make further progress (and thus make an attempt to acquire another lock).

ndkoval: Let's rename to `waitingMonitor` for clarity

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -916,3 +941,12 @@ internal object ForcibleExecutionFinishException : RuntimeException() {
 }
 
 private const val COROUTINE_SUSPENSION_CODE_LOCATION = -1 // currently the exact place of coroutine suspension is not known
+
+private const val ObstructionFreedomSpinLockViolationMessage =
```
ndkoval, line 945: Please use THIS_CONSTANT_NAME_FORMAT

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: DOne

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Done

ndkoval (approved):

PR #153, opened 2023-04-01
ShortGen unit test

SampathKumarAmex:


ndkoval:
Could you please add tests for all standard parameter generators? Please also take a look at #164.

ndkoval:
@SampathKumarAmex, I've looked at the PR and noticed you use `ShortGen` directly. Such a test would be significantly more reliable when using `short` in Lincheck tests. Therefore, I'm declining the PR and opening a GitHub issue #193.

PR #154, opened 2023-04-04
Fix handleExceptionsAsResult documentation

pettermahlen:
The annotation parameter needs to be an array of classes.


ndkoval:
@pettermahlen I have removed this code block from the README. We are also going to handle ALL exceptions as possible results in a next release, so I believe that this API part will no longer be an issue.

PR #155, opened 2023-04-04
ShortGen to kotlin

SampathKumarAmex:


ndkoval:
By the way, what is the purpose of this change? It rewrites only one of the generators from Java to Kotlin. It makes the code less consistent, neither introduces new functionality. We cannot accept this PR.

PR #156, opened 2023-04-05
Add Gradle instructions for Java 9+

pettermahlen:
This may save readers some time trying to figure out how to run their tests.


ndkoval:
@pettermahlen, thanks for the suggestion! We have recently rewritten README, and I hope that it now includes all the necessary information.

Issue #151, opened 2023-03-29
Try to provide an interleaving trace for stress tests


ndkoval:
When Lincheck finds an error via stress testing, it could run a model-checking test in the hope of reproducing the bug and, this way, provide a trace of the error. We can fix the number of invocations for this model checking test to some big constant, e.g., `10000`.

Issue #150, opened 2023-03-26
Remove `@ModelCheckingCTest` and `@ModelCheckingCTest` annotations


ndkoval:

alefedor:
Please consider marking them as deprecated instead as many tests used these annotations

ndkoval:
Note that bucket4j and grpc-jaca use these annotations:
https://grep.app/search?q=kotlinx.lincheck&filter[repo][0]=bucket4j/bucket4j 
https://grep.app/search?q=kotlinx.lincheck&filter[repo][0]=grpc/grpc-java

ndkoval:
@alefedor, sure, we will deprecate them first.

eupp:
@jetbrains-junie please take this issue.

ghost:
### ✅ The execution was successful
> The annotations `@ModelCheckingCTest` and `@StressCTest` were successfully marked as deprecated with appropriate messages guiding users to new options-based configurations. Additionally, deprecation warnings were suppressed in related code and tests to maintain build stability without errors. The changes were completed, and the build was successful, ready for submission.



---


### 📌 Issues Detected During Processing

- Some issues occurred that may have limited the context available during generation.
- These **did not prevent** the operation, but fixing them could improve the results.

**Issues**:
> ❌ **I tried to build the project before getting started, but it didn’t work.**
> 
> There might be issues with the environment or the code itself. Please check [build log](https://github.com/JetBrains/lincheck/actions/runs/17492348228/junie/3936748829) for more details.
> 
> 
> 
> 




More information:
* 🔗 [GitHub Action](https://github.com/JetBrains/lincheck/actions/runs/17492348228)
* 💬 [Report the issue on our Discord and help us improve](https://jb.gg/junie/github)

PR #158, opened 2023-04-06
Single Lincheck options class with `testingTime(..)` option

### LINKED ISSUE #137: Add `testingTime(..)` option to configure Lincheck tests
There is almost no intuition on how to configure Lincheck tests. How to choose the number of threads and operations in each of them? How many different scenarios should be studied? Which number of invocations is sufficient to examine each of them? Users usually adjust these parameters in a way to achieve acceptable execution time. Lincheck should provide a way to configure only the total execution time, automatically tuning all the parameters most efficiently. This feature lowers the entry threshold and improves the test quality. 

While this feature is non-trivial, we can start with some straightforward implementation,  improving it later. The intuition is that any straightforward implementation should not be worse than the current approach, as nobody knows how to configure Lincheck tests properly.

eupp:
Closes #137, #150, #151

This PR deprecates `StressOptions`, `ModelCheckingOption`, and also `StressCTest`, `ModelCheckingCTest` annotations.
Instead, new class `LincheckOptions` is introduced to configure Lincheck run.
Configuration of scenarios number and invocations per iteration number is also deprecated.
Instead, now the preferred way to configure Lincheck run is through single `testingTime` option.

Adaptive strategy to adjust scenarios and invocations numbers dynamically during the run is implemented (see class `Planner`).
The implementation tracks the average execution time over last N invocations.
Using the average execution time, the estimated remaining time is computed.
If the estimated time is greater (smaller) that allocated testing time by an order of magnitude (K times), 
number of invocations is multiplied (divided) by a constant K.
If the estimated time is greater (smaller) that allocated testing time (but not that significantly), number of scenarios is increased (decreased) by a constant I.

For debugging purposes, I also add logging of invocations count and execution time after each iteration. 

Besides that, the following logic is implemented:
- 25% of time during testing is allocated to stress strategy, 75% to model checking
- if stress strategy discovers a bug, model checking strategy is run with the hope to reproduce trace.
    
As a backward compatibility (and for a transition period), I left the possibility to manually set strategy (stress or model checking), number of scenarios and number of invocations per scenario.

To configure strategy, I added `mode` option to `LincheckOptions`.
Currently it is visible to the end user, but perhaps we want to hide it.
Other deprecated options also left visible, we can hide them if necessary.

I have not touch existing tests. Some of them manually set number of scenarios/invocations.
For test classes derived from `AbstractLincheckTest` we can easily add new subset of tests for new behavior called `testWithHybridStrategy`.
It will allow us to test each strategy individually, as well as hybrid automatic strategy implemented in this PR.


src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestConfiguration.kt
```diff
@@ -21,21 +21,19 @@
...
-import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
-import org.jetbrains.kotlinx.lincheck.strategy.stress.*
 import org.jetbrains.kotlinx.lincheck.verifier.*
 import org.jetbrains.kotlinx.lincheck.verifier.linearizability.*
 import java.lang.reflect.*
 
 /**
  * Abstract configuration for different lincheck modes.
  */
+@Deprecated(
```
ndkoval, line 33: Please also annotate with Kotlin's `@Deprecated` annotation as well. Let's use the `ERROR` level, thus, forcing users to update the code. Please also indicate here that both ModelChecking and Stress Options are deprecated and that LincheckOptions incorporates both these functionalities.

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestConfiguration.kt
```diff
@@ -64,28 +62,4 @@ abstract class CTestConfiguration(
         const val DEFAULT_MINIMIZE_ERROR = true
         const val DEFAULT_TIMEOUT_MS: Long = 10000
     }
-}
-
-internal fun createFromTestClassAnnotations(testClass: Class<*>): List<CTestConfiguration> {
```
ndkoval, line 69: You still need this code; both the annotations and the old `Option`s must work until the next release. We will remove them after that, cleaning up the codebase accordingly.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+/*
+ * Some atomic primitives are common and can be analyzed from a higher level
+ * of abstraction or can not be transformed (i.e, Unsafe or AFU).
+ * Thus, we do not transform them and improve the trace representation.
+ *
+ * For example, in the execution trace where `AtomicLong.get()` happens,
+ * we print the code location where this atomic method is called
+ * instead of going deeper inside it.
+ */
+private fun isTrustedPrimitive(className: String) =
```
ndkoval, line 388: No need to be here.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+        this.eliminateLocalObjects = eliminateLocalObjects
+    }
+
+    /**
+     * Set to `true` to make Lincheck log all events in an incorrect execution trace.
+     * By default, Lincheck collapses the method invocations that were not interrupted
+     * (e.g., due to a switch to another thread), and omits all the details except for
+     * the method invocation result.
+     */
+    fun verboseTrace(verboseTrace: Boolean = true) = apply {
```
ndkoval, line 286: No need to have it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+        this.verboseTrace = verboseTrace
+    }
+
+    /**
+     * Set logging level, [DEFAULT_LOG_LEVEL] is used by default.
+     */
+    fun logLevel(logLevel: LoggingLevel) = apply {
```
ndkoval, line 293: No need to have it in the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+     * ab obstruction-freedom violation if [checkObstructionFreedom] is set.
+     */
+    fun hangingDetectionThreshold(hangingDetectionThreshold: Int) = apply {
+        this.hangingDetectionThreshold = hangingDetectionThreshold
+    }
+
+    /**
+     * Internal, DO NOT USE.
+     */
+    internal fun eliminateLocalObjects(eliminateLocalObjects: Boolean) = apply {
```
ndkoval, line 276: No need to have it in the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+    fun requireStateEquivalenceImplCheck(require: Boolean) = apply {
+        requireStateEquivalenceImplementationCheck = require
+    }
+
+    /**
+     * Use the specified maximum number of repetitions to detect endless loops (hangs).
+     * A found loop will force managed execution to switch the executing thread or report
+     * ab obstruction-freedom violation if [checkObstructionFreedom] is set.
+     */
+    fun hangingDetectionThreshold(hangingDetectionThreshold: Int) = apply {
```
ndkoval, line 269: No need to have it in the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+    fun checkObstructionFreedom(checkObstructionFreedom: Boolean = true) = apply {
+        this.checkObstructionFreedom = checkObstructionFreedom
+    }
+
+    /**
+     * Require correctness check of test instance state equivalency relation defined by the user.
+     * It checks whether two new instances of a test class are equal.
+     * If the check fails [[IllegalStateException]] is thrown.
+     */
+    fun requireStateEquivalenceImplCheck(require: Boolean) = apply {
```
ndkoval, line 260: No need to have it in the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * Add a guarantee that methods in some classes are either correct in terms of concurrent execution or irrelevant.
+     * These guarantees can be used for optimization. For example, we can add a guarantee that all the methods
+     * in `java.util.concurrent.ConcurrentHashMap` are correct and this way the strategy will not try to switch threads
+     * inside these methods. We can also mark methods irrelevant (e.g., in logging classes) so that they will be
+     * completely ignored (so that they will neither be treated as atomic nor interrupted in the middle) while
+     * studying possible interleavings.
+     */
+    fun addGuarantee(guarantee: ManagedStrategyGuarantee) = apply {
```
ndkoval, line 243: Nobody uses this feature, and it is hard to maintain. Let's remove it from the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /**
+     * The specified class defines the sequential behavior of the testing data structure;
+     * it is used by [Verifier] to build a labeled transition system,
+     * and should have the same methods as the testing data structure.
+     *
+     * By default, the provided concurrent implementation is used in a sequential way.
+     */
+    fun sequentialSpecification(clazz: Class<*>?) = apply {
```
ndkoval, line 231: Let's rename to `sequentialImplementation`

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+     */
+    fun invocationsPerIteration(invocations: Int) = apply {
+        this.invocations = invocations
+        this.adjustInvocations = false
+    }
+
+    /**
+     * The maximal amount of time dedicated to testing.
+     */
+    internal fun testingTime(timeMs: Long) = apply {
```
ndkoval, line 206: Let's rename it to `testingTimeInSeconds`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+     */
+    fun iterations(iterations: Int) = apply {
+        this.iterations = iterations
+        this.adjustIterations = false
+    }
+
+    /**
+     * Run each test scenario the specified number of times.
+     */
+    fun invocationsPerIteration(invocations: Int) = apply {
```
ndkoval, line 198: You don't need this parameter anymore, do you?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+     * @see LincheckMode
+     */
+    fun mode(mode: LincheckMode) = apply {
+        this.mode = mode
+    }
+
+    /**
+     * Number of different test scenarios to be executed.
+     */
+    fun iterations(iterations: Int) = apply {
```
ndkoval, line 190: You don't need this parameter anymore, do you?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+        this.minimizeFailedScenario = minimizeFailedScenario
+    }
+
+    /**
+     * The mode used for running tests.
+     *
+     * @see LincheckMode
+     */
+    fun mode(mode: LincheckMode) = apply {
+        this.mode = mode
```
ndkoval, line 184: Is that an internal API?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+    fun addCustomScenario(scenarioBuilder: DSLScenarioBuilder.() -> Unit) =
+        addCustomScenario(scenario { scenarioBuilder() })
+
+    /**
+     * If this feature is enabled and an invalid interleaving has been found,
+     * *lincheck* tries to minimize the corresponding scenario in order to
+     * construct a smaller one so that the test fails on it as well.
+     * Enabled by default.
+     */
+    fun minimizeFailedScenario(minimizeFailedScenario: Boolean) = apply {
```
ndkoval, line 174: You should ALWAYS minimize failed scenarios. Should it be an internal API?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+     * @see ExecutionScenario.postExecution
+     */
+    fun actorsAfter(actorsAfter: Int) = apply {
+        this.actorsAfter = actorsAfter
+    }
+
+    /**
+     * Use the specified execution generator.
+     */
+    fun executionGenerator(executionGenerator: Class<out ExecutionGenerator?>) = apply {
```
ndkoval, line 151: There is only one execution generator in Lincheck, let's remove this parameter.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * Generate the specified number of operation for the last sequential part of an execution.
+     *
+     * Note, that the the actual number of operations can be less due to some restrictions
+     * like [Operation.runOnce].
+     *
+     * @see ExecutionScenario.postExecution
+     */
+    fun actorsAfter(actorsAfter: Int) = apply {
```
ndkoval, line 144: You don't need this parameter.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * Generate the specified number of operation for the initial sequential part of an execution.
+     *
+     * Note, that the the actual number of operations can be less due to some restrictions
+     * like [Operation.runOnce].
+     *
+     * @see ExecutionScenario.initExecution
+     */
+    fun actorsBefore(actorsBefore: Int) = apply {
```
ndkoval, line 132: You don't need this parameter.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckOptions.kt
```diff
@@ -0,0 +1,394 @@
...
+
+    /**
+     * Generate the specified number of operations for each thread of the parallel part of an execution.
+     *
+     * Note, that the the actual number of operations can be less due to some restrictions
+     * like [Operation.runOnce].
+     *
+     * @see ExecutionScenario.parallelExecution
+     */
+    fun actorsPerThread(actorsPerThread: Int) = apply {
```
ndkoval, line 120: It should be `operationsPerThread`. Let's get rid of the word "actor".

src/jvm/main/org/jetbrains/kotlinx/lincheck/Planner.kt
```diff
@@ -0,0 +1,254 @@
...
+
+        // factor of invocations multiplied/divided by when we over- or under-perform the plan
+        // by an order of magnitude
+        private const val INVOCATIONS_FACTOR = 5
+
+        // lower bound of invocations allocated by iteration
+        private const val INVOCATIONS_LOWER_BOUND = 1_000
+
+        // upper bound of invocations allocated by iteration
+        private const val INVOCATIONS_UPPER_BOUND = 100_000
```
ndkoval, line 248: Does it work well with the current model checker? I guess the tests will fail with `OutOfMemoryError`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingOptions.kt
```diff
@@ -27,12 +27,10 @@ import org.jetbrains.kotlinx.lincheck.strategy.managed.*
...
-    override fun createTestConfigurations(testClass: Class<*>): ModelCheckingCTestConfiguration {
-        return ModelCheckingCTestConfiguration(testClass, iterations, threads, actorsPerThread, actorsBefore, actorsAfter,
-                executionGenerator, verifier, checkObstructionFreedom, hangingDetectionThreshold, invocationsPerIteration,
-                guarantees, requireStateEquivalenceImplementationCheck, minimizeFailedScenario,
-                chooseSequentialSpecification(sequentialSpecification, testClass), timeoutMs, eliminateLocalObjects,
-                verboseTrace, customScenarios)
-    }
-}
\ No newline at end of file
+@Deprecated(
```
ndkoval, line 30: Please deprecate with both Java's and Kotlin's `@Deprecated` annotation.

ndkoval (changes requested):

eupp:
@ndkoval ready for re-review

eupp:
@ndkoval 

I still have a small concern with respect to extensive use of internal API in tests. Maybe we should consider keeping some common options in public API?

Also, currently the following code pattern is used to access internal API in tests:
```
LincheckOptions {
    this as LincheckOptionsImpl
    maxThreads = 2
    maxOperationsInThread = 2
    mode = LincheckMode.ModelChecking
}
``` 

Maybe replace it with the following pattern (in my opinion it better reflects the intention here):
```
LincheckInternalOptions {
    maxThreads = 2
    maxOperationsInThread = 2
    mode = LincheckMode.ModelChecking
}
```

PR #159, opened 2023-04-07
Parameter generators random seeds and execution random seed made not equal

avpotapov00:
Parameter generators random seeds and execution random seed made not equal

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/RandomFactory.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/RandomFactory.kt
@@ -1,7 +1,7 @@
 /*
  * Lincheck
  *
- * Copyright (C) 2019 - 2020 JetBrains s.r.o.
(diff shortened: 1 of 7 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/RandomFactory.kt
```diff
@@ -0,0 +1,43 @@
...
+
+package org.jetbrains.kotlinx.lincheck
+
+import java.util.*
+
+
+object RandomFactory {
+
+    @JvmStatic
+    fun intGeneratorRandom() = Random(0)
```
btwilk, line 29: this would still be prone to subtle correlation issues between multiple generators of the same type

e.g. using two int generators to test the API below

`fun getAndAdd(index: Int, amount: Int)`

ndkoval:
The fix will be provided as a part of #164

PR #164, opened 2023-04-12
149 make parameter generators smarter

avpotapov00:


src/jvm/main/org/jetbrains/kotlinx/lincheck/RandomFactory.kt
```diff
@@ -0,0 +1,37 @@
...
+object RandomFactory {
+
+    private var seedGenerator = Random(SEED_GENERATOR_SEED)
+
+    fun resetSeedGenerator() {
+        seedGenerator = Random(SEED_GENERATOR_SEED)
+    }
+
+    fun createRandom(): Random = Random(seedGenerator.nextLong())
+
```
ndkoval, line 36: remove empty line

src/jvm/main/org/jetbrains/kotlinx/lincheck/RandomFactory.kt
```diff
@@ -0,0 +1,37 @@
...
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>
+ */
+
+package org.jetbrains.kotlinx.lincheck
+
+import java.util.Random
+
+private const val SEED_GENERATOR_SEED = 0L
+
+object RandomFactory {
```
ndkoval, line 27: Let's explain why you need `RandomFactory`. 

Wouldn't it be more convenient to provide a `RandomFactory` instance to the generators?

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/GeneratorSeedTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.kotlinx.lincheck.annotations.Param
+import org.jetbrains.kotlinx.lincheck.check
+import org.jetbrains.kotlinx.lincheck.paramgen.IntGen
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingOptions
+import org.junit.Test
+
+/**
+ * This test checks that parameter generators random use different seeds than executions generator.
+ * This test fails if seeds are equals.
+ * @see <a href="https://github.com/Kotlin/kotlinx-lincheck/issues/120">Corresponding bug description</a>.
```
ndkoval, line 35: In KDoc, the syntax for links is similar to the one in Markdown

src/jvm/test/org/jetbrains/kotlinx/lincheck/test/GeneratorSeedTest.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+@Param(name = "key", gen = IntGen::class, conf = "0:1")
+class GeneratorSeedTest {
```
ndkoval, line 38: You should also check that different values are generated when an operation has multiple arguments of the same type. The same if one of the arguments has `@Param` annotation, or both of them.

src/jvm/main/org/jetbrains/kotlinx/lincheck/paramgen/strategy/real/RandomDoubleGenStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>
+ */
+
+package org.jetbrains.kotlinx.lincheck.paramgen.strategy.real;
+
+
+public interface RandomDoubleGenStrategy {
+
+    double nextDouble();
```
ndkoval, line 26: All generators should use the same generation scheme you are introducing. Users should still be able to configure the range, but it should not affect the scheme.

ndkoval (changes requested):

ndkoval (approved):

PR #178, opened 2023-05-03
Change the license to MPL-2.0

### LINKED ISSUE #166: Change the license to MPL-2.0
This project was initially forked from the [lin-check](https://github.com/devexperts/lin-check) tool by Devexperts, where I was previously employed. As the original license was LGPL, we had to stick to it. Some users have pointed out that LGPL is "viral" (which is partially true) and sometimes prevents using the framework in big projects. 

Thanks to Devexperts, they have recently [changed](https://github.com/devexperts/lin-check/releases/tag/lincheck-3.0) the original license to MPL-2.0, enabling us to do the same.

ndkoval:
Closes #166


ndkoval:
Approved by Jan Cichy

PR #182, opened 2023-05-09
Add a link to the KotlinConf 2023 talk on Lincheck.

lowasser:


ndkoval (approved):
Thanks for the PR! I've also updated the list of public talks in the user guide.

PR #185, opened 2023-05-12
Drop "kotlinx" in package names of Lincheck classes and change the artifact location

### LINKED ISSUE #184: Move to "JetBrains" GitHub organization
As Lincheck is not a Kotlin-specific framework, the repository has been recently moved from "Kotlin" GitHub organization to "JetBrains." However, to complete the transition, two additional steps are necessary:

1. We need to change the package of Lincheck classes from `org.jetbrains.kotlinx.lincheck` to `org.jetbrains.lincheck`. Yes, this is a backward incompatible change, but it reduces the association with Kotlin.
2. Next, we need to change the artifact location from `org.jetbrains.kotlinx:lincheck:<version>` to `org.jetbrains.lincheck:lincheck:<version>`.

ndkoval:
Fixes #184


README.md
```diff
@@ -13,6 +13,17 @@ generates a set of random concurrent scenarios,
 examines them using either stress-testing or bounded model checking, and
 verifies that the results of each invocation satisfy the required correctness property (linearizability by default).
 
+## 📣 Package Renaming in Lincheck 2.18
```
whyoleg, line 16: minor comment from outside, as next version will be:
* fully backward incompatible
* new maven artefacts
* new license

May be it's a good idea to call it lincheck 3.0? This will help users a lot during migration period :)

ndkoval: That's a good point, thanks. The version "2.18" is more or less a placeholder for now, and the new release will likely be "3.0".

As we are also going to change the API dramatically, we are considering postponing this change and applying it only to the new API, keeping the current one in the "org.jetbrains.kotlinx.lincheck" package.

Issue #186, opened 2023-05-15
Parameter generator for `enum` classes


ndkoval:
Let's create a parameter generator for `enum` classes. We can reuse the existing integer generator for that, with the only addition of shuffling `enum` values in `reset`, so that different values will prevail in different scenarios.

PR #187, opened 2023-05-18
EnumGen added

avpotapov00:


src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -76,20 +76,15 @@ public static CTestStructure getFromTestClass(Class<?> testClass) {
     }
 
     @SuppressWarnings("removal")
-    private static void readTestStructureFromClass(Class<?> clazz, Map<String, ParameterGenerator<?>> namedGens,
+    private static void readTestStructureFromClass(Class<?> clazz,
```
ndkoval, line 79: Let's keep the "one argument per line" rule

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -210,6 +264,11 @@ private static ParameterGenerator<?> createGenerator(Param paramAnn, RandomProvi
         }
     }
 
+    @SuppressWarnings({"rawtypes", "unchecked"})
```
ndkoval, line 267: Just "@Suppress"?

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -199,7 +246,14 @@ private static ParameterGenerator<?> getOrCreateGenerator(Method m,
         if (!paramAnn.name().isEmpty())
             return checkAndGetNamedGenerator(namedGens, paramAnn.name());
         // Otherwise create new parameter generator
-        return createGenerator(paramAnn, randomProvider);
+        if (p.getType().isEnum()) {
```
ndkoval, line 249: Would moving this logic to "createGenerator" improve the code clarity?

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -154,6 +149,52 @@ private static void readTestStructureFromClass(Class<?> clazz, Map<String, Param
...
+                namedGens.put(paramAnn.name(), createGenerator(paramAnn, randomProvider));
+            }
+        }
+        return namedGens;
+    }
+
+    private static Map<String, Class<? extends Enum<?>>> collectNamedEnumGeneratorToClassMap(Class<?> clazz) {
+        Map<String, Class<? extends Enum<?>>> enumGeneratorNameToClassMap = new HashMap<>();
+
+        Arrays.stream(clazz.getDeclaredMethods())
```
ndkoval, line 175: Please document the code -- it's hard to read it now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -154,6 +149,52 @@ private static void readTestStructureFromClass(Class<?> clazz, Map<String, Param
         }
     }
 
+    private static Map<String, ParameterGenerator<?>> createNamedGens(Class<?> clazz, RandomProvider randomProvider) {
+        Map<String, ParameterGenerator<?>> namedGens = new HashMap<>();
+        // Traverse all operations to determine EnumGens types or throw if one named enum gen is associated with many types
+        Map<String, Class<? extends Enum<?>>> enumGeneratorNameToClassMap = collectNamedEnumGeneratorToClassMap(clazz);
+        // Read named parameter paramgen (declared for class)
+        for (Param paramAnn : clazz.getAnnotationsByType(Param.class)) {
+            if (paramAnn.name().isEmpty()) {
```
ndkoval, line 158: Use "when" instead of sever "if"s and document the code, please.

src/jvm/main/org/jetbrains/kotlinx/lincheck/ParameterGenerators.kt
```diff
@@ -37,6 +37,44 @@ internal object DummyParameterGenerator : ParameterGenerator<Any?> {
...
+
+    private val enumValues: MutableList<T>
+    private val intGenerator: ExpandingRangeIntGenerator
+    private val shuffleRandom = randomProvider.createRandom()
+
+    init {
+        val allEnumValues = enumClass.enumConstants.toList()
+
+        enumValues = if (configuration.isEmpty()) allEnumValues else {
+            configuration.replace("\\s", "").split(",").map { enumStr ->
```
ndkoval, line 54: It's safer to replace "\\s*,\\s*'" with "," (not sure that the regex syntax is correct here)

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
avpotapov00: Complier doesn't allow me to write `@Suppress`, it forces me to write `@Suppress(names = {...})`

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
avpotapov00: To my mind, in this place we can't use java's switch properly to improve the code quality.

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
avpotapov00: I think no, as we use different conditions to call first or second factory method in another places.

src/jvm/main/org/jetbrains/kotlinx/lincheck/ParameterGenerators.kt
avpotapov00: We use current approach in all parameter generation parsing methods, why do you think that it's safer?

ndkoval: Consider you have an `enum class` in Kotlin with whitespaces in value names. Unfortunately, you cannot specify such values with the current approach, since you remove all whitespaces. Please add the corresponding test and check.

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
ndkoval: `@Suppress("unchecked", "rawtypes")`?

avpotapov00: As we're in Java file, this approach doesn't work, only `@Suppress(names = {"unchecked", "rawtypes"})` works, but it suppresses nothing, so I suggest to keep current version.

ndkoval (approved):

ndkoval:
Fixes #186

PR #198, opened 2023-06-28
Validation exception stacktrace shortened

### LINKED ISSUE #140: Cut exception stack straces thrown in validation functions
When a validation function (annotated with `@Validate`) throws an exception, the full stack trace, including the Lincheck internals, is provided. It takes dozens of lines in the console, being useless at the same time. The suggestion is to cut the stack trace at the point of the validation function call.

avpotapov00:
Fixes #140


src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -81,7 +82,15 @@ private fun executeValidationFunction(instance: Any, validationFunction: Method)
     try {
         m.invoke(instance)
     } catch (e: Throwable) {
-        return e.cause
+        // cut off lincheck-related and reflection-related stacktrace elements
+        val validationException = e.cause ?: return e
+        val wrapperExceptionStackTraceLength = e.stackTrace.size
+
+        if (e is InvocationTargetException) {
```
ndkoval, line 89: Can it be not an `InvocationTargetException`?

ndkoval: Please add a comment explaining when another exception type can be thrown, clarifying why simply returning it is correct. Maybe, we need to throw an internal exception instead.

ndkoval (changes requested):
Please address the comment in `Utils.kt`

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
avpotapov00: Done

ndkoval (approved):

Issue #194, opened 2023-06-19
Hang in FixedActiveThreadsExecutor (Lincheck 2.19)


ben-manes:
When running on 2.18.1 a [github action](https://github.com/ben-manes/caffeine/actions/runs/5300645021/jobs/9594324647) takes 8 minutes to run the tests. When upgrading, it is still [running](https://github.com/ben-manes/caffeine/actions/runs/5306807313/jobs/9604830438) after 58 minutes. When running locally the tests complete without an error. I'll revert that upgrade to isolate it, but the other changes were benign code readability changes to please a static analyzer. So for now this is an fyi as I try to reproduce locally.

It canceled after a 60 minute timeout trying to verify an embedded copy of JCTools' `MpscGrowableArrayQueue`.

```
Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.MpscGrowableArrayQueueLincheckTest > modelCheckingTest STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.MpscGrowableArrayQueueLincheckTest > stressTest STARTED
```

ben-manes:
It ran fine when [downgrading](https://github.com/ben-manes/caffeine/actions/runs/5307320733/jobs/9605810561) to 2.18.1. Since it appears to [hang](https://github.com/ben-manes/caffeine/actions/runs/5307428313/jobs/9606023745) on 2.19, I probably need to ssh into it and run jstack or capture a jfr profile.

ben-manes:
```console
"Test worker" #1 prio=5 os_prio=0 cpu=99787.06ms elapsed=202.38s tid=0x00007f8a7002c800 nid=0x8ea waiting on condition  [0x00007f8a74f14000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@11.0.19/Native Method)
        at java.util.concurrent.locks.LockSupport.parkNanos(java.base@11.0.19/LockSupport.java:357)
        at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.getResult(FixedActiveThreadsExecutor.kt:112)
        at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.awaitTask(FixedActiveThreadsExecutor.kt:93)
        at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.await(FixedActiveThreadsExecutor.kt:89)
[...]

PR #173, opened 2023-05-01
Optimize internal synchronization

ndkoval:
Surprisingly, this small change speeds up Lincheck twice.

PR #204, opened 2023-07-03
Optimize synchronization when the number of threads exceeds the number of cores

ndkoval:
After #173, `Thread.yield()` and `park()` are called after spinning for a long time. However, this strategy may lead to starvation when the number of parallel threads exceeds the number of cores. As this is a rare case, we can call `yield()` and `park()` immediately on such hardware.

### DIFF
--- a/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
+++ b/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
@@ -11,48 +11,45 @@
 package org.jetbrains.kotlinx.lincheck_test.util
 
 import org.jetbrains.kotlinx.lincheck.*
-import org.jetbrains.kotlinx.lincheck.strategy.LincheckFailure
-import org.jetbrains.kotlinx.lincheck.verifier.LTS
+import org.jetbrains.kotlinx.lincheck.strategy.*
+import org.jetbrains.kotlinx.lincheck.verifier.*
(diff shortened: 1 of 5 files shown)


ndkoval:
I guess, this PR should also fix #194 (not sure about that, but let's add the fix anyway)

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -11,49 +11,46 @@
...
+// - pattern `org.jetbrains.kotlinx.lincheck.runner.TestThreadExecution(\d+)` (the number of thread may vary)
+// - everything from `java.base/` (because code locations may vary between different versions of JVM)
+private val TEST_EXECUTION_TRACE_ELEMENT_REGEX = listOf(
+    "(\\W*)at org\\.jetbrains\\.kotlinx\\.lincheck\\.runner\\.TestThreadExecution(\\d+)\\.run\\(Unknown Source\\)",
+    "(\\W*)at java.base\\/(.*)"
+).joinToString(separator = ")|(", prefix = "(", postfix = ")").toRegex()
+
+private val LINE_NUMBER_REGEX = Regex(":(\\d+)\\)")
+
 private fun assertValuesEqualsAndPrintAllOutputsIfFailed(
```
avpotapov00, line 54: Please remove this function as it has no usages now

avpotapov00 (approved):

Issue #206, opened 2023-07-05
ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 2


asm0dey:
```
Index -1 out of bounds for length 2
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 2
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.constructTraceGraph(TraceReporter.kt:139)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.extractDebugTrace(ModelCheckingStrategy.kt:143)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.runImpl(ModelCheckingStrategy.kt:126)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.run(ManagedStrategy.kt:129)
	at org.jetbrains.kotlinx.lincheck.LinChecker.run(LinChecker.kt:128)
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl(LinChecker.kt:82)
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl$lincheck(LinChecker.kt:48)
	at org.jetbrains.kotlinx.lincheck.LinChecker.check(LinChecker.kt:38)
	at org.jetbrains.kotlinx.lincheck.LinChecker$Companion.check(LinChecker.kt:151)
	at org.jetbrains.kotlinx.lincheck.LinCheckerKt.check(LinChecker.kt:166)
	at TestBase.modelCheckingTest(TestBase.kt:26)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
[...]

PR #213, opened 2023-07-15
Validation functions now running in ignored section

avpotapov00:


src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -446,6 +453,17 @@ internal open class ParallelThreadsRunner(
         super.close()
         executor.close()
     }
+
+    private inline fun executeInIgnoredSection(iThread: Int, action: () -> Unit) {
+        if (strategy is ManagedStrategy) {
+            strategy.enterIgnoredSection(iThread)
```
ndkoval, line 459: Do we really need to pass `iThread` here? It should always be `ManagedStrategy.currentThread`.

ndkoval (changes requested):

ndkoval:
@avpotapov00 thanks for the quick fix!

ndkoval:
Fixes #206

PR #214, opened 2023-07-19
Do not call `Thread.stop()`

### LINKED ISSUE #210: Tests fail from `Thread.stop` call as of JDK 20(?)
Since [that method now throws `UnsupportedOperationException`](https://github.com/openjdk/jdk/blame/master/src/java.base/share/classes/java/lang/Thread.java#L1654), we're seeing failures, e.g.,

```
java.lang.UnsupportedOperationException
	at java.base/java.lang.Thread.stop([Thread.java:1654](https://cs.corp.google.com/#search/&sq=package:%5Epiper$%20project:openjdk&q=java/lang/Thread.java:1654&ws=tap-presubmit-rerun-server/4622332&snapshot=2))
	at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.close([FixedActiveThreadsExecutor.kt:220](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt.html))
	at org.jetbrains.kotlinx.lincheck.runner.ParallelThreadsRunner.close([ParallelThreadsRunner.kt:339](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt.html))
	at kotlin.io.CloseableKt.closeFinally(Closeable.kt:56)
	at org.jetbrains.kotlinx.lincheck.strategy.stress.StressStrategy.run([StressStrategy.kt:67](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/strategy/stress/StressStrategy.kt.html))
	at org.jetbrains.kotlinx.lincheck.LinChecker.run([LinChecker.kt:152](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/LinChecker.kt.html))
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl([LinChecker.kt:87](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/LinChecker.kt.html))
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl$third_party_java_kotlinx_lincheck_kotlinx_lincheck([LinChecker.kt:59](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/LinChecker.kt.html))
	at org.jetbrains.kotlinx.lincheck.LinCheckerKt.checkImpl([LinChecker.kt:233](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/LinChecker.kt.html))
	at org.jetbrains.kotlinx.lincheck.test.AbstractLincheckTest.runInternalTest([AbstractLincheckTest.kt:40](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/test/AbstractLincheckTest.kt.html))
	at org.jetbrains.kotlinx.lincheck.test.AbstractLincheckTest.testWithStressStrategy([AbstractLincheckTest.kt:57](http://www.docjar.com/html/api/org/jetbrains/kotlinx/lincheck/test/AbstractLincheckTest.kt.html))
```

For now, we're just patching out the call. That doesn't appear to cause any issues for our runs of Lincheck's own tests (including under JDK 11), though I assume it could cause problems if people really did need the threads to exit.

An intermediate option would be to call `stop` but to catch the `UnsupportedOperationException`.

I haven't looked at whether interrupting the threads is likely to be of any use. I assume that you'd already be doing it if it were a general solution, so it's probably just a question of whether it's better than nothing.

ndkoval:
Do not call Thread.stop(), it does nothing and is no longer supported in JDK 20

Fixes #210

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt
@@ -36,13 +36,6 @@ internal class FixedActiveThreadsExecutor(private val nThreads: Int, runnerHash:
     private val results = atomicArrayOfNulls<Any>(nThreads)
 
     /**
-     * This flag is set to `true` when [await] detects a hang.
-     * In this case, when this executor is closed, [Thread.stop]
-     * is called on all the internal threads.
-     */
-    private var hangDetected = false
-
-    /**
      * Threads used in this runner.
      */
     val threads = Array(nThreads) { iThread ->
@@ -80,12 +73,6 @@ internal class FixedActiveThreadsExecutor(private val nThreads: Int, runnerHash:
             val i = task.iThread
             submitTask(i, task)
         }
-    }
-
-    private fun shutdown() {
-        // submit the shutdown tasks
-        for (i in 0 until nThreads)
-            submitTask(i, SHUTDOWN)
     }
 
     private fun submitTask(iThread: Int, task: Any) {
@@ -131,7 +118,6 @@ internal class FixedActiveThreadsExecutor(private val nThreads: Int, runnerHash:
             while (results[iThread].value === currentThread) {
                 val timeLeft = deadline - System.nanoTime()
                 if (timeLeft <= 0) {
-                    hangDetected = true
                     throw TimeoutException()
                 }
                 LockSupport.parkNanos(timeLeft)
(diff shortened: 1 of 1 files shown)


ndkoval:
If we simply get rid of `Thread.stop()`, threads may keep running, which significantly slows down the build.

ndkoval:
Another approach will be provided in #271

Issue #207, opened 2023-07-06
Unclear exception result


zuevmaxim:
![image](https://github.com/JetBrains/lincheck/assets/31644752/e57198c8-3670-46b8-8be0-452eb1775655)
Use this project to reproduce https://github.com/ndkoval/concurrent-programming-intensive-berlin

Two issues are here:
1) Minimisation did not drop the second operation. [See this check for the cause](https://github.com/JetBrains/lincheck/blob/e81e79e73a02934f24ff20a5e6fe3254c546d0d2/src/jvm/main/org/jetbrains/kotlinx/lincheck/execution/ExecutionScenario.kt#L100)

2) The cause of the lincheck failure is an invalid first operation, while the exception in the second operation is expected. However, this is unclear to the user. It would be nice to either _state that exception is expected_ or _highlight the concrete point that makes execution incorrect_ (where the linearisation fails)

ndkoval:
I don't think we can find a good solution. I suggest using ChatGPT to generate an error explanation in IDEA. It may fix the issue.

eupp:
The first point (1) "Minimisation did not drop the second operation." is actually easy to fix --- we need to patch minimizer so that it will not split single thread scenarios into init/parallel/post parts. As a byproduct of that, the minimizer will be able to further remove some actors from the scenario. 

In the example attached to this issue, it would allow the minimizer to remove actor that raises exception.

ndkoval:
We will allow removing all operations in the parallel part, which will fix the current issue: #215. As we do not have a better way to show exception results, I'm closing this issue.

PR #217, opened 2023-07-20
Fix scenario minimization

### LINKED ISSUE #215: Allow removing all operations in the parallel part during minimization

eupp:
Implement the following fixes in the scenario minimization:

* single-thread scenarios are not split into init/post parts, only parallel part with one thread is left;
* if after minimization parallel part of the first thread becomes empty, it is removed (the next non-empty thread in parallel part becomes first thread).

Closes #215.


eupp:
I also add 2 new tests for minimization, that should emulate the problem reported in #207.

src/jvm/test/resources/expected_logs/validation_function_failure.txt
```diff
@@ -23,8 +21,6 @@ The following interleaving leads to the error:
 | inc()    |
 | inc()    |
 | inc()    |
-| inc()    |
-| inc()    |
 | -------- |
 
 Detailed trace:
```
ndkoval, line 26: It's unclear from the output when the validation function fails

ndkoval (changes requested):

ndkoval (approved):

src/jvm/test/resources/expected_logs/validation_function_failure.txt
ndkoval: Sure, but can we do something with that? I suggest opening a separate issue and continuing the discussion there.

eupp: Ok, I opened the issue #220

PR #226, opened 2023-08-02
Clocks representation fix when initial executions are present

avpotapov00:
When scenario has initial part, clocks doesn't reflect them, considering only operations in the parallel part.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
@@ -45,9 +45,25 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         check(testConfigurations.isNotEmpty()) { "No Lincheck test configuration to run" }
         for (testCfg in testConfigurations) {
             val failure = testCfg.checkImpl()
-            if (failure != null) return failure
+            if (failure != null) return failure.fixClocksIfLinearisabilityVerifier(testCfg)
         }
         return null
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


src/jvm/test/resources/expected_logs/trace_reporting_init_post_parts.txt
```diff
@@ -1,17 +1,20 @@
...
-| foo(): 1          | bar(): void |
-| ------------------------------- |
-| enterPost(): void |             |
-| ------------------------------- |
+| ------------------------------------- |
+|     Thread 1      |     Thread 2      |
+| ------------------------------------- |
+| enterInit(): void |                   |
+| ------------------------------------- |
+| foo(): 1 [1,0]    | bar(): void [1,0] |
```
ndkoval, line 7: These clocks do not provide new information, let's omit them

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -45,11 +45,27 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         check(testConfigurations.isNotEmpty()) { "No Lincheck test configuration to run" }
         for (testCfg in testConfigurations) {
             val failure = testCfg.checkImpl()
-            if (failure != null) return failure
+            if (failure != null) return failure.fixClocksIfLinearisabilityVerifier(testCfg)
```
ndkoval, line 48: Can we support all verifiers?

ndkoval (changes requested):

ndkoval:
@avpotapov00 gentle reminder

Issue #218, opened 2023-07-21
Spin cycle start label placed incorrectly when first cycle operation is some non-atomic method call


avpotapov00:
Consider the following example:

```
**
 * Checks that if spin cycle starts right after method call then spin cycle start event will be placed correctly.
 */
class SpinCycleFirstExecutionIsFirstInMethodCallTest {

    val counter = AtomicInteger(0)
    private val someUselessSharedState = AtomicBoolean(false)

    @Operation
    fun trigger() {
        counter.incrementAndGet()
        counter.decrementAndGet()
    }

    @Operation
    fun causesSpinLock() {
        if (counter.get() != 0) {
            deadSpinCycle()
        }
    }

    private fun deadSpinCycle() {
        do {
            val sharedVariableValue = getSharedVariable()
            someUselessSharedState.compareAndSet(sharedVariableValue, !sharedVariableValue)
        } while (true)
    }

    private fun getSharedVariable(): Boolean = someUselessSharedState.get()


    @Test
    fun test() = ModelCheckingOptions()
        .addCustomScenario {
            parallel {
                thread { actor(SpinCycleFirstExecutionIsFirstInMethodCallTest::trigger) }
                thread { actor(SpinCycleFirstExecutionIsFirstInMethodCallTest::causesSpinLock) }
            }
        }
        .minimizeFailedScenario(false)
        .check(this::class)

}
```

The output is:

<img width="1083" alt="image" src="https://github.com/JetBrains/lincheck/assets/45040429/df09a2e7-dc59-4416-8a21-3b6a014d8e00">

`/* The following events repeat infinitely: */   ` Label is hidden into the `getSharedVariable` call, it's undesirable.

PR #228, opened 2023-08-02
Spin locks detection enhancements

### LINKED ISSUE #219: Spin cycle period detected incorrect when the first operation before a spin cycle is the same as the last operation in the cycle.
Consider the following example:

```
@Param(name = "element", gen = IntGen::class, conf = "0:3")
class FaaQueueSpinLockTest {

    private val queue = FlatCombiningQueue<Int>()

    @Operation
    fun enqueue(@Param(name = "element") element: Int) = queue.enqueue(element)

    @Operation
    fun dequeue() = queue.dequeue()

    @Test
    fun modelCheckingTest() {
        ModelCheckingOptions()
            .iterations(100)
            .invocationsPerIteration(5_000)
            .actorsBefore(2)
            .threads(3)
            .actorsPerThread(2)
            .actorsAfter(2)
            .checkObstructionFreedom(false)
            .minimizeFailedScenario(false)
            .sequentialSpecification(FaaIntQueueSequential::class.java)
            .check(this::class.java)
    }

}

class FlatCombiningQueue<E> {
    private val queue = ArrayDeque<E>() // sequential queue
    private val combinerLock = atomic(false) // unlocked initially
    private val tasksForCombiner = atomicArrayOfNulls<Any?>(TASKS_FOR_COMBINER_SIZE)

    private fun tryLock() = combinerLock.compareAndSet(false, true)
    private fun unlock() = check(combinerLock.compareAndSet(true, false))

    fun enqueue(element: E) {
        var index = randomCellIndex() // code locations: 59, 60, 61
        while (true) {
            if (tasksForCombiner[index].compareAndSet(null, element)) break // code locations: 17, 18
            index = randomCellIndex() // code locations: 59, 60, 61
        }

        while (!tryLock()) {
            val value = tasksForCombiner[index].value
            if (value is Result<*>) {
                tasksForCombiner[index].value = null
                return
            }
        }
        help()
        unlock()
        tasksForCombiner[index].value as Result<*>
        tasksForCombiner[index].value = null
        return
    }

    fun dequeue(): E? {
        var index = randomCellIndex()
        while (true) {
            if (tasksForCombiner[index].compareAndSet(null, Dequeue)) break
            index = randomCellIndex()
        }

        while (!tryLock()) {
            val value = tasksForCombiner[index].value
            if (value is Result<*>) {
                tasksForCombiner[index].value = null
                return value.value as? E
            }
        }
        help()
        unlock()

        val value = tasksForCombiner[index].value as Result<*>
//         Uncomment the line below to get correct implementation
//        tasksForCombiner[index].value = null
        return value.value as? E
    }


    private fun help() {
        repeat(TASKS_FOR_COMBINER_SIZE) {
            val task = tasksForCombiner[it].value ?: return@repeat
            if (task is Result<*>) return@repeat
            if (task === Dequeue) {
                tasksForCombiner[it].value = Result(queue.removeFirstOrNull())
            } else {
                queue.add(task as E)
                tasksForCombiner[it].value = Result(Unit)
            }
        }
    }

    private fun randomCellIndex(): Int =
        ThreadLocalRandom.current().nextInt(tasksForCombiner.size)
}

private const val TASKS_FOR_COMBINER_SIZE = 3 // Do not change this constant!

private object Dequeue

private class Result<V>(
    val value: V
)

class FaaIntQueueSequential {
    private val q = ArrayList<Int>()

    fun enqueue(element: Int) {
        q.add(element)
    }

    fun dequeue() = q.removeFirstOrNull()
    fun remove(element: Int) = q.remove(element)
}
```

This code hung in the `enqueue` operation, and in the output we see the following:
image
In trace is shown that cycle start from line 71, but it's not true.
image
(Also #218 Issue is present in the output above)

avpotapov00:
Closes #219 #218 and adds recursive spin-locks support.

Subtasks:
- [ ] Rebase on `develop` with `javaagent` 2d
- [ ] Code review 3d


src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestConfiguration.kt
```diff
@@ -51,7 +51,7 @@ abstract class CTestConfiguration(
         val DEFAULT_EXECUTION_GENERATOR: Class<out ExecutionGenerator?> = RandomExecutionGenerator::class.java
         val DEFAULT_VERIFIER: Class<out Verifier> = LinearizabilityVerifier::class.java
         const val DEFAULT_MINIMIZE_ERROR = true
-        const val DEFAULT_TIMEOUT_MS: Long = 10000
+        const val DEFAULT_TIMEOUT_MS: Long = 15000
```
avpotapov00, line 54: To make BufferedChannelTest green on TeamCity.

ndkoval (changes requested):
Here is an intermediate review. I need to switch to java agents now; after that, I'll finish the review.

ndkoval:
The changes will be delivered under #331

PR #229, opened 2023-08-10
Bytecode instrumentation cache

eupp:
Currently bytecode instrumentation can take significant time. Moreover, it is performed for each scenario even when testing the same data structure. 

In the context of the adaptive planning feature #158, instrumentation time can interfere with the invocations time, thus making it harder to get reliable prediction for the average invocation time. 

This PR provides simple instrumented bytecode caching, thus mitigating the performance overhead of instrumentation.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
@@ -13,6 +13,8 @@ package org.jetbrains.kotlinx.lincheck;
 import org.jetbrains.kotlinx.lincheck.runner.*;
 import org.jetbrains.kotlinx.lincheck.strategy.*;
 import org.jetbrains.kotlinx.lincheck.strategy.managed.*;
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*;
+import org.jetbrains.kotlinx.lincheck.strategy.stress.*;
 import org.objectweb.asm.*;
 import org.objectweb.asm.commons.*;
 import org.objectweb.asm.util.*;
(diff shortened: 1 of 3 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
```diff
@@ -120,7 +132,16 @@ public Class<?> loadClass(String name) throws ClassNotFoundException {
                 return result;
             }
             try {
-                byte[] bytes = instrument(originalName(name));
+                byte[] bytes = null;
+                if (bytecodeCache != null) {
+                    bytes = bytecodeCache.get(name);
```
ndkoval, line 137: Please use `putIfAbscent(..) { .. }` instead

ndkoval (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/TransformationClassLoader.java
eupp: Here the usage of `computeIfAbsent` in fact makes code more verbose, because:

1) `instrument` throws checked exception `IOException` which has to be handled inside lambda.
2) `bytecodeCache` might be null and it is Java file, so we cannot use Kotlin's `?` syntax sugar.

eupp: Here is how the code looks like:

```java
                if (bytecodeCache != null) {
                    bytes = bytecodeCache.computeIfAbsent(name, className -> {
                            try {
                                return instrument(originalName(className));
                            } catch (IOException e) {
                                throw new RuntimeException(e);
                            }
                        }
                    );
                } else {
                    bytes = instrument(originalName(name));
                }
```

eupp: We can remove checked exception from `instrument`, but still after this there would be two similar calls to `instrument` (in case when we have the cache, and when we don't)

PR #231, opened 2023-08-22
Specialized collections with package remapping 

avpotapov00:
1. Standart java collections replaced with Fastutils collections
2. Remapping `Fastutil` package to `Lincheck` package using Gradle shadow plugin

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,6 +16,7 @@ plugins {
     kotlin("multiplatform")
     id("maven-publish")
     id("kotlinx.team.infra") version "0.3.0-dev-64"
+    id("com.github.johnrengelman.shadow") version "7.1.2"
 }
 
 repositories {
@@ -44,6 +45,8 @@ kotlin {
             val kotlinxCoroutinesVersion: String by project
             val asmVersion: String by project
             val reflectionsVersion: String by project
+            val fastUtilVersion: String by project
+
             dependencies {
                 api("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
                 api("org.jetbrains.kotlin:kotlin-stdlib-common:$kotlinVersion")
@@ -52,6 +55,7 @@ kotlin {
                 api("org.ow2.asm:asm-commons:$asmVersion")
                 api("org.ow2.asm:asm-util:$asmVersion")
                 api("org.reflections:reflections:$reflectionsVersion")
+                api("it.unimi.dsi:fastutil:$fastUtilVersion")
             }
         }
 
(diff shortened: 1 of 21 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
```diff
@@ -49,13 +51,13 @@ private CTestStructure(List<ActorGenerator> actorGenerators, List<ParameterGener
      * Constructs {@link CTestStructure} for the specified test class.
      */
     public static CTestStructure getFromTestClass(Class<?> testClass) {
-        Map<String, OperationGroup> groupConfigs = new HashMap<>();
-        List<ActorGenerator> actorGenerators = new ArrayList<>();
-        List<Method> validationFunctions = new ArrayList<>();
-        List<Method> stateRepresentations = new ArrayList<>();
+        Map<String, OperationGroup> groupConfigs = new Object2ObjectOpenHashMap<>();
```
eupp, line 54: Just a random suggestion. 
Maybe add type-aliases like `LincheckHashMap` and factory functions like `lincheckListOf()` and use them everywhere? 
I think such solution has several benefits:
1) The underlying collection implementations can be changed quite easily by changing type-aliases.
2) We would hide collections' implementation details in one place.
3) The name `Object2ObjectOpenHashMap` is quire verbose compared to something like `LincheckHashMap`.

eupp: Oh, I see you did something similar in `CollectionUtils.kt`, but I guess there is no way to use type-aliases from Java : (

eupp: Anyway, what about using `LincheckMutableHashMap` naming scheme instead of `MutableObjectToObjectMap`?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/CTestStructure.java
avpotapov00: I agree with you, and due to this limitation on type aliases in Java, the solution with factory-like top-level functions as `lincheckListOf()` looks pretty nice, especially taking into account your arguments above.  @ndkoval Do you still think it is redundant?

ndkoval: I also like the flexibility it brings.

ndkoval: Please put this discussion as a comment.

build.gradle.kts
```diff
@@ -16,6 +16,7 @@ plugins {
     kotlin("multiplatform")
     id("maven-publish")
     id("kotlinx.team.infra") version "0.3.0-dev-64"
+    id("com.github.johnrengelman.shadow") version "7.1.2"
```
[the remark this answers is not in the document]
avpotapov00: It seems it's impossible

ndkoval:
With the current lazy transformation approach for model checking, we can safely keep using the standard collections. Therefore, I'm closing the PR. Thanks @avpotapov00 for investigating potential performance improvements!

PR #249, opened 2023-11-17
Transformation via javaagent

ndkoval:


ndkoval:
Will be delivered under #285 and #296

PR #262, opened 2024-01-22
Strategy run API refactoring

eupp:
This PR proposes a simple refactoring of the `Strategy` class API. 

**Motivation**

Currently, the `Strategy` API provides the following method to run the scenario:

```kotlin
abstract fun run(): LincheckFailure?
```

This method should run the scenario (some) number of times and returns the failure, in case if any invocation fails. 

This API leads to code duplication in the current code base.
* Both stress and model-checking strategies in fact run the scenario fixed number of times in a loop. This loop is duplicated in both implementations. 
* Both strategies run verifier on each invocation result. Again, the implementation of verification method is duplicated in both strategies. 

This PR proposes to change the API, so that the `Strategy` instead provides the method to run single invocation:
```kotlin
abstract fun runInvocation(): InvocationResult
``` 

This change allows to implement the previously duplicated functionality in a single place, as a bunch of `Strategy` interface extension methods. 

Note that in addition, the new API adds two (optional) methods to implement:
```kotlin
    /**
     * Sets the internal state of strategy to run the next invocation.
     *
     * @return true if there is next invocation to run, false if all invocations have been studied.
     */
    open fun nextInvocation(): Boolean = true

    /**
     * Initializes the invocation.
     * Should be called before each call to [runInvocation].
     */
    open fun initializeInvocation() {}
```

Thus, the pattern to run the strategy becomes as follows:
```kotlin
// run single invocation (if available)
with(strategy) {
     if (nextInvocation()) {
        initializeInvocation()
        runInvocation()
     }
}
```

For deterministic strategies (like model checking strategy), consecutive calls of `runInvocation` withouth `nextInvocation` calls, should lead to the same results:

```kotlin
with(strategy) {
     if (nextInvocation()) {
        // first invocation run
        initializeInvocation()
        val r1 = runInvocation()
        // second invocation run
        initializeInvocation()
        val r2 = runInvocation()
        // two invocations should return the same result
        check(r1 == r2)
     }
}
```

This change also simplifies implementation of the new features. 

* In #250 we need to collect the statistics of running each invocation (e.g. its running time). With the old API, this again would require the code duplication in all `Strategy` implementers. 

* In #158 we want to implement the adaptive dynamic selection of the number of invocations to be run. This change requires to offload the decision on how many times to run the scenario from the `Strategy` class into the `Planner` class. Again, with the new API this becomes trivial.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
@@ -16,12 +16,14 @@ import org.jetbrains.kotlinx.lincheck.strategy.*
 import org.jetbrains.kotlinx.lincheck.transformation.withLincheckJavaAgent
 import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingCTestConfiguration
 import org.jetbrains.kotlinx.lincheck.verifier.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
+import org.jetbrains.kotlinx.lincheck.strategy.stress.*
 import kotlin.reflect.*
 
 /**
  * This class runs concurrent tests.
  */
-class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
+class LinChecker(private val testClass: Class<*>, options: Options<*, *>?) {
     private val testStructure = CTestStructure.getFromTestClass(testClass)
     private val testConfigurations: List<CTestConfiguration>
     private val reporter: Reporter
@@ -72,40 +74,39 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
     }
 
     private fun CTestConfiguration.checkImpl(): LincheckFailure? {
-        val exGen = createExecutionGenerator(testStructure.randomProvider)
-        for (i in customScenarios.indices) {
-            val verifier = createVerifier()
-            val scenario = customScenarios[i]
-            scenario.validate()
-            reporter.logIteration(i + 1, customScenarios.size, scenario)
-            val failure = scenario.run(this, verifier)
-            if (failure != null) {
-                runReplayForPlugin(failure, verifier)
-                return failure
+        var verifier = createVerifier()
+        val generator = createExecutionGenerator(testStructure.randomProvider)
+        val randomScenarios = generateSequence {
+            generator.nextExecution().also {
+                // reset the parameter generator ranges to start with the same initial bounds for each scenario.
+                testStructure.parameterGenerators.forEach { it.reset() }
             }
         }
-        checkAtLeastOneMethodIsMarkedAsOperation(testClass)
-        var verifier = createVerifier()
-        repeat(iterations) { i ->
+        val scenarios = customScenarios.asSequence() + randomScenarios.take(iterations)
+        val scenariosSize = customScenarios.size + iterations
+        scenarios.forEachIndexed { i, scenario ->
+            val isCustomScenario = (i < customScenarios.size)
             // For performance reasons, verifier re-uses LTS from previous iterations.
-            // This behaviour is similar to a memory leak and can potentially cause OutOfMemoryError.
+            // This behavior is similar to a memory leak and can potentially cause OutOfMemoryError.
             // This is why we periodically create a new verifier to still have increased performance
             // from re-using LTS and limit the size of potential memory leak.
             // https://github.com/Kotlin/kotlinx-lincheck/issues/124
             if ((i + 1) % VERIFIER_REFRESH_CYCLE == 0)
                 verifier = createVerifier()
-            val scenario = exGen.nextExecution()
             scenario.validate()
-            reporter.logIteration(i + 1 + customScenarios.size, iterations, scenario)
-            val failure = scenario.run(this, verifier)
-            if (failure != null) {
-                val minimizedFailedIteration = if (!minimizeFailedScenario) failure else failure.minimize(this)
-                reporter.logFailedIteration(minimizedFailedIteration)
-                runReplayForPlugin(minimizedFailedIteration, verifier)
-                return minimizedFailedIteration
+            reporter.logIteration(i + 1, scenariosSize, scenario)
+            var failure = scenario.run(i, this, verifier)
+            if (failure == null)
+                return@forEachIndexed
+            if (minimizeFailedScenario && !isCustomScenario) {
+                var j = i + 1
+                reporter.logScenarioMinimization(scenario)
+                failure = failure.minimize { minimizedScenario ->
+                    minimizedScenario.run(j++, this, createVerifier())
+                }
             }
-            // Reset the parameter generator ranges to start with the same initial bounds on each scenario generation.
-            testStructure.parameterGenerators.forEach { it.reset() }
+            reporter.logFailedIteration(failure)
+            return failure
         }
         return null
     }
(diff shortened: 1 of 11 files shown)


ndkoval:
@eupp could you please rebase the PR onto `develop`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -50,80 +52,65 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
...
-            val scenario = customScenarios[i]
-            scenario.validate()
-            reporter.logIteration(i + 1, customScenarios.size, scenario)
-            val failure = scenario.run(this, verifier)
-            if (failure != null) return failure
-        }
         var verifier = createVerifier()
-        repeat(iterations) { i ->
+        val generator = createExecutionGenerator(testStructure.randomProvider)
+        val randomScenarios = generateSequence {
```
ndkoval, line 64: I don't think you need a sequence here, as right after that you generate all the scenarios.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -22,18 +25,130 @@ import org.objectweb.asm.ClassVisitor
...
+fun Strategy.runIteration(iteration: Int, invocationsBound: Int, verifier: Verifier): LincheckFailure? {
+    var spinning = false
+    for (invocation in 0 until invocationsBound) {
+        if (!(spinning || nextInvocation()))
+            return null
+        spinning = false
+        initializeInvocation()
+        val failure = run {
+            val result = runInvocation()
+            spinning = (result is SpinCycleFoundAndReplayRequired)
```
ndkoval, line 127: This logic is specific to the model checking mode. Can we encapsulate it in the `runInvocation` logic?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/stress/StressStrategy.kt
```diff
@@ -44,19 +42,11 @@ class StressStrategy(
...
-                    is CompletedInvocationResult -> {
-                        if (!verifier.verifyResults(scenario, ir.results))
-                            return IncorrectResultsFailure(scenario, ir.results)
-                    }
-                    else -> return ir.toLincheckFailure(scenario)
-                }
-            }
-            return null
-        }
+    override fun runInvocation(): InvocationResult {
```
ndkoval, line 45: `override fun runInvocation() = runner.run()`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -160,9 +152,10 @@ abstract class ManagedStrategy(
     protected abstract fun chooseThread(iThread: Int): Int
 
     /**
-     * Returns all data to the initial state.
+     * Resets all internal data to the initial state and initializes current invocation to be run.
      */
-    protected open fun initializeInvocation() {
+    override fun initializeInvocation() {
+        super.initializeInvocation()
```
ndkoval, line 158: Why do you need this call?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -219,37 +209,27 @@ abstract class ManagedStrategy(
         runner = createRunner()
         ManagedStrategyStateHolder.setState(runner.classLoader, this, testClass)
         runner.initialize()
+        initializeInvocation()
```
ndkoval, line 212: Shouldn't it be in the `runInvocation` logic?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -22,18 +25,130 @@ import org.objectweb.asm.ClassVisitor
...
+/**
+ * Runs one Lincheck's test iteration with the given strategy and verifier.
+ *
+ * @param iteration the id of the iteration.
+ * @param invocationsBound number of invocations to run.
+ * @param verifier the verifier to be used.
+ *
+ * @return the failure, if detected, null otherwise.
+ */
+fun Strategy.runIteration(iteration: Int, invocationsBound: Int, verifier: Verifier): LincheckFailure? {
```
ndkoval, line 118: The `iteration` parameter is unused

ndkoval, line 118: Why `invocationBound`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -50,80 +52,65 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         check(testConfigurations.isNotEmpty()) { "No Lincheck test configuration to run" }
         for (testCfg in testConfigurations) {
             val failure = testCfg.checkImpl()
-            if (failure != null) return failure
+            if (failure != null)
```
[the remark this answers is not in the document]
eupp: Reverted

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -50,80 +52,65 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
...
-            val verifier = createVerifier()
-            val scenario = customScenarios[i]
-            scenario.validate()
-            reporter.logIteration(i + 1, customScenarios.size, scenario)
-            val failure = scenario.run(this, verifier)
-            if (failure != null) return failure
-        }
         var verifier = createVerifier()
-        repeat(iterations) { i ->
+        val generator = createExecutionGenerator(testStructure.randomProvider)
```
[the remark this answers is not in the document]
eupp: To unify the code for running custom scenarios and generated scenarios. 
The logic was the same, but the code was duplicated. 

The only essential difference is that minimization is not run for custom scenarios, and this is currently done using a simple boolean flag.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
eupp: With sequences I wanted to preserve the same logic as before, so that the scenarios are generated lazily on demand. But also I wanted to have single loop to handle both custom and random scenarios. 

So I join into a signle sequence:
* custom scenarios list (converting it into a sequence) and 
* a sequence generating random scenarios -- I truncate it using `take` method, so that only `N = number of iterations` scenarios are generated. 

Please note, that `Sequence.take` [method](https://github.com/JetBrains/kotlin/blob/0938b46726b9c6938df309098316ce741815bb55/libraries/stdlib/common/src/generated/_Sequences.kt#L556) does not forces the evaluation of the sequence. 
So the random scenarios are generated lazily on demand.
If we discover failure earlier, subsequent scenarios will not be generated.

eupp: I can add a comment to the code to clarify this logic.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Added it just for safety. 
In case if in the future we add some code to super method (which is currently empty, yes).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Do you mean that we can remove `initializeInvocation` from the public interface `Strategy`, and instead just make it protected method of `ManagedStrategy`, which is called at the beginning of `runInvocation` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
eupp: I ported this code from #250 where this argument is actually used.

eupp: What would be a better name?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -109,6 +108,10 @@ abstract class ManagedStrategy(
         }
     }
 
+    override fun close() {
+        runner.close()
```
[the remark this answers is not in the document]
eupp: Yes, alternatively we can make `runner` a public `val` property of `Strategy` interface and call its `close` directly. 

Do you wish me to implement this approach?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
eupp: It depend on whether we want to count invocations with `SpinCycleFoundAndReplayRequired` in the total number of invocations. 
If we want, then, unfortunately, this check has to remain here. 
If not, the replay logic can be moved into model checking mode.

ndkoval:
@eupp could you please rebase the PR?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
eupp: I removed the unused argument for now.

eupp:
> @eupp could you please rebase the PR?

Rebased.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
ndkoval: Let's do count this case as an invocation, this is just a cost for detecting spin cycles

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
ndkoval: That would be wonderful!

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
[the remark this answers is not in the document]
ndkoval: Thanks for clarification!

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ndkoval: Sounds good.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
ndkoval: Just `invocations`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
ndkoval: Let's move the implementation to the `Strategy` class.

ndkoval (approved):

ndkoval:
@eupp please address the remaining minor comments and merge the PR  🎉

PR #274, opened 2024-02-13
Improve UX when running `*IsolatedTest`s from IntelliJ IDEA

ndkoval:
Now, when running `jvmTest` from IntelliJ IDEA, it includes the `*IsolatedTest` tests, which makes the task extremely slow. We need to be able to run these tests but isolate them somehow. One possible solution is to launch a new JVM instance after N (e.g., 10) tests.


ndkoval:
It's fine the the integration test with the latest kotlinx.coroutines version fails; the issue will be fixed under #273

avpotapov00 (approved):

PR #277, opened 2024-02-20
Fix potential IndexOutOfBoundsException in curActorIsBlocking and concurrentActorCausesBlocking

eupp:
both properties access the current actor id `actorId = currentActorId[iThread]` but do not handle the case when `actorId = -1`, that is the case before any actor has started executing


eupp:
@ndkoval I've looked into this, and it seems it would be hard to trigger the error and make such test deterministic.
For this problem to occur, it has to be the case, that one thread has started working and detected deadlock, while other threads has not yet started execution (i.e. did not call `onActorStart` for the first time). We have no control over threads execution order on this level. 

One possible way to write test would be to manually create `ModelCheckingStrategy` instance and call its methods in specific order to trigger the error. Do you think we should take this road?

ndkoval:
@eupp, in this case, I would suggest adding a comment describing why the added logic is required and leaving a link to this PR.

ndkoval:
Maybe it would be better to rewrite the code in a way similar to

```
// <comment describing the issue>
if (actorId < 0) return null
```

ndkoval (approved):

ndkoval (changes requested):
Please fix the build failure on Java 17 (Mac).

ndkoval (approved):

PR #281, opened 2024-02-21
Bugfix for Loop Detector

eupp:
Reset code location counters inside `LoopDetector` on each actor start


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -995,6 +996,11 @@ abstract class ManagedStrategy(
             return detectedFirstTime || detectedEarly
         }
 
+        fun onActorStart(iThread: Int) {
```
ndkoval, line 999: Please add a comment explaining what this function does and why this logic is important

ndkoval (changes requested):

ndkoval (approved):

Issue #280, opened 2024-02-20
Lincheck tests are slow when the number of CPUs is lower than the number of parallel threads


ndkoval:
Please see https://github.com/JetBrains/lincheck/issues/278#issuecomment-1953512719 for details:
> It seems to have [hung](https://app.circleci.com/pipelines/github/ben-manes/caffeine/2855/workflows/f69f4a38-26d4-4a24-bd43-964516299e24/jobs/71526) on aarch64.
>
> I have run it successfully locally on an M3 macbook, but not on CircleCI, likely because of fewer cpu cores (as we saw that affected Lincheck in past bugs).

ben-manes:
Thanks, working beautifully now!

ben-manes:
was there a regression? I tried to upgrade 2.27 to 2.28.x and it times out ([2.28](https://app.circleci.com/pipelines/github/ben-manes/caffeine/3025/workflows/5adff2ee-9902-4555-a56e-8a4240a1f162/jobs/76742/parallel-runs/0/steps/0-114), [2.28.1](https://app.circleci.com/pipelines/github/ben-manes/caffeine/3022/workflows/5977230b-b865-4e7a-a1ed-5c150a5fc176/jobs/76632)) like before.

eupp:
Hi @ben-manes !
Yes, this is probably a regression. I'll take a look at it. 
Could you please pin the version 2.27 for now?

ben-manes:
thanks @eupp! Please consider adding a test case.

PR #298, opened 2024-03-28
Optimize spin loops and fix regression

eupp:
This PR implements optimizes and simplified spin loops:
1. make the counter local variable;
2. count only the number if `spin` calls. 

It also fixes the regression problem on machines with small number of CPUs by changing the `nThreads` parameter of `resultSpinner` to also count for the main thread. If the number of scenario threads + main thread is greater than the number of available CPUs, the main thread does not spin and parks immediately.


eupp:
BTW, @ndkoval can we consider running single-CPU CI configuration for each PR automatically, so that in future we can spot the regressions related to #280 earlier?

ndkoval:
> BTW, @ndkoval can we consider running single-CPU CI configuration for each PR automatically, so that in future we can spot the regressions related to #280 earlier?

We already have "Build (Linux) on Java 17 - Model checking tests on a single CPU (Lincheck)"

eupp:
Is it run on each PR automatically, or do we have to launch it manually?

ndkoval:
On my laptop, the change slows down the build by x1.5

ndkoval (approved):

PR #299, opened 2024-04-05
Idea plugin

zuevmaxim:


src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,146 @@
...
+import kotlin.coroutines.Continuation
+
+fun traverseTestObject(obj: Any): Map<Any, Int> {
+    val result = hashMapOf<Any, Int>()
+    traverseTestObject(obj, Collections.newSetFromMap(IdentityHashMap()), result)
+
+    return result
+}
+
+private fun traverseTestObject(obj: Any, visualized: MutableSet<Any>, result: MutableMap<Any, Int>) {
```
ndkoval, line 40: What is this result? 

Please document the functions and their internals

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,146 @@
...
+                it == REMAPPED_PACKAGE_CANONICAL_NAME + "java.util.concurrent.atomic.AtomicBoolean"
+    } ?: false
+    return atomic
+}
+
+// Try to construct a string representation
+private fun shouldProcessFurther(obj: Any?, result: MutableMap<Any, Int>): Boolean {
+    if (obj == null || obj.javaClass.isImmutableWithNiceToString)
+        return false
+//    getObjectNumber(obj.javaClass, obj)
```
ndkoval, line 120: should it be uncommented?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,146 @@
...
+        it == REMAPPED_PACKAGE_CANONICAL_NAME + "java.util.concurrent.atomic.AtomicInteger" ||
+                it == REMAPPED_PACKAGE_CANONICAL_NAME + "java.util.concurrent.atomic.AtomicLong" ||
+                it == REMAPPED_PACKAGE_CANONICAL_NAME + "java.util.concurrent.atomic.AtomicReference" ||
+                it == REMAPPED_PACKAGE_CANONICAL_NAME + "java.util.concurrent.atomic.AtomicBoolean"
+    } ?: false
+    return atomic
+}
+
+// Try to construct a string representation
+private fun shouldProcessFurther(obj: Any?, result: MutableMap<Any, Int>): Boolean {
```
ndkoval, line 117: What is the semantics of this function?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,146 @@
...
+import org.jetbrains.kotlinx.lincheck.strategy.managed.getObjectNumber
+import java.lang.reflect.Modifier
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.util.*
+import java.util.concurrent.atomic.*
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.coroutines.Continuation
+
+fun traverseTestObject(obj: Any): Map<Any, Int> {
```
ndkoval, line 33: The name is unclear. Let's indicate that it enumerates the objects

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,146 @@
...
+                }
+
+                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicRef") value =
+                    value.javaClass.getDeclaredField("value").apply { isAccessible = true }.get(value)
+                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicInt") value =
+                    value.javaClass.getDeclaredField("value").apply { isAccessible = true }.get(value)
+                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicLong") value =
+                    value.javaClass.getDeclaredField("value").apply { isAccessible = true }.get(value)
+                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicBoolean") value =
+                    value.javaClass.getDeclaredField("_value").apply { isAccessible = true }.get(value)
```
ndkoval, line 77: Please take a look how Lincheck reads fields in the `LincheckJavaAgent` class via `Unsafe`

src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -43,6 +43,10 @@ class Reporter(private val logLevel: LoggingLevel) {
         val output = if (logLevel == WARN) outErr else out
         output.println(sb)
     }
+
+    fun logFailedIterationWarn(failure: LincheckFailure) = log(WARN) {
```
ndkoval, line 47: The name is unclear

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -89,6 +96,16 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         return null
     }
 
+    private fun CTestConfiguration.runReplay(failure: LincheckFailure, verifier: Verifier) {
```
ndkoval, line 99: Please document

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -48,6 +49,8 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
      */
     internal fun checkImpl(): LincheckFailure? {
         check(testConfigurations.isNotEmpty()) { "No Lincheck test configuration to run" }
+        // We need this call to disable breakpoints until an error has been found
+        ideaPluginEnabled()
```
ndkoval, line 53: Please make a separate function for clarity

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,107 @@
...
+@Suppress("UNUSED_PARAMETER")
+fun visualizeInstance(
+    testObject: Any,
+    numbersArrayMap: Array<Any>,
+    threadsArrayMap: Array<Any>,
+    threadToLincheckThreadIdMap: Array<Any>
+) {
+}
+
+fun needVisualization(): Boolean = false // may be replaced with 'true' in plugin
```
ndkoval, line 57: Always `true`?

ndkoval (changes requested):

ndkoval:
@avpotapov00 could you please add `beforeEvent` calls only when the plugin is enabled? This should fix the performance issue.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckClassLoader.java
```diff
@@ -158,7 +159,7 @@ private byte[] instrument(String className) throws IOException {
         ClassVersionGetter infoGetter = new ClassVersionGetter();
         cr.accept(infoGetter, 0);
         ClassWriter cw = new TransformationClassWriter(infoGetter.getClassVersion(), remapper);
-        LincheckClassVisitor cv = new LincheckClassVisitor(transformationMode, cw);
+        LincheckClassVisitor cv = new LincheckClassVisitor(transformationMode, cw, ideaPluginEnabled());
```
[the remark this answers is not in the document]
ndkoval: should it be a `val`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -318,4 +328,17 @@ internal object Injections {
...
+    // == PLUGIN ==
+
+    @JvmStatic
+    fun shouldInvokeBeforeEvent(): Boolean {
+        return eventTracker.shouldInvokeBeforeEvent()
+    }
+
+    @Suppress("UNUSED_PARAMETER") // for debug
+    @JvmStatic
+    fun readNextEventId(type: String): Int {
```
ndkoval, line 341: Is "read" a correct word here?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -318,4 +328,17 @@ internal object Injections {
 
     @JvmStatic
     val VOID_RESULT = Any()
+
+    // == PLUGIN ==
```
ndkoval, line 332: What plugin? Please clarify.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -204,9 +214,9 @@ internal object Injections {
      * Called from the instrumented code before any array cell write.
      */
     @JvmStatic
-    fun beforeWriteArray(array: Any?, index: Int, value: Any?, codeLocation: Int) {
-        if (array == null) return // Ignore, NullPointerException will be thrown
-        eventTracker.beforeWriteArrayElement(array, index, value, codeLocation)
+    fun beforeWriteArray(array: Any?, index: Int, value: Any?, codeLocation: Int): Boolean {
```
ndkoval, line 217: Please explain in the documentation what this function should return.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -187,9 +197,9 @@ internal object Injections {
      * Called from the instrumented code before each field write.
      */
     @JvmStatic
-    fun beforeWriteField(obj: Any?, className: String, fieldName: String, value: Any?, codeLocation: Int) {
-        if (obj == null) return // Ignore, NullPointerException will be thrown
-        eventTracker.beforeWriteField(obj, className, fieldName, value, codeLocation)
+    fun beforeWriteField(obj: Any?, className: String, fieldName: String, value: Any?, codeLocation: Int): Boolean {
```
ndkoval, line 200: Please explain in the documentation what this function should return.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -170,9 +180,9 @@ internal object Injections {
      * Called from the instrumented code before any array cell read.
      */
     @JvmStatic
-    fun beforeReadArray(array: Any?, index: Int, codeLocation: Int) {
-        if (array == null) return // Ignore, NullPointerException will be thrown
-        eventTracker.beforeReadArrayElement(array, index, codeLocation)
+    fun beforeReadArray(array: Any?, index: Int, codeLocation: Int): Boolean {
```
ndkoval, line 183: Please explain in the documentation what this function should return.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -153,9 +163,9 @@ internal object Injections {
      * Called from the instrumented code before each field read.
      */
     @JvmStatic
-    fun beforeReadField(obj: Any?, className: String, fieldName: String, codeLocation: Int) {
-        if (obj == null) return // Ignore, NullPointerException will be thrown
-        eventTracker.beforeReadField(obj, className, fieldName, codeLocation)
+    fun beforeReadField(obj: Any?, className: String, fieldName: String, codeLocation: Int): Boolean {
```
ndkoval, line 166: Please explain in the documentation what this function should return.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -83,21 +88,26 @@ internal object Injections {
...
     @JvmStatic
-    fun wait(monitor: Any, codeLocation: Int) {
-        eventTracker.wait(monitor, codeLocation, withTimeout = false)
+    fun wait(monitor: Any) {
+        eventTracker.wait(monitor, withTimeout = false)
     }
 
 
     /**
      * Called from the instrumented code instead of [Object.wait].
```
ndkoval, line 106: The documentation is outdated

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -83,21 +88,26 @@ internal object Injections {
         eventTracker.unpark(thread, codeLocation)
     }
 
+    @JvmStatic
+    fun beforeWait(codeLocation: Int) {
+        eventTracker.beforeWait(codeLocation)
+    }
+
     /**
      * Called from the instrumented code instead of [Object.wait].
```
ndkoval, line 97: The documentation is outdated

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -51,12 +51,17 @@ internal object Injections {
         return t.inTestingCode && !t.inIgnoredSection
     }
 
+    @JvmStatic
+    fun beforeLock(codeLocation: Int) {
+        eventTracker.beforeLock(codeLocation)
+    }
+
     /**
      * Called from instrumented code instead of the MONITORENTER instruction.
```
ndkoval, line 60: The documentation is outdated

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -51,12 +51,17 @@ internal object Injections {
         return t.inTestingCode && !t.inIgnoredSection
     }
 
+    @JvmStatic
+    fun beforeLock(codeLocation: Int) {
```
ndkoval, line 55: Please explain what this function does and why you need to have both `beforeLock` and `lock`. The same applies to the `wait`s.

src/jvm/main/org/jetbrains/kotlinx/lincheck/EventTracker.kt
```diff
@@ -46,4 +48,10 @@ internal interface EventTracker {
 
     fun onNewObjectCreation(obj: Any)
     fun onWriteToObjectFieldOrArrayCell(obj: Any, fieldOrArrayCellValue: Any?)
+
+    // plugin
+
+    fun shouldInvokeBeforeEvent(): Boolean
+
```
ndkoval, line 55: Please remove the empty line

src/jvm/main/org/jetbrains/kotlinx/lincheck/EventTracker.kt
```diff
@@ -46,4 +48,10 @@ internal interface EventTracker {
 
     fun onNewObjectCreation(obj: Any)
     fun onWriteToObjectFieldOrArrayCell(obj: Any, fieldOrArrayCellValue: Any?)
+
+    // plugin
```
ndkoval, line 52: Please clarify what plugin

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+package org.jetbrains.kotlinx.lincheck
+import org.jetbrains.kotlinx.lincheck.runner.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
+
+const val MINIMAL_PLUGIN_VERSION = "0.0.1"
+
+// ============== This methods are used by debugger from IDEA plugin to communicate with Lincheck ============== //
+
+/**
+ * Invoked from the strategy [ModelCheckingStrategy] when Lincheck found a bug.
```
ndkoval, line 21: found -> finds

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+fun testFailed(
+    failureType: String,
+    trace: Array<String>,
+    version: String?,
+    minimalPluginVersion: String
+) {
+}
+
+/**
+ * Debugger replaces the result of this method to `true` if idea plugin is enabled.
```
ndkoval, line 44: Doesn't it simply set the system property you are reading?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+fun ideaPluginEnabled(): Boolean {
+    // treat as enabled in tests if we want so
+    return eventIdSequentialCheck
+}
+
+/**
+ * If Debugger needs to replay execution (due to earlier trace point selection), it replaces the result of this
+ * method to `true`.
+ */
+fun replay(): Boolean {
```
ndkoval, line 55: shouldReplayInterleaving?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Debugger replaces the result of this method to `true` if idea plugin is enabled.
+ */
+fun ideaPluginEnabled(): Boolean {
+    // treat as enabled in tests if we want so
+    return eventIdSequentialCheck
+}
+
+/**
+ * If Debugger needs to replay execution (due to earlier trace point selection), it replaces the result of this
```
ndkoval, line 52: the debugger

ndkoval: the execution

ndkoval: Please check the grammar of the documentation with Grammarly, if possible -- it helps a lot with detecting most grammar mistakes

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+
+const val MINIMAL_PLUGIN_VERSION = "0.0.1"
+
+// ============== This methods are used by debugger from IDEA plugin to communicate with Lincheck ============== //
+
+/**
+ * Invoked from the strategy [ModelCheckingStrategy] when Lincheck found a bug.
+ * The Debugger creates a breakpoint on this method, so when it's called, debugger receives all the information about the
+ * failed test.
+ * This method is called first, to provide all required information,
```
ndkoval, line 24: "This method is called first" -- this part is unclear.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * If Debugger needs to replay execution (due to earlier trace point selection), it replaces the result of this
+ * method to `true`.
+ */
+fun replay(): Boolean {
+    return false // should be replaced with `true` to replay the failure
+}
+
+/**
+ * This method is called on every trace point shown to the user.
```
ndkoval, line 60: Please indicate that 
1) this method is called before the actual event, such as the read/write/MONITORENTER/MONITOREXIT/...
2) this method should not be called from a nesting function

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+ */
+@Suppress("UNUSED_PARAMETER", "unused")
+fun beforeEvent(eventId: Int, type: String) {
+    if (needVisualization()) {
+        val strategy = (Thread.currentThread() as? TestThread)?.eventTracker ?: return
+        visualize(strategy)
+    }
+}
+
+fun needVisualization(): Boolean = false // may be replaced with 'true' in plugin
```
ndkoval, line 74: isn't it always `true`? Why? When does the plugin call this method?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * This method receives all information about the test object instance to visualize.
+ * The Debugger creates a breakpoint inside this method and uses this method parameters to create the diagram.
+ *
+ * We pass Maps as Arrays due to difficulties with passing objects to the debugger (class version, etc.).
```
ndkoval, line 80: due to difficulties with passing objects to the debugger -- which objects?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @param testObject tested data structure.
+ * @param numbersArrayMap an array structured like [Object, objectNumber, Object, objectNumber, ...]. Represents a `Map<Any, Int>`.
+ * @param threadsArrayMap an array structured like [Thread, threadId, Thread, threadId, ...]. Represents a `Map<Any, Int>`.
+ * @param threadToLincheckThreadIdMap an array structured like [CancellableContinuation, threadId, CancellableContinuation, threadId, ...]. Represents a `Map<Any, Int>`.
+ */
+@Suppress("UNUSED_PARAMETER")
+fun visualizeInstance(
+    testObject: Any,
```
ndkoval, line 89: testInstance?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+    numbersArrayMap: Array<Any>,
+    threadsArrayMap: Array<Any>,
+    threadToLincheckThreadIdMap: Array<Any>
+) {
+}
+
+fun onThreadSwitchesOrActorFinishes() {}
```
ndkoval, line 96: Please document

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+
+// ======================================================================================================== //
+
+internal val eventIdSequentialCheck = System.getProperty("lincheck.debug.eventIdSequentialCheck") != null
```
ndkoval, line 100: The property name does not reflect what exactly it enables

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+
+private fun visualize(strategyObject: Any) = runCatching {
+    val strategy = strategyObject as ModelCheckingStrategy
+    val runner = strategy.runner as ParallelThreadsRunner
+    val testObject = runner.testInstance
+    val threads = runner.executor.threads
+
+    val labelsMap = createObjectToNumberMapAsArray(testObject)
```
ndkoval, line 108: why "lables"?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,190 @@
...
+import java.math.BigInteger
+import java.util.*
+import java.util.concurrent.atomic.*
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.coroutines.Continuation
+
+/**
+ * Traverses an object to enumerate it and all nested objects.
+ */
+fun createObjectToNumberMap(obj: Any): Map<Any, Int> {
```
ndkoval, line 39: This function does not need to be private. Please consider either making it `internal` or moving to `IdeaPlugin.kt` as a `private` function

ndkoval, line 39: It's unclear how you use these numbers and how they relate to the enumeration on the `Reporter` side: should it be the same code?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,190 @@
...
+
+                if (value is AtomicIntegerArray) {
+                    value = (0..value.length()).map { (value as AtomicIntegerArray).get(it) }.toIntArray()
+                }
+                if (value is AtomicReferenceArray<*>) {
+                    value = (0..value.length()).map { (value as AtomicReferenceArray<*>).get(it) }.toTypedArray()
+                }
+
+                if (value?.javaClass?.canonicalName?.startsWith("java.lang.invoke.") == true) {
+                    // Ignore
```
ndkoval, line 99: Why?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,190 @@
...
+                } else if (value is AtomicReferenceFieldUpdater<*, *> || value is AtomicIntegerFieldUpdater<*> || value is AtomicLongFieldUpdater<*>) {
+                    // Ignore
+                } else if (value is ReentrantLock) {
+                    // Ignore
```
ndkoval, line 103: Why not `ReentrantLock(isLocked = <isLocked>)`?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
```diff
@@ -25,9 +25,13 @@ public abstract class CachedVerifier implements Verifier {
...
     @Override
     public boolean verifyResults(ExecutionScenario scenario, ExecutionResult results) {
-        boolean newResult = previousResults.computeIfAbsent(scenario, s -> new HashSet<>()).add(results);
-        if (!newResult) return true;
-        return verifyResultsImpl(scenario, results);
+        Set<ExecutionResult> executionResults = previousResults.computeIfAbsent(scenario, s -> new HashSet<>());
+        if (executionResults.contains(results)) return true;
+        boolean isValid = verifyResultsImpl(scenario, results);
+        if (isValid) {
+            executionResults.add(results);
```
ndkoval, line 32: Why do you need this change?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
zuevmaxim: Because `executionResults ` should contain only valid results. While before it cached all the results and incorrectly returned `true` when called repeadetly

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -48,13 +49,20 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
...
         for (testCfg in testConfigurations) {
             val failure = testCfg.checkImpl()
             if (failure != null) return failure
         }
         return null
     }
 
+    private fun disableBreakpointsUntilErrorIsFound() {
+        // We need this call to disable breakpoints until an error has been found
+        ideaPluginEnabled()
```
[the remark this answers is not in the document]
zuevmaxim: We added a separate method with clear semantics

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
```diff
@@ -318,4 +328,17 @@ internal object Injections {
 
     @JvmStatic
     val VOID_RESULT = Any()
+
+    // == PLUGIN ==
+
+    @JvmStatic
+    fun shouldInvokeBeforeEvent(): Boolean {
+        return eventTracker.shouldInvokeBeforeEvent()
```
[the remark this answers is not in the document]
zuevmaxim: No, beforeEvent should not be called in ignored section for safe code ignorance when needed (see toString)

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
ndkoval: Why is it becoming an issue now?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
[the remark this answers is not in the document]
ndkoval: I see. However, I'm asking about the `shouldInvokeBeforeEvent` function, which is non-trivial under the hood.

build.gradle.kts
```diff
@@ -98,12 +98,17 @@ tasks {
...
+        val extraArgs = mutableListOf(
             "--add-opens", "java.base/java.lang=ALL-UNNAMED",
             "--add-opens", "java.base/jdk.internal.misc=ALL-UNNAMED",
             "--add-exports", "java.base/jdk.internal.util=ALL-UNNAMED",
-            "--add-exports", "java.base/sun.security.action=ALL-UNNAMED"
+            "--add-exports", "java.base/sun.security.action=ALL-UNNAMED",
         )
+        val withEventIdSequentialCheck: String by project
+        if (withEventIdSequentialCheck.toBoolean()) {
+            extraArgs.add("-Dlincheck.debug.eventIdSequentialCheck=true")
```
ndkoval, line 109: What does "sequential" mean here? In the context of concurrency, it can be unclear. Maybe "ordering" would suit better here?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -671,6 +696,9 @@ internal class LincheckClassVisitor(
                             loadNewCodeLocationId()
                             // STACK: array: Array, index: Int, array: Array, index: Int, value: Object, codeLocation: Int
                             invokeStatic(Injections::beforeWriteArray)
+                            ifStatement(condition = { /* already on stack */ }, ifClause = {
```
ndkoval, line 699: please make it multi-line, as in the existing usages

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -644,6 +666,9 @@ internal class LincheckClassVisitor(
                             loadNewCodeLocationId()
                             // STACK: array: Array, index: Int, array: Array, index: Int, codeLocation: Int
                             invokeStatic(Injections::beforeReadArray)
+                            ifStatement(condition = { /* already on stack */ }, ifClause = {
```
ndkoval, line 669: please make it multi-line, as in the existing usages

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -612,6 +631,9 @@ internal class LincheckClassVisitor(
                             loadNewCodeLocationId()
                             // STACK: owner: Object, owner: Object, fieldName: String, fieldName: String, value: Object, codeLocation: Int
                             invokeStatic(Injections::beforeWriteField)
+                            ifStatement(condition = { /* already on stack */ }, ifClause = {
```
ndkoval, line 634: please make it multi-line, as in the existing usages

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -89,7 +90,11 @@ internal class LincheckClassVisitor(
                 mv
             }
         }
-        if (methodName == "<clinit>") {
+        if (methodName == "<clinit>" ||
+            // Debugger implicitly evaluates toString for variables rendering
+            // We need to disable breakpoints in such a case, as the numeration will break
```
ndkoval, line 95: How does it help to disable breakpoints? Please clarify.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -25,7 +25,8 @@ import java.util.*
 
 internal class LincheckClassVisitor(
     private val transformationMode: TransformationMode,
-    classVisitor: ClassVisitor
+    classVisitor: ClassVisitor,
+    private val ideaPluginEnabled: Boolean
```
ndkoval, line 29: Why do you need to pass it as an argument? This property is system-wide, you can just read it.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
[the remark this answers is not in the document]
ndkoval: Actually, `beforeEvent` is ALWAYS called in an ignored section now -- please see the changes in `LincheckClassVisitor`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -28,11 +28,11 @@ data class Trace(val trace: List<TracePoint>)
  * [callStackTrace] helps to understand whether two events
  * happened in the same, nested, or disjoint methods.
  */
-sealed class TracePoint(val iThread: Int, val actorId: Int, callStackTrace: CallStackTrace) {
+sealed class TracePoint(val iThread: Int, val actorId: Int, callStackTrace: CallStackTrace, var beforeEventId: Int = -1) {
     // This field assignment creates a copy of current callStackTrace using .toList()
     // as CallStackTrace is a mutable list and can be changed after this trace point is created.
     internal val callStackTrace = callStackTrace.toList()
-    protected abstract fun toStringImpl(): String
+    internal abstract fun toStringImpl(): String
```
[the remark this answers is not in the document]
avpotapov00: No, please check how it's used now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LincheckClassLoader.java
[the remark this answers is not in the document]
avpotapov00: this parameter removed from the constructor

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
zuevmaxim: Because previously `CachedVerifier` was never used after incorrect result is found

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
zuevmaxim: We do not instrument toString and enter ignored section, so there are no beforeEvents inside

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
avpotapov00: This method only enumerates objects, not creating their string representation

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
avpotapov00: No, we with @zuevmaxim think that setting a system-wide property is less convenient than using an existing approach with an interception of the exact method result.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
avpotapov00: Updated documentation, please re-check

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -89,6 +101,21 @@ class LinChecker (private val testClass: Class<*>, options: Options<*, *>?) {
         return null
     }
 
+    /**
+     * Enables replay mode and re-runs the failed scenario if Lincheck IDEA plugin is enabled.
+     * Also, this logic cannot be placed only in the strategy code, as we may need to call it with a minimized
```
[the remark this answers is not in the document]
avpotapov00: It's mentioned in the line below.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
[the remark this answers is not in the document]
zuevmaxim: > Actually, beforeEvent is ALWAYS called in an ignored section now

When we fix it, shouldInvokeBeforeEvent should not be called in ignored section, as it checks `thread.inIgnoredSection`

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
ndkoval: What is the problem now?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
ndkoval: Why don't you need to enumerate ReentrantLock?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
ndkoval: Why?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
ndkoval: Is it clear from the code?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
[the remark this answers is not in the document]
ndkoval: I see that you call `toStringImpl()`, while `toString()` would do the same.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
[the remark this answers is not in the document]
ndkoval: I made it unclear: let's rephrase it (this is just a suggestion)

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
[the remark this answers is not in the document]
ndkoval: Ok. Anyway, we are discussing `shouldInvokeBeforeEvent` here

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
ndkoval: Please clarify in the documentation

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
avpotapov00: Ok, there is no reason to ignore it. Removed this if

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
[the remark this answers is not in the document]
avpotapov00: Yes, and it was like that before. Let's do all necessary refactoring in a separate PR?

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
avpotapov00: Removed

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
avpotapov00: Now, we re-use the same verifier after an error is found, and with the old logic, it would have found the same wrong result in the cache and indicated that execution is correct, while it's not. So, I absolutely agree that this change is necessary. I reflected it in the comments.

build.gradle.kts
```diff
@@ -98,12 +98,19 @@ tasks {
...
+        val extraArgs = mutableListOf(
             "--add-opens", "java.base/java.lang=ALL-UNNAMED",
             "--add-opens", "java.base/jdk.internal.misc=ALL-UNNAMED",
             "--add-exports", "java.base/jdk.internal.util=ALL-UNNAMED",
-            "--add-exports", "java.base/sun.security.action=ALL-UNNAMED"
+            "--add-exports", "java.base/sun.security.action=ALL-UNNAMED",
         )
+        val withEventIdSequentialCheck: String by project
+        if (withEventIdSequentialCheck.toBoolean()) {
+            extraArgs.add("-Dlincheck.debug.eventIdOrderingCheck=true")
```
ndkoval, line 109: Can you rename the system property to `lincheck.debug.withEventIdSequentialCheck`, please?

avpotapov00 (approved):

build.gradle.kts
```diff
@@ -98,12 +98,19 @@ tasks {
...
+        }
+        val version: String by project
+        extraArgs.add("-Dlincheck.version=$version")
```
ndkoval, line 112: How would you get the Lincheck version in another project? It will not pass it as a system property.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
ndkoval: Explain it in the comments then, and please explain why.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,183 @@
...
+fun visualizeInstance(
+    testInstance: Any,
+    numbersArrayMap: Array<Any>,
+    threadsArrayMap: Array<Any>,
+    threadToLincheckThreadIdMap: Array<Any>
+) {
+}
+
+/**
+ * The Debugger creates a breakpoint inside this method to know when the thread is switched.
```
ndkoval, line 105: inside this method or on this method call?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -0,0 +1,183 @@
...
+ * Step-over expects that the next suspension point is in the same thread.
```
ndkoval, line 106: "The following "step over" call expects ..."

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,187 @@
...
+import java.lang.reflect.Modifier
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.util.*
+import java.util.concurrent.atomic.*
+import java.util.concurrent.locks.ReentrantLock
+import kotlin.coroutines.Continuation
+
+/**
+ * Traverses an object to enumerate it and all nested objects.
```
ndkoval, line 36: Please clarify in the documentation why you need to enumerate objects

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -0,0 +1,187 @@
...
+ * Uses the same a numeration map as TraceReporter via [getObjectNumber] method, so objects have the
+ * same numbers, as they have in the trace.
+ */
+internal fun createObjectToNumberMap(obj: Any): Map<Any, Int> {
```
ndkoval, line 40: Should it be called "enumerateObjects"?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -172,47 +182,47 @@ internal class MonitorEnterTracePoint(
     callStackTrace: CallStackTrace,
     stackTraceElement: StackTraceElement
 ) : CodeLocationTracePoint(iThread, actorId, callStackTrace, stackTraceElement) {
-    override fun toStringImpl(): String = "MONITORENTER at " + stackTraceElement.shorten()
+    override fun toStringImpl(withLocation: Boolean): String = "MONITORENTER at " + stackTraceElement.shorten()
```
ndkoval, line 185: Let's create an abstract method for the first part only (such as `MONITORENTER`) here. It will prevent you from having bugs like here -- you are ignoring the `withLocation` property. With the suggestion, this logic of adding the location part will be written once in the parent class only.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
```diff
@@ -473,12 +473,14 @@ private val Class<*>.simpleNameForAnonymous: String get() {
 internal fun getObjectNumber(clazz: Class<Any>, obj: Any): Int = objectNumeration
     .computeIfAbsent(clazz) { IdentityHashMap() }
     .computeIfAbsent(obj) { 1 + objectNumeration[clazz]!!.size }
-private fun getObjectNumber(obj: Any): Int = objectNumeration
-    .computeIfAbsent(obj.javaClass) { IdentityHashMap() }
-    .computeIfAbsent(obj) { 1 + objectNumeration[it.javaClass]!!.size }
 
 private val objectNumeration = WeakHashMap<Class<Any>, MutableMap<Any, Int>>()
+    @Synchronized get
```
ndkoval, line 478: Please use Collections.synchronizedMap instead -- this will protect only reading the hashmap reference, not the hashmap content.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingCTestConfiguration.kt
```diff
@@ -43,7 +43,12 @@ class ModelCheckingCTestConfiguration(testClass: Class<*>, iterations: Int, thre
     timeoutMs = timeoutMs,
     customScenarios = customScenarios
 ) {
+    private var replay = false
```
ndkoval, line 46: Please rename to "isReplayModeForIdeaPluginEnabled"?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1702,4 +1839,13 @@ private const val OBSTRUCTION_FREEDOM_LOCK_VIOLATION_MESSAGE =
...
 private const val OBSTRUCTION_FREEDOM_WAIT_VIOLATION_MESSAGE =
-    "The algorithm should be non-blocking, but a wait call is detected"
\ No newline at end of file
+    "The algorithm should be non-blocking, but a wait call is detected"
+
+/**
+ * With idea plugin enabled, we should not use default Lincheck timeout
+ * as debugging may take more time than default timeout.
+ */
+private const val DEBUGGER_TIMEOUT = 1000L * 60 * 60 * 24 * 365
```
ndkoval, line 1848: Shouldn't it be just "INFINITE_TIMEOUT" here? The reason you need this is to not to fail by timeout when the execution is paused in the debugger.

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/CachedVerifier.java
ndkoval: Thank you, now it is clear.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -310,4 +454,12 @@ internal class ModelCheckingStrategy(
 
         fun build() = Interleaving(switchPositions, threadSwitchChoices, lastNoninitializedNode)
     }
+
+    companion object {
+        /**
+         * We provide lincheck version to [testFailed] method to the plugin be able to
+         * determine if this version is compatible with the plugin version.
+         */
+        private val lincheckVersion = System.getProperty("lincheck.version")
```
ndkoval, line 463: Users will not pass the Lincheck version this way. Please find another solution.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -67,14 +69,153 @@ internal class ModelCheckingStrategy(
                 continue
             }
             usedInvocations++
-            checkResult(invocationResult)?.let { return it }
+            checkResult(invocationResult)?.let { failure ->
+                if (replay && failure.trace != null) {
```
ndkoval, line 73: Please move this logic to a separate method and document it. Please also document all the steps in this method.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -67,14 +69,153 @@ internal class ModelCheckingStrategy(
...
+    override fun shouldInvokeBeforeEvent(): Boolean {
+        if (!replay) return false
+        if (!collectTrace) return false
+        val thread = (Thread.currentThread() as? TestThread) ?: return false
+
+        return thread.inIgnoredSection && suddenInvocationResult == null
+    }
+
+
+    private val LincheckFailure.type: String
```
ndkoval, line 104: Why do you need it? How does the plugin use this information? Please clarify in the code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -67,14 +69,153 @@ internal class ModelCheckingStrategy(
...
+     * |--------------------------------|------|
+     * | REGULAR                        | 0    |
+     * | ACTOR                          | 1    |
+     * | RESULT                         | 2    |
+     * | SWITCH                         | 3    |
+     * | SPIN_CYCLE_START               | 4    |
+     * | SPIN_CYCLE_SWITCH              | 5    |
+     * | OBSTRUCTION_FREEDOM_VIOLATION  | 6    |
+     */
+    private fun extractDebugTrace(failure: LincheckFailure, trace: Trace): Array<String> {
```
ndkoval, line 143: constructTraceForPlugin?

ndkoval (changes requested):

build.gradle.kts
avpotapov00: Solved

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
avpotapov00: Solved

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -310,4 +454,14 @@ internal class ModelCheckingStrategy(
...
         fun build() = Interleaving(switchPositions, threadSwitchChoices, lastNoninitializedNode)
     }
+
+    companion object {
+        /**
+         * We provide lincheck version to [testFailed] method to the plugin be able to
+         * determine if this version is compatible with the plugin version.
+         */
+        private val lincheckVersion by lazy {
+            LinChecker::class.java.getResource("/version.txt")?.readText() ?: "UNKNOWN"
```
ndkoval, line 480: It should never be "UNKNOWN"

ndkoval: You definitely need a test for that :)

src/commonMain/resources/version.txt
```diff
@@ -0,0 +1 @@
+2.29-SNAPSHOT
```
ndkoval, line 1: This file should not exist in the source code

ndkoval (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
zuevmaxim: The main problem with the system property is that we need to add it to all debug processes when the plugin enabled. On the other hand, we have a stable and precise way of communication via breakpoints. Also, it is used in other part, so the obvious choice is to use it here also.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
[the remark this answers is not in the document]
zuevmaxim: @avpotapov00 `toStringImpl ` became internal as it had extra `verbose` parameter before. We omitted the class&method name with this parameter

src/jvm/main/org/jetbrains/kotlinx/lincheck/Injections.kt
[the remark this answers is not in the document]
zuevmaxim: `shouldInvokeBeforeEvent` should not be called in ignored section, as it checks `thread.inIgnoredSection` inside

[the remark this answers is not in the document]
zuevmaxim: Final version: both `shouldInvokeBeforeEvent ` and `beforeEvent ` can be called in the ignored section, and they are called in the ignored section. However, on one level up the instrumented code is not invoked in the ignored section, so breakpoints will not hit. I removed `thread.inIgnoredSection` check as we rely on the instrumentation code that the method is called only when needed.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
ndkoval: Thanks for the clarification. Could you please put it into the code documentation, please?

zuevmaxim: I don't know what to put in code documentation, as we use such a way of communication in this file, while system property is used only for tests

PR #296, opened 2024-03-27
Byte-code transformation via a dynamically attached Java agent

ndkoval:

PR #305, opened 2024-04-17
Bump minimal plugin version to 0.2

zuevmaxim:
The #296 update breaks compatibility with the previous plugin versions, so we need to increase the minimum version


ndkoval:
@zuevmaxim we are going to merge #296 now, so the plugin version will have to be adjusted

PR #314, opened 2024-04-24
Execution results are presented in case of any failure

### LINKED ISSUE #244: Add actor results to trace in case of any failure
Now we print actor results only failure type is `IncorrectResultsFailure`. But it looks weird and not consistent, especially in plugin.
Let's always display actor results.

avpotapov00:
Closes #244


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
```diff
@@ -16,50 +16,55 @@ import org.jetbrains.kotlinx.lincheck.strategy.managed.*
...
     override fun toString() = StringBuilder().appendFailure(this).toString()
 }
 
 internal class IncorrectResultsFailure(
     scenario: ExecutionScenario,
-    val results: ExecutionResult,
+    results: ExecutionResult,
     trace: Trace? = null
-) : LincheckFailure(scenario, trace)
+) : LincheckFailure(scenario,results, trace)
```
eupp, line 29: Missing space here (and in few other places in this file):
```
LincheckFailure(scenario, results, trace)
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -311,62 +311,74 @@ private fun<T, U> requireEqualSize(x: List<T>, y: List<U>, lazyMessage: () -> St
...
-    val initPart = scenario.initExecution.zip(executionResult.initResults) {
+    val initPart = scenario.initExecution.zip(initResults) {
         actor, result -> ActorWithResult(actor, result, exceptionStackTraces).toString()
     }
-    val postPart = scenario.postExecution.zip(executionResult.postResults) {
+    val postPart = scenario.postExecution.zip(postResults) {
         actor, result -> ActorWithResult(actor, result, exceptionStackTraces).toString()
     }
-    var hasClocks = false
+    val hasClocks = failure is IncorrectResultsFailure && parallelResultsWithClock.any { threadResults -> threadResults.any { it?.clockOnStart?.empty != true }  }
```
eupp, line 345: This line is too long and hard to read. Please split into several lines.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -308,35 +308,53 @@ internal open class ParallelThreadsRunner(
...
+        afterParallelStateRepresentation: String?,
+        afterPostStateRepresentation: String?
+    ): ExecutionResult {
+        // We also collect results in case when some actors have not finished due to some reasons
+        // (execution has hung or unexpected exception).
+        // Therefore, some results may be null as corresponding actors won't finish.
+        // To process them correctly and pass only ready results, we do `takeWhile { it != null }` to cut off
+        // suffix with `null` results.
+        return ExecutionResult(
+            initResults = initialPartExecution?.results?.toList()?.takeWhile { it != null }.orEmpty(),
```
eupp, line 345: Here you cut the `null` results, only to restore them later back in the reporter class. Let's avoid that, and just let `ExecutionResult` class contain `nulls` in the first place. This way we would also preserve the invariant that `scenarion` and `result` objects always have the same size.

src/jvm/main/org/jetbrains/kotlinx/lincheck/execution/ExecutionResult.kt
```diff
@@ -13,7 +13,9 @@ import org.jetbrains.kotlinx.lincheck.*
 
 /**
  * This class represents a result of [ExecutionScenario] execution.
- * All the result parts should have the same dimensions as the scenario.
+ * If the execution is completed, all the result parts should have the exact dimensions of the scenario.
```
eupp, line 16: Let us allow `ExecutionResult` to store `Result?` and document that `null` denotes an operation that haven't finished.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/InvocationResult.kt
```diff
@@ -27,20 +27,22 @@ class CompletedInvocationResult(
 /**
  * Indicates that the invocation has run into deadlock or livelock found by [ManagedStrategy].
  */
-data object ManagedDeadlockInvocationResult : InvocationResult()
+class ManagedDeadlockInvocationResult(val results: ExecutionResult) : InvocationResult()
```
eupp, line 30: Now that all invocation results contain `results` field, let's propagate this field to the parent class `InvocationResult`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -397,10 +409,17 @@ internal fun StringBuilder.appendExecutionScenarioWithResults(
     return this
 }
 
+private fun <T> List<T>.fillWithNullsToSize(requiredSize: Int): List<T?> {
```
eupp, line 412: Let's get rid of this function (see my comments about `Result?` in the `ExecutionResult` class).

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/execution/ExecutionResult.kt
avpotapov00: Now, I decided to store `NotFinished` and `NotStarted` results

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
```diff
@@ -16,50 +16,59 @@ import org.jetbrains.kotlinx.lincheck.strategy.managed.*
...
+internal class ManagedDeadlockFailure(
     scenario: ExecutionScenario,
-    // Thread dump is not present in case of model checking
-    val threadDump: Map<Thread, Array<StackTraceElement>>?,
+    results: ExecutionResult,
     trace: Trace? = null
-) : LincheckFailure(scenario, trace)
+) : LincheckFailure(scenario,results, trace)
+
+internal class TimeoutDeadlockFailure(
```
eupp, line 37: Let's rename to just `TimeoutFailure`

src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -556,46 +594,85 @@ internal fun collectExceptionStackTraces(executionResult: ExecutionResult): Exce
...
+ *
+ * @return scenario and execution result without actors that didn't start.
+ */
+private fun LincheckFailure.cutNotStartedActors(): Pair<ExecutionScenario, ExecutionResult> {
+    val initActorsToTakeCount = results.initResults.count { it !is NotStarted }
+    val parallelActorsToTakeCount = results.parallelResults.map { parallel -> parallel.count { it !is NotStarted } }
+    val postActorsToTakeCount = results.postResults.count { it !is NotStarted }
+
+    return ExecutionScenario(
+        initExecution = scenario.initExecution.take(initActorsToTakeCount),
```
eupp, line 663: Why not `takeWhile` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/InvocationResult.kt
eupp: So let's change it to also contain results?)

Then it would be uniform, no special cases, no needs to do type-casts and type-checks (i.e. `result is CompletedInvocationResult`) when we want to access `results` field, etc.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -308,35 +308,78 @@ internal open class ParallelThreadsRunner(
...
+    /**
+     * Composes a list of execution results from a raw array representation.
+     * When execution is finished in a normal way we have all non-nullable [TestThreadExecution.results].
+     * But when the execution has hung, results for some actors may be not set.
+     * This method checks if we have some unset results and if so, it replaces the first one as [NotFinished],
+     * and the rest as [NotStarted].
+     *
+     * @param executionPart indicates what execution part this result correspond to
+     */
+    private fun collectExecutionResults(rawResults: Array<Result?>, executionPart: ExecutionPart): List<Result> {
```
eupp, line 366: I am still a bit confused. 

Why it is not simpler to just leave the `null`-s in the results?
The results array is initialized with `null`-s, and the result is set to non-null when the actor finishes. 
So `null` in the array indicates that the actor was not finished.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -308,35 +308,78 @@ internal open class ParallelThreadsRunner(
...
+        // If execution ended in the init part, then all actors after it was not started.
+        if (currentExecutionPart == INIT && executionPart != INIT) return List(rawResults.size) { NotStarted }
+        // If execution ended in the parallel part, then all actors after it was not started.
+        if (currentExecutionPart == PARALLEL && executionPart == POST) return List(rawResults.size) { NotStarted }
+        // Find the first unset result - it means this actor was started but not finished.
+        val nullFirstNotSetResultIndex = rawResults.indexOfFirst { it == null }
```
eupp, line 372: I think this logic is specific to results representation, not the model itself.

When we print the results to the console, we  want to remove all not-started actors and display only the first non-finished (hung) actor.
In other cases, we might want to handle these results differently.
So it is better to move this logic to the `Reporter.kt`.

eupp (changes requested):

eupp (approved):

PR #316, opened 2024-04-29
Fix Java 11 performance bug

### LINKED ISSUE #308: Investigate performance issues due to Java agents on Java 11 builds
A new approach to byte-code transformations via Java-agent #296 introduced significant performance overhead on some Java configurations (most notably, the [Java 11 configuration](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildDevelopOnJava11/4581382?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true) on our CI builds). 

The source of this performance degradation is unclear and needs to be investigated.

eupp:
Closes #308

This PR fixes the performance problem accidentally introduced in #296 , that manifested on our Java 11 CI builds. 
The root cause of the problem and the solution are described below. 

The problem was related to the JVM classes re-transformation, performed by the Java agent. 
For model checking instrumentation mode, the class re-transformation was performed lazily on demand. 
However, for the stress strategy, all the loaded classes were re-transformed eagerly for each Lincheck test.
During a long JVM run (e.g. in case of our CI builds), a lot of classes can be loaded to the JVM, and for each Lincheck test run, the Lincheck JVM agent in the stress model would try to re-transform them.

However, for the stress strategy we actually only need to intercept the suspension points, that is calls to certain internal coroutine methods. The observation is that most of the loaded classes would not contain these calls, so there is no need to constantly re-transform them. 

So the solution, implemented in this PR, is to remember all the loaded classes that actually have coroutine calls inside them, and only re-transform these classes (or newly loaded classes) on subsequent Lincheck agent installations.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@ lastCopyrightYear=2023
 
 jdkToolchainVersion=17
 runAllTestsInSeparateJVMs=false
-instrumentAllClassesInModelCheckingMode=false
+instrumentAllClasses=false
 withEventIdSequentialCheck=false
 
 kotlinVersion=1.9.21
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
@@ -28,7 +28,7 @@ internal class LincheckClassVisitor(
     private var classVersion = 0
 
     private var fileName: String = ""
-    private var className: String = ""
+    private var className: String = "" // internal class name
 
     override fun visitField(
         access: Int,
@@ -195,4 +195,7 @@ private class WrapMethodInIgnoredSectionTransformer(
     }
 }
 
+// Set storing canonical names of the classes that call internal coroutine functions;
+// it is used to optimize class re-transformation in stress mode by remembering
+// exactly what classes need to be re-transformed (only the coroutines calling classes)
 internal val coroutineCallingClasses = HashSet<String>()
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
@@ -18,6 +18,9 @@ import org.objectweb.asm.commons.*
 import org.jetbrains.kotlinx.lincheck.transformation.InstrumentationMode.*
 import org.jetbrains.kotlinx.lincheck.transformation.transformers.*
 import sun.nio.ch.lincheck.*
+import sun.nio.ch.lincheck.Injections.*
+import java.util.*
+import kotlin.collections.HashSet
 
 internal class LincheckClassVisitor(
     private val instrumentationMode: InstrumentationMode,
@@ -73,7 +76,7 @@ internal class LincheckClassVisitor(
         if (access and ACC_NATIVE != 0) return mv
         if (instrumentationMode == STRESS) {
             return if (methodName != "<clinit>" && methodName != "<init>") {
-                CoroutineCancellabilitySupportTransformer(mv, access, methodName, desc)
+                CoroutineCancellabilitySupportTransformer(mv, access, className, methodName, desc)
             } else {
                 mv
             }
@@ -103,7 +106,7 @@ internal class LincheckClassVisitor(
         }
         mv = JSRInlinerAdapter(mv, access, methodName, desc, signature, exceptions)
         mv = TryCatchBlockSorter(mv, access, methodName, desc, signature, exceptions)
-        mv = CoroutineCancellabilitySupportTransformer(mv, access, methodName, desc)
+        mv = CoroutineCancellabilitySupportTransformer(mv, access, className, methodName, desc)
         if (access and ACC_SYNCHRONIZED != 0) {
             mv = SynchronizedMethodTransformer(fileName, className, methodName, mv.newAdapter(), classVersion)
         }
@@ -193,3 +196,5 @@ private class WrapMethodInIgnoredSectionTransformer(
         visitInsn(opcode)
     }
 }
+
+internal val coroutineCallingClasses = HashSet<String>()
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
@@ -18,8 +18,6 @@ import org.objectweb.asm.commons.*
 import org.jetbrains.kotlinx.lincheck.transformation.InstrumentationMode.*
 import org.jetbrains.kotlinx.lincheck.transformation.transformers.*
 import sun.nio.ch.lincheck.*
-import sun.nio.ch.lincheck.Injections.*
-import java.util.*
 import kotlin.collections.HashSet
 
 internal class LincheckClassVisitor(
@@ -29,8 +27,8 @@ internal class LincheckClassVisitor(
     private val ideaPluginEnabled = ideaPluginEnabled()
     private var classVersion = 0
 
-    private lateinit var fileName: String
-    private lateinit var className: String
+    private var fileName: String = ""
+    private var className: String = ""
 
     override fun visitField(
         access: Int,
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
@@ -354,7 +354,7 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
         internalClassName: String,
         classBytes: ByteArray
     ): ByteArray = transformedClassesCache.computeIfAbsent(internalClassName.canonicalClassName) {
-        nonTransformedClasses.putIfAbsent(internalClassName.canonicalClassName, classBytes.copyOf())
+        nonTransformedClasses.putIfAbsent(internalClassName.canonicalClassName, classBytes)
         val reader = ClassReader(classBytes)
         val writer = SafeClassWriter(reader, loader, ClassWriter.COMPUTE_FRAMES)
         try {
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
@@ -83,7 +83,7 @@ internal object LincheckJavaAgent {
     private var isBootstrapJarAddedToClasspath = false
 
     /**
-     * Names of the classes that were instrumented since the last agent installation.
+     * Names (canonical) of the classes that were instrumented since the last agent installation.
      */
     val instrumentedClasses = HashSet<String>()
 
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
@@ -13,11 +13,12 @@ package org.jetbrains.kotlinx.lincheck.transformation
 import net.bytebuddy.agent.*
 import org.jetbrains.kotlinx.lincheck.*
 import org.jetbrains.kotlinx.lincheck.transformation.InstrumentationMode.*
-import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransformer.nonTransformedClasses
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransformer.shouldTransform
+import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransformer.transformedClassesStress
+import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransformer.nonTransformedClasses
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentation
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentationMode
-import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentedClassesInTheModelCheckingMode
+import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentedClasses
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.INSTRUMENT_ALL_CLASSES_IN_MODEL_CHECKING_MODE
 import org.jetbrains.kotlinx.lincheck.util.readFieldViaUnsafe
 import sun.misc.Unsafe
@@ -84,7 +85,7 @@ internal object LincheckJavaAgent {
     /**
      * TODO
      */
-    val instrumentedClassesInTheModelCheckingMode = HashSet<String>()
+    val instrumentedClasses = HashSet<String>()
 
     /**
      * Dynamically attaches [LincheckClassFileTransformer] to this JVM instance.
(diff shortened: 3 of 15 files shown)


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/linearizability/RendezvousChannelCustomTest.kt
```diff
@@ -41,7 +41,7 @@ class RendezvousChannelCustomTest : VerifierState() {
     private val pollFun = RendezvousChannelCustomTest::poll
 
     @Test
-    fun testCancellation_01() = withLincheckJavaAgent(InstrumentationMode.STRESS) {
```
ndkoval, line 44: Shouldn't the verifier track the `getResult` calls?

gradle.properties
```diff
@@ -17,7 +17,7 @@ lastCopyrightYear=2023
 
 jdkToolchainVersion=17
 runAllTestsInSeparateJVMs=false
-instrumentAllClassesInModelCheckingMode=false
+instrumentAllClasses=false
```
ndkoval, line 20: Note: with this change, the Teamcity build name and configuration should be adjusted accordingly.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -155,11 +171,9 @@ internal object LincheckJavaAgent {
...
             .filter {
                 // Filter classes that were transformed by Lincheck and should be restored.
-                if (instrumentationMode == MODEL_CHECKING || !INSTRUMENT_ALL_CLASSES_IN_MODEL_CHECKING_MODE) {
-                    it.name in instrumentedClassesInTheModelCheckingMode
-                } else {
-                    true
-                }
+                if (!INSTRUMENT_ALL_CLASSES) {
+                    it.name in instrumentedClasses
+                } else true
```
ndkoval, line 176: Please wrap `true` with brackets:
```
if (...) {
} else {
  true 
}
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -312,17 +325,21 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
...
+        if (!INSTRUMENT_ALL_CLASSES &&
+            instrumentationMode == MODEL_CHECKING &&
+            className.canonicalClassName !in instrumentedClasses) {
+            return null
         }
         return transformImpl(loader, className, classBytes)
     }
 
-    private fun transformImpl(loader: ClassLoader?, className: String, classBytes: ByteArray): ByteArray = transformedClassesCache.computeIfAbsent(className) {
+    private fun transformImpl(loader: ClassLoader?, className: String, classBytes: ByteArray): ByteArray = transformedClassesCache.computeIfAbsent(className.canonicalClassName) {
```
ndkoval, line 342: Why `canonicalClassName`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -106,18 +107,33 @@ internal object LincheckJavaAgent {
...
+
+            // In the stress testing mode, Lincheck needs to track coroutine suspensions.
+            // As an optimization, we remember the set of loaded classes that actually
+            // have suspension points, so later we can re-transform only those classes.
+            instrumentationMode == STRESS -> {
+                check(instrumentedClasses.isEmpty())
+                val classes = getLoadedClassesToInstrument().filter {
+                    // new classes that were loaded after the latest STRESS mode re-transformation
+                    !transformedClassesStress.containsKey(it.name) ||
+                    // old classes that were already loaded before and have coroutine method calls inside
```
ndkoval, line 128: It might be the first time when a stress test is running, so `coroutineCallingClasses` is still empty. Therefore, you can ignore some classes that should not be ignored.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Yes, in this case `transformedClassesStress` would be empty, and the check `!transformedClassesStress.containsKey(it.name)` will return true.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Because the other place where I use `transformedClassesStress` is inside `install` method, where I have the `Class` instance, and I can only obtain its canonical name. 

So it is easier to convert ASM's internal name to the canonical name here, than the other way around.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/linearizability/RendezvousChannelCustomTest.kt
eupp: Yes, but `withLincheckJavaAgent` is already called from the `verify` method.

ndkoval: Is it? I don't see that. 

By the way, I also think we currently might have a bug with the model-checking mode, as the lazy transformation might not work in the verification phase, as we do not analyze anything in this phase, and, therefore, we do not ensure that the code dependencies are transformed.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
ndkoval: Alright, I got it. Can we please use `canonicalClassName` and `internalClassName` everywhere to highlight whether the class name is canonical or internal in each case? I understand that it is not a part of this PR, but with this explicit notion, it is easier to review the PR properly.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/linearizability/RendezvousChannelCustomTest.kt
eupp: > Is it?

Yes, here I mean special `verify` method that we only use in `RendezvousChannelCustomTest` and similar tests that test the verifier algorithm. This method calls `withLincheckJavaAgent`, so everything is okay. 

> By the way, I also think we currently might have a bug with the model-checking mode, as the lazy transformation might not work in the verification phase, as we do not analyze anything in this phase, and, therefore, we do not ensure that the code dependencies are transformed.

Yes, maybe, but that is a separate issue.

ndkoval: Could you write a test for this and check whether there is such an issue? It does not prevent the current PR from being merged, of course.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Ok, I renamed to explicit `canonicalClassName` or `internalClassName` everywhere in the `LincheckJavaAgent.kt`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/CoroutineSupportTransformers.kt
```diff
@@ -22,22 +23,30 @@ import sun.nio.ch.lincheck.*
...
+    desc: String?
+) : AdviceAdapter(ASM_API, mv, access, methodName, desc) {
 
-    override fun visitMethodInsn(opcode: Int, owner: String, name: String, desc: String, itf: Boolean) {
-        if (isCancellableContinuationGetResultMethod(owner, name)) {
+    override fun visitMethodInsn(
+        opcodeAndSource: Int,
+        className: String?,
+        methodName: String?,
+        descriptor: String?,
```
ndkoval, line 35: I've noticed that here "desc" was renamed to "descriptor", while on line 28 "descriptor" was renamed to "desc"

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -310,27 +329,39 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
...
     }
 
-    private fun transformImpl(loader: ClassLoader?, className: String, classBytes: ByteArray): ByteArray = transformedClassesCache.computeIfAbsent(className) {
-        nonTransformedClasses[className] = classBytes
+    private fun transformImpl(
+        loader: ClassLoader?,
+        internalClassName: String,
+        classBytes: ByteArray
+    ): ByteArray = transformedClassesCache.computeIfAbsent(internalClassName.canonicalClassName) {
+        nonTransformedClasses.putIfAbsent(internalClassName.canonicalClassName, classBytes.copyOf())
```
ndkoval, line 357: Should we need to call `copyOf()`? If so, please explain why in the comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -278,18 +298,17 @@ internal object LincheckJavaAgent {
...
     /**
      * FOR TEST PURPOSE ONLY!
-     * To test the byte-code transformation correctness for the
-     * model-checking strategy, we can transform all classes.
+     * To test the byte-code transformation correctness, we can transform all classes.
      */
-    internal val INSTRUMENT_ALL_CLASSES_IN_MODEL_CHECKING_MODE =
-        System.getProperty("lincheck.instrumentAllClassesInModelCheckingMode")?.toBoolean() ?: false
+    internal val INSTRUMENT_ALL_CLASSES =
+        System.getProperty("lincheck.instrumentAllClasses")?.toBoolean() ?: false
```
ndkoval, line 311: Could you please remind me why the property was renamed?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -82,9 +83,9 @@ internal object LincheckJavaAgent {
     private var isBootstrapJarAddedToClasspath = false
 
     /**
-     * TODO
+     * Names of the classes that were instrumented since the last agent installation.
```
ndkoval, line 86: Please indicate that you use `canonicalClassName` here (the same for similar usages). Please consider reflecting this in the field names.

ndkoval: Another option would be to have two value classes for that: `CanonicalClassName` and `InternalClassName`.

ndkoval: Or having one `ClassName` value class everywhere with `canonicalClassName` and `internalClassName` properties.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -192,4 +193,6 @@ private class WrapMethodInIgnoredSectionTransformer(
         }
         visitInsn(opcode)
     }
-}
\ No newline at end of file
+}
+
+internal val coroutineCallingClasses = HashSet<String>()
```
ndkoval, line 198: What is this `String`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -103,7 +104,7 @@ internal class LincheckClassVisitor(
         }
         mv = JSRInlinerAdapter(mv, access, methodName, desc, signature, exceptions)
         mv = TryCatchBlockSorter(mv, access, methodName, desc, signature, exceptions)
-        mv = CoroutineCancellabilitySupportTransformer(mv, access, methodName, desc)
+        mv = CoroutineCancellabilitySupportTransformer(mv, access, className, methodName, desc)
```
ndkoval, line 107: is this `className` internal or canonical? please indicate.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Because now stress strategy also has an optimization when it does not re-transform all the classes each time, but only:
1. subset of previously seen classes that have coroutine calls;
2. new unseen classes that were loaded into VM since the last re-transformation.

This flag now disables class re-transformation optimizations for both strategies, enforcing them to always re-transform all the loaded classes.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/CoroutineSupportTransformers.kt
eupp: Fair. I renamed all the arguments of `CoroutineCancellabilitySupportTransformer::visitMethodInsn` in accordance with the naming scheme used in all the other transformers:

```
override fun visitMethodInsn(opcode: Int, owner: String, name: String, desc: String, itf: Boolean)
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: I removed it. 

I think I've added it during debugging, but it is actually not needed.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: I added a comment explaining this variable.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Here I added the comment mentioning that it is canonical class name.

> Another option would be to have two value classes for that: CanonicalClassName and InternalClassName.

I really like this idea. But I would prefer to implement it in a separate PR. 
The reason is that there is several places when canonical/internal class names conversion happens, and I need to carefully review them all and enforce unified rules and invariants across all codebase. 

I would like to enforce the invariant that internal class names are only used in the `transformation` module, and they are always converted to canonical class names during byte-code transformation. Using two values classes can help to enforce this rule.

The reasons is that sometimes this conversion can also lead to performance degradation. 
Sometimes, the internal class name is passed into `ManagedStrategy` methods only to be immediately converted into canonical class name. Here is an example:
https://github.com/JetBrains/lincheck/blob/ecede824e9b3399a499c9efb5aba9d69dae91aa9/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt#L707

Because of this, additional redundant memory allocations happen at runtime (to create new `String`). 

But, as I said, there is several places in the code when this happens. So I think addressing this problem in a separate PR would be easier.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: I've added a comment to the `className` field. 
Please see the comment above about `CanonicalClassName` and `InternalClassName` value classes.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
ndkoval: Thank you for the detailed explanation; let's raise an issue then.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
ndkoval: I don't see the comment in the code. Could you check it on your side, please?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: Can you please check if you see this commit?
https://github.com/JetBrains/lincheck/pull/316/commits/ea912a746c3bc8ea9c72008e984922d0ed85dcb5

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: Done: https://github.com/JetBrains/lincheck/issues/328

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
ndkoval: Yes, now I see that

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
```diff
@@ -11,15 +11,12 @@
 package org.jetbrains.kotlinx.lincheck.transformation.transformers
 
 import org.jetbrains.kotlinx.lincheck.transformation.*
-import org.jetbrains.kotlinx.lincheck.transformation.ManagedStrategyMethodVisitor
```
ndkoval, line 14: Please always use "*" in imports

ndkoval (approved):

PR #322, opened 2024-05-06
Bytecode Transformation Refactoring

eupp:
This PR implements several refactorings of the byte-code transformation part of the Lincheck. 

Some of the refactorings are needed in preparation of the new model checking strategy merge, the others aim to just simplify the code. 

1. `LincheckClassVisitor.kt` is split into several files to simplify future development and reduce number of potential merge-conflicts. Independent bytecode transformers are moved into separate files and grouped logically. 

2. `ObjectAtomicWriteTrackerTransformer` is split into `AtomicFieldUpdaterTransformer`, `VarHandleTransformer` and `UnsafeTransformer` (but common methods of these transformers are still shared). This is because in the new strategy the byte-code transformation to track methods of these APIs (AFU, VH and Unsafe) is significantly more complicated. So it makes sense to split them and handle each API separatly.
    - as a part of this, I have also added handling of some missing VH and Unsafe methods, for example, `putObject`, `putReference`, `compareAndSetReference`, `compareAndSetReferenceRelease`, etc.

3. Several common code patterns are extracted into separate functions and moved into `TransformationUtils.kt`:
    -  `storeLocals(valueTypes: Array<Type>, localTypes: Array<Type>)` stores `#N` top values from the stack into local variables, performing boxing if requested by the caller.
    - `copyLocals(valueTypes: Array<Type>, localTypes: Array<Type>)` same as `storeLocals` but copies the values from the stack without modifying the stack.
    - `storeArguments(methodDescriptor: String)` and `copyArguments(methodDescriptor: String)` wrappers around `storeLocals` and `copyLocals` to save method arguments from the stack to local variables.
    - `pushNull` pushes `null` to the stack.
    - `pushArray(locals: IntArray)` wraps local variables into an array (boxing them if necessary) and pushes this array onto the stack.
  
4. Leave only one `beforeReadField` and `beforeWriteField` methods, passing `isStatic` and `isFinal` as arguments. This change allows the `ManagedStrategy` itself decide how to handle static and final fields (because in the new strategy we might also want to track accesses to final fields). 

5. Minor renamings:
  - `onMethodCallFinishedSuccessfully` --> `onMethodCallReturn`
  - `onMethodCallThrewException` --> `onMethodCallException`
  - `onWriteToObjectFieldOrArrayCell` --> `afterReflectiveSetter` (because it is only called from reflections APIs, such as VarHandle or AFU).


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -76,27 +73,30 @@ internal class LincheckClassVisitor(
         if (access and ACC_NATIVE != 0) return mv
         if (instrumentationMode == STRESS) {
             return if (methodName != "<clinit>" && methodName != "<init>") {
-                CoroutineCancellabilitySupportMethodTransformer(mv, access, methodName, desc)
+                CoroutineCancellabilitySupportTransformer(mv, access, methodName, desc)
             } else {
                 mv
             }
         }
+        val createAdapter : (MethodVisitor) -> GeneratorAdapter = {
```
avpotapov00, line 81: Let's make it as a regular nested function:

```kotlin
fun createAdapter(mv: MethodVisitor) = GeneratorAdapter(mv, access, methodName, desc)
```

avpotapov00 (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: Done

ndkoval (changes requested):

ndkoval (changes requested):

src/jvm/test/resources/expected_logs/jdk_unsafe_trace.txt
```diff
@@ -59,6 +59,8 @@ Detailed trace:
 |                                                                                                                                          |         MONITORENTER at ConcurrentHashMap.putVal(ConcurrentHashMap.java:1031)                                                         |
 |                                                                                                                                          |         ConcurrentHashMap.tabAt(Array<Node>#1,1): Node#1 at ConcurrentHashMap.putVal(ConcurrentHashMap.java:1032)                     |
 |                                                                                                                                          |           Array<Node>#1[1].getReferenceAcquire(): Node#1 at ConcurrentHashMap.tabAt(ConcurrentHashMap.java:760)                       |
+|                                                                                                                                          |         Node#1.val.READ: 2 at ConcurrentHashMap.putVal(ConcurrentHashMap.java:1040)                                                   |
```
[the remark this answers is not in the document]
ndkoval: Thanks for the investigation!

ndkoval (approved):
Approved. 

Thanks @eupp for the refactoring, and please merge the PR when you are ready.

PR #331, opened 2024-06-10
Spin locks detection enchancements

### LINKED ISSUE #218: Spin cycle start label placed incorrectly when first cycle operation is some non-atomic method call

avpotapov00:
Closes #218, #219,  adds recursive spin locks support and overall spin locks detection enhancements.


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -731,6 +734,7 @@ abstract class ManagedStrategy(
             lastReadTracePoint[iThread] = tracePoint
         }
         newSwitchPoint(iThread, codeLocation, tracePoint)
+        loopDetector.passValue(obj)
```
eupp, line 737: The fact that loop detector uses specific values to detect spin-loops is an algorithm's implementation detail. I think we can design the `LoopDetector` API to be a bit cleaner (and more flexible for potential future changes). 

Let's instead design the API in way, so that `LoopDetector` just listens for the specific events in the execution. For example, you already have the following methods in the `LoopDetector` API:
```
visitCodeLocation
beforeRegularMethodCall
afterRegularMethodCall
```

So let's just extend this API similarly for other types of events:
```
beforeReadField
beforeReadArrayElement
afterRead
...
```

Then in the implementation of these methods in `LoopDetector` you hash only the arguments that are required by the algorithm.

--

P.S.
Ideally, we should eventually converge to a design when different components of the managed strategy just implement `EventTracker` interface and subscribe for the events they are interested in (this is just an idea for the future, not for this PR).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -765,6 +771,9 @@ abstract class ManagedStrategy(
                 lastReadTracePoint[iThread] = null
             }
         }
+        runInIgnoredSection {
+            loopDetector.passValue(value)
+        }
```
eupp, line 776: Let's just wrap the entire `afterRead` into ignored section.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1396,32 +1430,300 @@ abstract class ManagedStrategy(
...
+         * 1. Take the first trace point of the spin cycle on the first iteration (point A) and on the second iteration (point B).
+         * 2. Take the last trace point before the spin cycle (point C).
+         * 3. We walk synchronously on their call stacks from the end to the beginning, while call from the A stack
+         * is the same as a call from the stack B. But it also equals the call from the call C, then we need to stop.
+         * The count of the call passed the condition above equals to the call we need to lift from the first spin cycle
+         * node to get the correct spin cycle start trace point depth.
+         *
+         * @param beforeMethodCallSwitch flag if this method invoked right after [MethodCallTracePoint] is added to a trace,
+         * before the corresponding method is called.
+         */
```
eupp, line 1630: Because all this logic is related to the spin-lock detection, can we consider moving at least part of it out of the `TraceCollector` into `LoopDetector.kt` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -222,6 +249,68 @@ internal class LoopDetector(
             Decision.Idle
     }
 
+    /**
+     * Called before regular method calls.
+     */
+    fun beforeRegularMethodCall(codeLocation: Int, params: Array<Any?>) {
```
eupp, line 255: Maybe just `beforeMethodCall`, because:
1. there is no other kinds of method calls in the `LoopDetector` API
2. to keep it in-sync with the naming scheme in `ManagedStrategy` (which uses names `beforeMethodCall` and `beforeAtomicMethodCall`)

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -19,7 +20,19 @@ import java.util.ArrayList
  *
  * The LoopDetector functions in two modes: default and replay mode.
  *
- * In default mode:
+ * In the default mode:
+ *
+ * In the default mode, LoopDetector has two states.
+ * When a spin cycle is not detected yet, LoopDetector ignores method parameters and read/write values and receivers,
+ * operating only with switch points code locations.
```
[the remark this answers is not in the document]
avpotapov00: Yes, fixed

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -222,6 +249,68 @@ internal class LoopDetector(
...
+        }
+        lastInterleavingHistoryNode.addExecution(codeLocation)
+        loopTrackingCursor.onNextExecutionPoint()
+    }
+
+    /**
+     * Called after any method calls.
+     */
+    fun afterRegularMethodCall() {
+        val afterMethodCallLocation = 0
```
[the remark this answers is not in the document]
avpotapov00: Because we want to track the fact of the method exit,
 but the sequence of the beforeMethodCall code locations values determines what exactly method we exit.
That's why we can use just 0 as a label of a method exit.

I added it as a comment.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.kt
```diff
@@ -44,7 +43,7 @@ interface EventTracker {
 
     fun afterReflectiveSetter(receiver: Any?, value: Any?)
 
-    fun beforeMethodCall(owner: Any?, className: String, methodName: String, codeLocation: Int, params: Array<Any?>)
+    fun beforeMethodCall(owner: Any?, className: String, methodName: String, codeLocation: Int, methodId: Int, params: Array<Any?>)
```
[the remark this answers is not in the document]
avpotapov00: No, it shouldn't. The only reason for methodId is to understand in case of the recursive spin lock the depth of the spin cycle start. We compare them when a walking up by the stack of the trace points. But in case of the atomicMethods no trace point can exist inside this method, so methodId is useless. That's why we don't pass it in beforeAtomicMethodCall.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -899,7 +914,7 @@ abstract class ManagedStrategy(
                         } else {
                             params
                         }
-                        beforeMethodCall(owner, currentThread, codeLocation, className, methodName, params)
+                        beforeMethodCall(owner, currentThread, codeLocation, 0, className, methodName, params)
```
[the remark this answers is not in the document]
avpotapov00: This is a "Stub" as we don't actually need methodId in case of atomic method call as there can't be any other trace points -> no possible recursive spin cycles. So it's just a workaround caused by beforeMethodCall is taking `methodId` parameter.

build.gradle.kts
```diff
@@ -20,6 +20,9 @@ plugins {
 
 repositories {
     mavenCentral()
+    flatDir {
+        dirs = mutableSetOf(File("/Users/Aleksandr.Potapov/Documents/Work/lincheck/libs"))
+    }
```
eupp, line 25: I think you've added this by mistake.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/ScalaBugTest.kt
```diff
@@ -0,0 +1,19 @@
...
+
+package org.jetbrains.kotlinx.lincheck_test.representation
+
+class ScalaBugTest {
+
+
+    fun x() {
+    }
+
+}
```
eupp, line 19: I think you've added this by mistake.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
eupp: Let's maybe extract it into a constant, like: 
```
const val ATOMIC_METHOD_ID = 0
```

and add a comment explaining why it is always zero

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -222,6 +250,92 @@ internal class LoopDetector(
...
+            return /* If we already ran into cycle and haven't switched than no need to track executions */
+        }
+        lastInterleavingHistoryNode.addExecution(codeLocation)
+        loopTrackingCursor.onNextExecutionPoint()
+    }
+
+    /**
+     * Called after any method calls.
+     */
+    fun afterRegularMethodCall() {
```
eupp, line 274: Let's rename to `afterMethodCall` for consistency with `beforeMethodCall`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -373,12 +556,63 @@ internal class LoopDetector(
...
+        // of integers fast, so we keep the contract:
+        // 1. All potential switch points code locations, method call code locations are strictly greater than 0.
+        // 2. All method exits have zero (0) code locations.
+        // 3. All parameters, receivers, and value integer representations (the things we have to ignore in case
+        // After this line [hashCode] is positive value or Integer.MIN_VALUE in case of the overflow.
+        // So by `if` below we make sure that this value is strictly negative.
+        return if (hashCode == 0) -1 else -hashCode
+    }
+
+    private enum class WorkMode {
```
eupp, line 586: maybe just `Mode` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -87,6 +87,19 @@ private fun Class<out Any>.getMethod(name: String, parameterTypes: Array<Class<o
         method.name == name && method.parameterTypes.map { it.name } == parameterTypes.map { it.name }
     } ?: throw NoSuchMethodException("${getName()}.$name(${parameterTypes.joinToString(",")})")
 
+/**
+ * @return hashcode of the unboxed value if [value] represents a boxed primitive, otherwise returns [System.identityHashCode]
+ * of the [value].
+ */
+internal fun primitiveOrIdentityHashCode(value: Any?): Int {
+    return if (value.isPrimitiveWrapper) return value.hashCode() else System.identityHashCode(value)
```
eupp, line 95: There is a redundant second return in `return value.hashCode()`

eupp (changes requested):

bootstrap/src/sun/nio/ch/lincheck/EventTracker.kt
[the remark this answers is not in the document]
eupp: Ok, I understood that you don't need the method ids of atomic methods for the spin-lock detection algorithm. 
But from the API point of view it looks a bit strange:

```kotlin
fun beforeMethodCall(owner: Any?, className: String, methodName: String, codeLocation: Int, methodId: Int, params: Array<Any?>)

fun beforeAtomicMethodCall(owner: Any?, className: String, methodName: String, codeLocation: Int, params: Array<Any?>)
``` 

We now have two similar methods, which have almost exactly the same signature, except that `beforeMethodCall` takes additional argument of `methodId`, but `beforeAtomicMethodCall` does not. 

Atomic methods can be seen as a sub-kind of all methods, so it is natural to expect that they tracker methods provide to the strategy all the information that tracker for other methods provide.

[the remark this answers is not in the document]
ndkoval: Do we need `beforeAtomicMethodCall` at all?

eupp (approved):

src/jvm/test/resources/expected_logs/obstruction_freedom_violation_with_no_detected_cycle.txt
```diff
@@ -126,5 +127,6 @@ Detailed trace:
 |     counter.incrementAndGet(): 101 at ObstructionFreedomActiveLockRepresentationTest.incrementManyTimes(ObstructionFreedomRepresentationTest.kt:44) |
 |   incrementManyTimes() at ObstructionFreedomActiveLockRepresentationTest.operation(ObstructionFreedomRepresentationTest.kt:37)                      |
 |     counter.get(): 101 at ObstructionFreedomActiveLockRepresentationTest.incrementManyTimes(ObstructionFreedomRepresentationTest.kt:41)             |
+|     counter.incrementAndGet(): 102 at ObstructionFreedomActiveLockRepresentationTest.incrementManyTimes(ObstructionFreedomRepresentationTest.kt:44) |
```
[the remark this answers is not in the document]
avpotapov00: There was a bug, this is the correct output now.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/AbstractLincheckTest.kt
```diff
@@ -68,4 +68,4 @@ abstract class AbstractLincheckTest(
     }
 }
 
-private const val TIMEOUT = 3 * 60_000L // 3 min
\ No newline at end of file
+private const val TIMEOUT = 10 * 60_000L // 10 min
```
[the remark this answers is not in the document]
avpotapov00: In the most of the builds (in teamcity and on my machine) the performance drop was insignificant. Take a look on the build times in this PR and, for example, this one: https://github.com/JetBrains/lincheck/pull/333). The reason why I increased the timeout is that the only build where we have a considerable build time increase is "Model checking tests on a single CPU". Now this build takes 13 minutes, while previously it took 9 minutes. But it's completely expected, because the tests where time increased are "SpinLockSetTest" and "ReentrantLockSetTest". Summarising: this timeout increase was for only two tests in the only one build.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
```diff
@@ -39,7 +40,7 @@ internal object CodeLocations {
     fun newCodeLocation(stackTraceElement: StackTraceElement): Int {
         val id = codeLocations.size
         codeLocations.add(stackTraceElement)
-        return id
+        return id + 1
```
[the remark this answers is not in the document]
avpotapov00: For me, using a "garbage" object looks not elegant, but if you insist, let's do it.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.kt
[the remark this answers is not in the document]
eupp: As was discussed with `@ndkoval`, we are going to remove `beforeAtomicMethodCall`, leave only `beforeMethodCall`, and detect atomic primitives dynamically. 

Another reason why we want to do that is because currently we do not handle classes inheriting from `Atomic*` classes correctly, because their methods are not detected as "atomic" statically by the bytecode transformer. Thus we want to perform the type-check dynamically. 

But all of these problems are orthogonal to this PR, I am currently working on a solution in a separate PR.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/AbstractLincheckTest.kt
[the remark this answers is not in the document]
ndkoval: Thanks for the explanation! I agree that this performance degradation is insignificant.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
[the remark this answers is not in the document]
ndkoval: BTW, why should ids start from 1, not from 0?

src/jvm/test/resources/expected_logs/obstruction_freedom_violation_with_no_detected_cycle.txt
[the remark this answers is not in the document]
ndkoval: Got it, thanks.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -434,16 +434,21 @@ internal class ModelCheckingStrategy(
...
     /**
      * This class specifies an interleaving that is re-producible.
      */
     private inner class Interleaving(
-        private val switchPositions: List<Int>,
-        private val threadSwitchChoices: List<Int>,
-        private var lastNotInitializedNode: SwitchChoosingNode?
+         private val switchPositions: List<Int>,
+         private val threadSwitchChoices: List<Int>,
+         private val initialLastNotInitializedNode: SwitchChoosingNode?
```
ndkoval, line 449: Can it be only a constructor parameter, not a field?

ndkoval:
@avpotapov00, could you please check the documentation with Grammarly?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
```diff
@@ -489,17 +506,137 @@ internal class ActorResultNode(
...
         return next
     }
 }
 
-private const val TRACE_INDENTATION = "  "
+/**
+ * Provides the prefix output for the [TraceNode].
+ * @see TraceNodePrefixFactory
+ */
+internal fun interface PrefixProvider {
```
ndkoval, line 518: Isn't it always the same string for each instance?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1742,3 +1798,8 @@ private const val INFINITE_TIMEOUT = 1000L * 60 * 60 * 24 * 365
 
 private fun getTimeOutMs(strategy: ManagedStrategy, defaultTimeOutMs: Long): Long =
     if (strategy is ModelCheckingStrategy && strategy.replay) INFINITE_TIMEOUT else defaultTimeOutMs
+
+// This is a "Stub" as we don't need methodId in case of atomic method call as there can't be
+// any other trace points -> no possible recursive spin cycles.
```
ndkoval, line 1803: > as there can't be any other trace points

What do you mean?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -927,14 +936,18 @@ abstract class ManagedStrategy(
                         LincheckJavaAgent.ensureClassHierarchyIsTransformed(className.canonicalClassName)
                     }
                 }
+                runInIgnoredSection {
+                    loopDetector.beforeMethodCall(codeLocation, params)
+                }
                 if (collectTrace) {
                     runInIgnoredSection {
+                        traceCollector!!.checkActiveLockDetected()
```
ndkoval, line 944: Why should we do this only when we are collecting the trace?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -821,6 +828,7 @@ abstract class ManagedStrategy(
             null
         }
         newSwitchPoint(iThread, codeLocation, tracePoint)
+        loopDetector.beforeWriteArrayElement(array, index, value)
```
ndkoval, line 831: Should it be called before `newSwitchPoint` or after? Why?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -797,6 +803,7 @@ abstract class ManagedStrategy(
             null
         }
         newSwitchPoint(iThread, codeLocation, tracePoint)
+        loopDetector.beforeWriteField(obj, value)
```
ndkoval, line 806: Should it be called before newSwitchPoint or after? Why?

bootstrap/src/sun/nio/ch/lincheck/EventTracker.kt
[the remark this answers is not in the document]
ndkoval: @eupp should we wait for your solution or merge this PR first?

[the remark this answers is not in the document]
eupp: I do not see any reason to wait, these two PRs are orthogonal.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RecursiveSpinLockTest.kt
```diff
@@ -0,0 +1,502 @@
...
+        .checkImpl(this::class.java)
+        .checkLincheckOutput("spin_lock/recursive_two_threads_spin_lock.txt")
+}
+
+
+/**
+ * Checks recursive spin lock representation when execution hung due to alternation of two threads
+ * in recursive live-lock.
+ */
+class BrokenCas2RecursiveLiveLockTest {
```
eupp, line 271: Let's rename into `AtomicArrayWithCAS2LiveLockTest` and together with the `AtomicArrayWithCAS2` class extract it into separate file.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RecursiveSpinLockTest.kt
```diff
@@ -0,0 +1,502 @@
...
+                if (index == index1) update1 else update2
+            } else {
+                if (index == index1) expected1 else expected2
+            }!!
+        }
+
+        /**
+         * @param broken if true then we never will go out of recursive spin-cycle
+         */
+        fun apply(initial: Boolean = true, turn: Int = -1, broken: Boolean = false) {
```
eupp, line 356: Why do you need `broken: Boolean = false` flag here?
Maybe it is better to just split this method into two methods, with one of them being "broken" (and its name should reflect this, like `applyBroken`).

eupp: In general, could you please document this test, since it is not trivial. 
In particular, to explain why you need two versions of `apply` and `installOrHelp` methods (broken one and the correct one).

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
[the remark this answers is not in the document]
avpotapov00: Because code location 0 is standing for the method exit. I've added this as a comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
avpotapov00: No, it cannot. See usages of this field.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
avpotapov00: What do you mean?

avpotapov00: It is not the same string, prefix reflects depth call and sometimes includes arrows parts of the spin cycle. See the implementation and usages.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
avpotapov00: In the current implementation it doesn't matter. But it looks more logical after the `newSwitchPoint` call as after we back to a thread we want to trace field read after the switch, not before (as it will actually happen).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
avpotapov00: In the current implementation it doesn't matter. But it looks more logical after the newSwitchPoint call as after we back to a thread we want to trace field read after the switch, not before (as it will actually happen).

bootstrap/src/sun/nio/ch/lincheck/EventTracker.kt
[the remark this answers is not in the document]
ndkoval: We have removed `beforeAtomicMethodCall` first.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
ndkoval: When you create a new PrefixProvider, does it always generate the same prefix? Or how does it work? What does `Provider` mean?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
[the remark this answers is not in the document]
ndkoval: Why can't you use another constant for that, so you don't need to play with indices here?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
ndkoval: Let's add documentation to all these paramterers, please

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
avpotapov00: Please take a look at the `org.jetbrains.kotlinx.lincheck.strategy.managed.TraceNodePrefixFactory` documentation: it is described there.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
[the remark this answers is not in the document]
avpotapov00: The negative values are reserved for the hashCodes of the parameters, values and receivers, used by LoopDetector. It's important for us to distingush them from the code locations to be able to find the cycle without considering them. (For example in case of a spin lock with random indexes).
See `org.jetbrains.kotlinx.lincheck.strategy.managed.LoopDetector` - the algorithm is explained in details there.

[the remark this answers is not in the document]
ndkoval: We discussed offline not to reserve `0`, thus avoiding changing indices back and forth. In addition, we decided to remove the performance optimization of storing both hash codes and indexes with such a contract, replacing them with wrappers (sealed class?) that clearly show possible logical values (of course, if this performance optimization is not necessary).

ndkoval (approved):
@eupp, please review the recent changes and merge the PR when it's ready.

eupp (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
```diff
@@ -30,6 +30,7 @@ internal object CodeLocations {
 
     /**
      * Registers a new code location and returns its unique ID.
+     * Code locations are positive values.
```
ndkoval, line 33: Please revert

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -19,7 +23,19 @@ import java.util.ArrayList
...
- * In default mode:
+ * In the default mode:
+ *
+ * In the default mode, LoopDetector has two states.
+ * When a spin cycle is not detected yet, LoopDetector ignores method parameters and read/write values and receivers,
+ * operating only with switch points and code locations.
+ * When a spin cycle is detected, LoopDetector requests the replay of the execution and starts tracking
+ * the parameters, etc., to detect a spin cycle period properly, taking into account all changes during the spin cycle.
+ * When we run into a spin cycle again, LoopDetector tries to find the period using [currentThreadCodeLocationsHistory].
+ * Parameters etc. are converted to the negative integer representation using hashcode (see [paramToIntRepresentation]).
```
ndkoval, line 34: You now use the `CodeIdentity` interface, please update the documentation accordingly.

PR #333, opened 2024-06-11
Reference to the Lincheck IDEA plugin added

avpotapov00:


ndkoval:
Hi @avpotapov00! Could you please add a video that illustrates the plugin and update the guide?

Issue #254, opened 2024-01-18
Coverage support


ndkoval:
Let's add some coverage support to the model-checking mode. This will provide the coverage metric and allow us to analyze which code branches were not executed, thus potentially helping us to improve test generation.

Subtasks:
- [x] Integrate Coverage and Lincheck's bytecode transformers:
   - [x] apply Lincheck transformer after coverage transformer (otherwise the coverage may incorretly report branches introduced by Lincheck transformer);
   - [x] do not instrument by Lincheck's transformer field accesses and methods injected by the coverage transformer.
- [ ] Add coverage metrics checks in our CI pipeline

ndkoval:
The issue remains open as we have no tests for this functionality. However, we are not going to add them in the foreseeable future, and we do not find this necessary. Thus, I'm closing the issue as completed.

PR #335, opened 2024-06-14
Coverage bytecode filter

dmitrii-artuhov:
---
## Notice

This PR also changes the way classes bytecode is restore to the initially loaded into Lincheck agent. Instead of `redefineClasses(...)` now `retransformClasses(...)` is used in order to avoid double-instrumentation by other attached agents (e.g. intellij-coverage agent was throwing `ClassFormatError` because of this problem).

---

## Description

Relates to #254.
 
 This PR adds a bytecode filter class, which "hides" bytecode generated by intellij-coverage agent (e.g. when using [kover](https://github.com/Kotlin/kotlinx-kover) gradle plugin) from transformation that exist in Lincheck.
 
- [x] Hide bytecode generated by default kover settings (`hits[index] = 1`)
- [x] Hide bytecode generated by default "Run with coverage" settings in Intellij IDEA (`hits[index] += 1`)

The bytecode tracking is implemented as state machine with 5 states (`+` symbol designates the direction of an arrow):
```txt
                           INITIAL  +----------------------------------------
                       /            +                                       |
                     /                \                                     |
                   +                    +                                   |
          HITS_INIT           HITS_INIT_FIELD                               |
                    \                    /                                  |
                     +                  +                                   |
                        HITS_IN_LOCAL   -------------------------------------  
                         |         +                
                         +         |                
                      HITS_BEFORE_ASSIGN 

```

Short overview:
- The state of automata is reset when starting to instrument a new method
- `HITS_INIT` and `HITS_INIT_FIELD` states occur, when the `hits` array (which stores coverage data) is loaded onto stack. The difference between state is the bytecode that they target: the 1st one occurs when `hits` was loaded via `LDC` instruction, the 2nd when `GETSTATIC` instruction was used.
- The way intellij coverage loads `hits` using `LDC` is as follows:
  ```java
  LDC __$hits$__ : Ljava/lang/Object; ...
  CHECKCAST [Z
  ASTORE #n
  ```
  and using the `GETSTATIC`:
  ```java
  GETSTATIC CLAZZ.__$hits$__ : [Z
  IFNONNULL L0
  ...
  INVOKESTATIC com/intellij/rt/coverage/instrumentation/CoverageRuntime.getHitsMask (Ljava/lang/String;)[Z
  PUTSTATIC CLAZZ.__$hits$__ : [Z
  L0
  GETSTATIC CLAZZ.__$hits$__ : [Z
  ASTORE #n
  ```
  Here `GETSTATIC` is called twice, that is why on the diagram above there is a double-sided arrow between `INITIAL` and `LOADED_ON_STACK_STATIC` states (first time `hits` is loaded on stack to check if it is null, then it is discarded, and lastly it is put on stack once again).
- `HITS_IN_LOCAL` occurs when the `hits` array is already saved to local variables of the method and the automata knows its index.
- When the `ALOAD #n` instruction occurs automata goes into `HITS_BEFORE_ASSIGN` state, then it searches for bytecode of `hits[index] = 1` or/and `hits[index] += 1`. When the assignment completed the state goes back to `HITS_IN_LOCAL`.
- When the method is fully transformed, the automata is put to `INITIAL` state again.


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -178,16 +182,11 @@ internal object LincheckJavaAgent {
...
                     return@mapNotNull null
                 }
-                // For each class, retrieve its original bytecode.
-                val bytes = nonTransformedClasses[clazz.name]
-                check(bytes != null) {
-                    "Original bytecode for the transformed class ${clazz.name} is missing!"
-                }
-                ClassDefinition(clazz, bytes)
+                clazz
             }
```
eupp, line 186: Could you please then change `mapNotNull` to `filter` and also rename `classDefinitions` into just `classes`.

eupp (approved):
All good, thank you! 

Please address the last minor comment, and then it is ready to be merged!

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
dmitrii-artuhov: done

PR #336, opened 2024-06-27
Detect atomic primitive method calls dynamically

### LINKED ISSUE #288: Don't track all Unsafe methods as atomic
Now we treat all methods of Unsafe as atomic methods. Let's do it only for the necessary ones.

eupp:
Currently, we have two injection methods to handle method calls: `beforeMethodCall` and `beforeAtomicMethodCall`. 
This solution has several problems:

1. Code duplication.

2. We do not handle classes inheriting from `Atomic*` classes propertly, because their methods are not detected as "atomic" statically by the `MethodCallTransformer` bytecode transformer that injects `beforeMethodCall` and `beforeAtomicMethodCall` invocations. This is because this transformer decides what method to insert statically based on the class name of the invoked method. It checks only for `j.u.c.Atomic*` method names, and do not handles subclasses of atomics. 

This PR removes `beforeAtomicMethodCall` injection, letting the managed strategy to decide how to handle the method dynamically at runtime, based on the actual type of the method receiver.

I've also added a new test `AtomicSubclassTest.kt` checking that subclasses of `Atomic*` classes are properly supported. 

Please note, that with this PR the memory access methods of `Atomic*`, `Atomic*FieldUpdater`, `VarHandle` and `Unsafe` APIs are treated as atomic. Therefore, this PR also closes #288.

The latter also implies that `Atomic*Array::length()` method is no longer considered to be atomic method. 
Because of this, it has disappeared from the trace output of some tests.


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
```diff
@@ -173,39 +118,32 @@ internal class MethodCallTransformer(
...
+        className == "java/lang/Float" ||
+        className == "java/util/Locale" ||
+        className == "org/slf4j/helpers/Util" ||
+        className == "java/util/Properties"
 
-    private fun isAtomicArrayClass(className: String) =
-        className == "java/util/concurrent/atomic/AtomicIntegerArray" ||
-        className == "java/util/concurrent/atomic/AtomicLongArray" ||
-        className == "java/util/concurrent/atomic/AtomicReferenceArray"
+    private fun isExceptionThrowingMethod(className: String, methodName: String) =
```
ndkoval, line 138: This is a performance optimization. Do we need it?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
```diff
@@ -0,0 +1,288 @@
...
+import org.jetbrains.kotlinx.lincheck.util.AtomicMethodKind.*
+import org.jetbrains.kotlinx.lincheck.util.MemoryOrdering.*
+import java.lang.invoke.VarHandle
+
+internal data class AtomicMethodDescriptor(
+    val kind: AtomicMethodKind,
+    val ordering: MemoryOrdering,
+)
+
+internal enum class AtomicMethodKind {
```
ndkoval, line 23: Do I understand correctly that you've added atomic method kinds for the upcoming model checker?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
```diff
@@ -0,0 +1,288 @@
...
+    override fun toString(): String = when (this) {
+        PLAIN       -> "Plain"
+        OPAQUE      -> "Opaque"
+        RELEASE     -> "Release"
+        ACQUIRE     -> "Acquire"
+        VOLATILE    -> "Volatile"
+    }
+}
+
+internal fun getAtomicMethodDescriptor(obj: Any?, className: String, methodName: String): AtomicMethodDescriptor? {
```
ndkoval, line 46: className is unused

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -871,76 +871,43 @@ abstract class ManagedStrategy(
         return null
     }
 
+    /**
+     * This method is invoked by a test thread before each method invocation.
```
ndkoval, line 875: This documentation (including the parameters) looks outdated

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
eupp: I am not sure. I added it so the code functionally remains the same as before.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
eupp: Yes, for this too. 

When implementing this PR I went once again through the javadocs of all related APIs, so I decided to set-up the infrastructure to avoid doing this again in the future.

ndkoval: Sounds reasonable 👍

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
ndkoval: Could you investigate, please? The shorter the code the better.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: I decided to remove this kdoc, because
1. It mostly duplicates kdoc of `Injections::beforeMethodCall`
2. There is no documentation for other methods of `ManagedStrategy` like `beforeReadField`. 
3. The documentation should be put in `EventTracker` interface anyway, because it is the base interface.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
eupp: I checked on my machine. I do not observe any difference, so I removed special handling of atomic methods.

ndkoval (approved):

PR #339, opened 2024-07-09
AllowExtraSuspension parameter deprecated, SUSPENDED prefix removed from the output

### LINKED ISSUE #319: Always allow extra suspensions and do not show "SUSPENDED" in the result when an operation completes
Let's deprecate the `allowExtraSuspension` parameter

avpotapov00:
Closes #319


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/AllowExtraSuspensionTest.kt
```diff
@@ -36,23 +36,6 @@ class AllowExtraSuspensionCorrectTest : AbstractLincheckTest() {
     }
 }
 
-class AllowExtraSuspensionIncorrectTest : AbstractLincheckTest(IncorrectResultsFailure::class) {
```
ndkoval, line 39: Please keep this test, it should still pass

src/jvm/main/org/jetbrains/kotlinx/lincheck/verifier/LTS.kt
```diff
@@ -121,19 +121,19 @@ class LTS(private val sequentialSpecification: Class<*>) {
...
 
-        private fun Result.isLegalByFollowUp(transitionInfo: TransitionInfo, allowExtraSuspension: Boolean) =
+        private fun Result.isLegalByFollowUp(transitionInfo: TransitionInfo) =
             this == transitionInfo.result ||
             this is ValueResult && transitionInfo.result is ValueResult && this.value == transitionInfo.result.value &&
-                (!wasSuspended && transitionInfo.result.wasSuspended || wasSuspended && allowExtraSuspension) ||
+                (!wasSuspended && transitionInfo.result.wasSuspended || wasSuspended) ||
             this is ExceptionResult && transitionInfo.result is ExceptionResult && this.tClassCanonicalName == transitionInfo.result.tClassCanonicalName &&
-                (!wasSuspended && transitionInfo.result.wasSuspended || wasSuspended && allowExtraSuspension) ||
+                (!wasSuspended && transitionInfo.result.wasSuspended || wasSuspended) ||
```
ndkoval, line 134: I don't think you need the `wasSuspended` marker anymore {---} operations either complete with a result, get canceled, or get suspended and never finish.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
```diff
@@ -24,36 +24,30 @@ import kotlin.coroutines.*
...
-    override val wasSuspended get() = false
-    override fun toString() = wasSuspendedPrefix + VOID
+object VoidResult : Result {
+    override fun toString() = VOID
 }
 
-object SuspendedVoidResult : Result() {
-    override val wasSuspended get() = true
-    override fun toString() = wasSuspendedPrefix + VOID
+object SuspendedVoidResult : Result {
```
ndkoval, line 43: Do you need it anymore?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
```diff
@@ -24,36 +24,30 @@ import kotlin.coroutines.*
...
+    override fun toString() = VOID
 }
 
 private const val VOID = "void"
 
-object Cancelled : Result() {
-    override val wasSuspended get() = true
-    override fun toString() = wasSuspendedPrefix + "CANCELLED"
+object Cancelled : Result {
+    override fun toString() ="CANCELLED"
```
ndkoval, line 50: Should it remain SUSPENDED + CANCELLED? What are the pros and cons?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
```diff
@@ -64,63 +58,56 @@ class ExceptionResult private constructor(
      * Exception is stored to print it's stackTrace in case of incorrect results
      */
     val throwable: Throwable,
-    override val wasSuspended: Boolean,
     /**
      * Normalized version of the exception class
```
ndkoval, line 62: Do we need it with the javaagents being merged?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
```diff
@@ -64,63 +58,56 @@ class ExceptionResult private constructor(
...
-object NoResult : Result() {
-    override val wasSuspended get() = false
+object NoResult : Result {
     override fun toString() = "-"
 }
 
-object Suspended : Result() {
-    override val wasSuspended get() = true
+object Suspended : Result {
     override fun toString() = "S"
```
ndkoval, line 103: Why "S"?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -102,21 +102,21 @@ private fun Class<out Any>.getMethod(name: String, parameterTypes: Array<Class<o
...
 }
 
 private fun kotlin.Result<Any?>.toLinCheckResult(wasSuspended: Boolean) =
     if (isSuccess) {
         when (val value = getOrNull()) {
             is Unit -> if (wasSuspended) SuspendedVoidResult else VoidResult
             // Throwable was returned as a successful result
-            is Throwable -> ValueResult(value::class.java, wasSuspended)
-            else -> ValueResult(value, wasSuspended)
+            is Throwable -> ValueResult(value::class.java)
```
ndkoval, line 116: Shouldn't it be ExceptionResult?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
avpotapov00: Nope, this is a case when Throwable was returned as a successful result, see comment above

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
avpotapov00: I don't see anything useful in printing "SUSPENDED +" at all, because this information can be extracted from the trace. In the results it looks redundant and may confuse a user.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
ndkoval: Why should you extract a class in one case and a value in another? `ExceptionResult` should be a successful one as well.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Result.kt
```diff
@@ -64,63 +54,55 @@ class ExceptionResult private constructor(
...
      */
     val throwable: Throwable,
-    override val wasSuspended: Boolean,
     /**
-     * Normalized version of the exception class
+     * Exception class canonical name
      */
-    tClazz: Class<out Throwable>,
-) : Result() {
+    internal val tClassCanonicalName: String
```
ndkoval, line 60: You don't need to pass it to the constructor. If needed, the class name is accessible from `throwable`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
ndkoval: I suppose the current code is (and was) buggy here.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
avpotapov00: Ok, I agree, let's treat this case as a regular one, creating `ValueResult(value)` here.

ndkoval (approved):

PR #344, opened 2024-07-22
Object numeration mismatch fix

### LINKED ISSUE #343: Object numeration doesn't correspond between plugin and text output
Consider the following example and the comments in the code:

```kotlin
class MyTest {
    private var ref1 = Node(1)
    private var ref2: Node? = null
    private var c: Int = 0

    fun increment(): Int {
        // When the plugin is enabled, before this instruction  method`org.jetbrains.kotlinx.lincheck.IdeaPluginKt.visualize` is called,
        // and it will recursively traverse all the MyTest instance to enumerate objects and find out ref1 -> Node(1) object, so it would have number 1.
        // when plugin is disabled, we won't even consider this object as it hasn't been involved in the execution yet.
        c++
        // Here, when the plugin is enabled, in `beforeWriteField` we'll give `Node(0)` object number 2, as 
        // we already have seen an object of type `Node`. When the plugin is disabled, this object will be the 
        // first instance of `Node` type and will get the number 1.
        ref2 = Node(0)
        return c
    }

    @Test
    fun test() = ModelCheckingOptions()
        .addCustomScenario {
            parallel {
                thread { actor(::increment) }
                thread { actor(::increment) }
            }
        }.check(this::class)

    class Node(val value: Int)

}
```

So, generally, the problem is that when the Plugin is enabled, we may see some objects before they appear in the trace. Therefore, a smaller number will be assigned to them, and a bigger number will be assigned to the next object of the same type in the trace.

avpotapov00:
closes #343


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/TransformationUtils.kt
```diff
@@ -231,7 +231,9 @@ internal fun GeneratorAdapter.invokeBeforeEvent(debugMessage: String, setMethodE
             push(debugMessage)
             invokeStatic(Injections::beforeEvent)
         },
-        elseClause = {}
+        elseClause = {
+            invokeStatic(Injections::enumerateObjectsIfNeeded)
```
ndkoval, line 235: This logic contradicts with the `invokeBeforeEvent` function name

src/jvm/test/resources/expected_logs/obstruction_freedom_violation_events_cut.txt
```diff
@@ -10,12 +10,12 @@ The following interleaving leads to the error:
 |                                                      Thread 1                                                       |                                                      Thread 2                                                      |
 | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 |                                                                                                                     | enqueue(-1)                                                                                                        |
-|                                                                                                                     |   q.enqueue(-1) at ObstructionFreedomViolationEventsCutTest.enqueue(SpinlockEventsCutTests.kt:34)                  |
+|                                                                                                                     |   q.enqueue(-1) at ObstructionFreedomViolationEventsCutTest.enqueue(SpinlockEventsCutTests.kt:33)                  |
```
ndkoval, line 13: Is there any difference besides the line numbers? Please check other outputs as well.

ndkoval (changes requested):
Hi @avpotapov00! I've reviewed the PR and left some comments. Could you please also rebase it onto the current `develop`?

ndkoval:
@avpotapov00 could you please rebase the PR onto `develop` and address the comments?

ndkoval:
@eupp should we close this PR without merging it?

eupp:
> should we close this PR without merging it?

Yes, the whole approach to object numeration was rethought in https://github.com/JetBrains/lincheck/pull/702

PR #348, opened 2024-07-28
Multiple suspension points support

### LINKED ISSUE #122: Support operations with multiple suspension points
See #121 for an example of Lincheck failing in the case of multiple suspension points per operation

### LINKED ISSUE #347: Support multiple suspension points

avpotapov00:
Closes #122


eupp:
We discussed this PR today with @avpotapov00 , and come to the conclusion that it overlaps with #361. 

This PR implements three things:
1. Fix for events tracking in a resumption part of a coroutine.
2. Fix for the trace collection in a resumption part of a coroutine. 
3. Multiple suspension points support. 

Independently, while working on the new model checking strategy, I also discovered problems (1) and (2), and opened PR #361.

At first I was only aware that this PR fixes (1), so I borrowed the relevant code in the commit [ab760ef](https://github.com/JetBrains/lincheck/pull/361/commits/ab760efd7b3a52579d8f95caebd110b3e44dcc46). 

Then I independently developed different fix for (2). 
As we discussed today, this fix also handles more complicated case, when the resumed part of the coroutine has several nested `suspend` functions.  To handle this case, I save the stack of suspended functions, and then restore it on resumption. 

So after the discussion, we thought that it would be good idea to split this PR into two parts. 

1. The first part should contain fixes for problems (1) and (2), and will be delivered under #361 
2. The second part should contain the multiple suspension points logic, and should be delivered after the first part.

eupp:
@avpotapov00 @ndkoval I've rebased the PR on `develop`, please have a look.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -258,16 +258,25 @@ internal open class ParallelThreadsRunner(
...
-            // to track the events inside resumption.
+        // by the resuming thread, and invoke the follow-up part in this case
+        val suspendResultToContinuation = completion.resWithCont.get()
+        if (suspendResultToContinuation !== null) {
+            // Suspended thread got result of the suspension point and continuation to resume
+            val (resumedValue, continuation) = suspendResultToContinuation
+            // Erase the current result
+            completion.resWithCont.set(null)
+            // We must exit the ignored section to keep tracking execution after the resumption.
             runOutsideIgnoredSection(thread) {
```
ndkoval, line 269: Why should this code be called outside the ignored section?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -970,19 +970,25 @@ abstract class ManagedStrategy(
...
             }
             if (collectTrace) {
                 traceCollector!!.checkActiveLockDetected()
-                addBeforeMethodCallTracePoint(owner, codeLocation, methodId, className, methodName, params, atomicMethodDescriptor)
+                addBeforeMethodCallTracePoint(threadId, owner, codeLocation, methodId, className, methodName, params,
+                    atomicMethodDescriptor
+                )
             }
-            if (guarantee == ManagedGuaranteeType.TREAT_AS_ATOMIC) {
+            if (guarantee == ManagedGuaranteeType.TREAT_AS_ATOMIC &&
```
ndkoval, line 988: Could you please add a comment explaining what is going on here?

ndkoval (changes requested):

ndkoval (approved):

PR #357, opened 2024-08-07
Fix byte array element type in the bytecode transformation code

eupp:
In JVM, `baload` and `bastore` instructions are used to access both byte arrays and boolean arrays. 
In the `SharedMemoryTransformer` class, the instruction is used to determine the type of elements read from an array. 
This type is then used to box the read value and pass it into the injected method `afterRead(value)`, which, in turn, stores the value into trace to display it later. 

For `baload` and `bastore` instructions, the `BOOLEAN_TYPE` was used. Because of this, in the trace the byte values could have been displayed as boolean, i.e. narrowed to `true`/`false`.  This PR fixes this problem by using `BYTE_TYPE` instead. 

However, now boolean values are represented in the trace as byte values `0` or `1`. This is a lesser problem. 
It is unclear how to distinguish byte/boolean in this situation, due to a dynamic nature of `baload`/`bastore` instructions which are used for both these types.


ndkoval:
I would assume that `boolean` is used much more often than `byte`, and presenting it correctly is critical. We probably can access the array type in the transformer.

eupp:
> We probably can access the array type in the transformer.

You right. I pushed the required fixes, now `BYTE_TYPE` and `BOOLEAN_TYPE` are properly distinguished. Please have a look.

ndkoval (approved):
Super, thank you! I've approved the changes.

PR #361, opened 2024-08-15
Fix resumption trace collection and printing logic

eupp:
Events from coroutine resumptions were not printed before. 
Moreover, the "stack trace recovery after resumption" logic seem to be broken too.

This PR fixes these problems. 

See the output of the affected tests to spot the difference. 
One new test with several nested `suspend fun`-s was also added, to test the case of more complicated resumption logic.


src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -221,7 +221,7 @@ internal open class ParallelThreadsRunner(
                     completedOrSuspendedThreads.incrementAndGet()
                 }
                 Cancelled
-            } else waitAndInvokeFollowUp(iThread, actorId)
+            } else waitAndInvokeFollowUp(t, actorId)
```
ndkoval, line 224: Let's rename `t` to `thread` for clarity?

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -234,7 +234,8 @@ internal open class ParallelThreadsRunner(
 
     // We need to run this code in an ignored section,
     // as it is called in the testing code but should not be analyzed.
-    private fun waitAndInvokeFollowUp(iThread: Int, actorId: Int): Result = runInIgnoredSection {
+    private fun waitAndInvokeFollowUp(thread: TestThread, actorId: Int): Result = runInIgnoredSection {
+        val iThread = thread.threadId
```
ndkoval, line 238: `iThread` -> `threadId`

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -261,7 +262,9 @@ internal open class ParallelThreadsRunner(
         if (completion.resWithCont.get() !== null) {
             // Suspended thread got result of the suspension point and continuation to resume
             val resumedValue = completion.resWithCont.get().first
-            completion.resWithCont.get().second.resumeWith(resumedValue)
+            runWithoutIgnoredSection(thread) {
```
ndkoval, line 265: Please leave a comment why `runWithoutIgnoredSection` should be here. Also, consider renaming the function to `runOutsideIgnoredSection`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -103,7 +103,9 @@ internal class ModelCheckingStrategy(
         // as the instrumented code might call `enterIgnoredSection` just before this call.
         return replay && collectTrace &&
                 Thread.currentThread() is TestThread &&
-                suddenInvocationResult == null
+                suddenInvocationResult == null &&
+                !shouldSkipNextBeforeEvent()
```
ndkoval, line 107: Why would you need to skip the next beforeEvent? Why should it be added now?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -914,13 +914,13 @@ internal fun afterSpinCycleTraceCollected(
         var firstI = spinCycleFirstTracePointCallStackTrace.lastIndex
         var count = 0
         while (firstI >= 0) {
-            val identifier = spinCycleFirstTracePointCallStackTrace[firstI].methodId
+            val identifier = spinCycleFirstTracePointCallStackTrace[firstI].methodInvocationId
             // Comparing corresponding calls.
-            if (identifier != currentCallStackTrace[currentI].methodId) break
+            if (identifier != currentCallStackTrace[currentI].methodInvocationId) break
```
ndkoval, line 919: `methodInvocation` or `methodCall`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -277,4 +284,25 @@ private inline fun <R> runInIgnoredSection(currentThread: Thread, block: () -> R
...
     }
 
+/**
+ * Exits the ignored section and invokes the provided [block] in the ignored section, setting
+ * the [TestThread.inIgnoredSection] flag to `false` in the beginning and setting it back
+ * in the end to `true`.
+ * This method **must** be called in an ignored section.
+ */
+@Suppress("UnusedReceiverParameter")
+internal inline fun <R> ParallelThreadsRunner.runWithoutIgnoredSection(currentThread: TestThread, block: () -> R): R {
```
ndkoval, line 294: Consider renaming to `runOutsideIgnoredSection`

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -277,4 +284,25 @@ private inline fun <R> runInIgnoredSection(currentThread: Thread, block: () -> R
...
+    if (!currentThread.inTestingCode) {
```
ndkoval, line 295: Can this function be called outside the testing code?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -92,17 +91,19 @@ abstract class ManagedStrategy(
     // Collector of all events in the execution such as thread switches.
     private var traceCollector: TraceCollector? = null // null when `collectTrace` is false
 
+    // Stores the global number of stack trace elements.
+    private var callStackTraceElementNextId = 0
```
ndkoval, line 95: why `nextId`?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
eupp: Currently, we use it only in single place (for coroutine resumption). 
Here it is guaranteed that the thread is in testing code. 
So I changed the `if` to `require`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Renamed to `callStackTraceElementId`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -47,7 +46,7 @@ abstract class ManagedStrategy(
     scenario: ExecutionScenario,
     private val validationFunction: Actor?,
     private val stateRepresentationFunction: Method?,
-    private val testCfg: ManagedCTestConfiguration
+    private val testCfg: ManagedCTestConfiguration,
```
ndkoval, line 49: unnecessary change, let's revert it

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1048,39 +1077,76 @@ abstract class ManagedStrategy(
...
             methodParams
         }
-        // Code location of the new method call is currently the last one
-        val tracePoint = createBeforeMethodCallTracePoint(owner, iThread, className, methodName, params, codeLocation, atomicMethodDescriptor)
-        methodCallTracePointStack[iThread] += tracePoint
-        // Method id used to calculate spin cycle start label call depth.
+        // The code location of the new method call is currently the last one
+        val tracePoint = createBeforeMethodCallTracePoint(
+            iThread,
+            owner,
```
ndkoval, line 1125: Please add argument names

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
eupp: This is essentially a hack to make the plugin integration working without refactoring too much code.

So the problem is because of the coroutine resumptions. 
We don't have coroutine resumptions events as trace points in the trace. Thus they don't have "eventId"-s which are used to integrate Lincheck with the plugin. 

However, when the resumption is called, a lot of stuff is happening under the hood. 
In particular, a `suspend fun` is called with the given completion object.
`MethodCallTransformer` instruments this call as it instruments other method calls, however, in case of resumption this `suspend fun` call should not create new trace point. 
`MethodCallTransformer` does not know about it, it always injects `beforeEvent` call regardless:

```
invokeStatic(Injections::beforeMethodCall)
invokeBeforeEventIfPluginEnabled("method call $methodName", setMethodEventId = true)  
```

Therefore, to "skip" this `beforeEvent` call following resumption call to `suspend fun`, we use this `skipNextBeforeEvent` flag. 


IMO, a better approach would be to refactor a code, and instead just assign `eventId`-s directly to trace points. 
Methods like `beforeMethodCall` then can return `eventId` of the created trace point, or something like `-1` in case when no trace point is created. Then subsequent `beforeEvent` call can just take this `eventId` from the stack. 

But this would be quite a large refactoring, I think it's better to implement it as separate PR. 

@avpotapov00 WDYT?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
eupp: I would prefer `methodInvocationId` or `methodCallId` name. 

The reason is that `methodInvocationId` is just an `Int` -- it is an identifier of method call, not a method call itself. 
The identifier is computed as hash-code of the method id and the hash-codes of all its arguments.

In case of `methodInvocation` name, I would expect the field to store the information about the method call itself, that is method name and all its arguments (as objects, not as their hash-codes).

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
eupp: Oh, sorry, I was mistaken. This function can actually be called outside of testing code in case of stress strategy. Let's just leave this `if` as is.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1048,39 +1077,76 @@ abstract class ManagedStrategy(
...
+            methodInvocationId = methodInvocationId
+        )
+        callStackTrace.add(stackTraceElement)
+        beforeMethodEnter(owner)
     }
 
     private fun createBeforeMethodCallTracePoint(
-        owner: Any?,
         iThread: Int,
+        owner: Any?,
```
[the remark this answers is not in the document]
eupp: For consistency with other methods in `ManagedStrategy`, which usually take `iThread` as a first argument. 

So `iThread` goes as first argument, then there goes a bunch of arguments related to method call: `owner`, `className`, `methodName`, `params`; and then other arguments.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1260,12 +1327,12 @@ abstract class ManagedStrategy(
 
     /* Methods to control the current call context. */
 
-    private fun beforeStaticMethodCall() {
-        callStackContextPerThread[currentThread].add(CallContext.StaticCallContext)
-    }
-
-    private fun beforeInstanceMethodCall(receiver: Any) {
-        callStackContextPerThread[currentThread].add(CallContext.InstanceCallContext(receiver))
+    private fun beforeMethodEnter(receiver: Any?) {
```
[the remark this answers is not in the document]
eupp: Object whose method is called. 

Should I rename it? Would `owner` or just `obj` be better?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
ndkoval: I'm OK with this hack, please document it in the source code.

eupp: Agreed to leave this for now. Should add a comment to source code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
ndkoval: Let's use `owner` or `obj`, similarly to other places in Lincheck

ndkoval (approved):

ndkoval:
@eupp Approved. Should we wait for a review by @avpotapov00?

eupp:
Yes, he said he want to have a look.

avpotapov00 (approved):

Issue #340, opened 2024-07-15
PriorityBlockingQueue - problem with getting the field offset on a hidden class.


Asinrus:
Hello Lincheck Team
Thank you for your product. It is a great tool. 

I'm reaching out to you to ask about a strange behavior that I hit during the implementation of an LFU cache and I thought you would find it interesting. 
Tech - java-correto-17.0.8 , lincheck-2.32
My local setup - MacOs 14.5 (23F79), Apple M1 Pro, 
Link to the repo  - https://github.com/Asinrus/lincheck-priority-queue-check 

As you can see in class CacheLFUService I made a simple implementation with PriorityBlockingQueue and ConcurrentHashMap. 

To check the correctness of my program I made two tests: 
1. me.asinrus.CacheLFUServiceTest#modelCheckingTest2threads - the test is passed.
2. me.asinrus.CacheLFUServiceTest#modelCheckingTest3threads - test is failing 
Additionally in this case, I see in the logs 
```
java.lang.UnsupportedOperationException: can't get field offset on a hidden class: private final java.util.function.ToIntFunction java.util.Comparator$$Lambda$273/0x000000f80106f0c0.arg$1
	at jdk.unsupported/sun.misc.Unsafe.objectFieldOffset(Unsafe.java:645)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.FieldSearchHelper.findObjectField(FieldSearchHelper.kt:109)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.FieldSearchHelper.findObjectField(FieldSearchHelper.kt:79)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.FieldSearchHelper.findObjectField(FieldSearchHelper.kt:79)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.FieldSearchHelper.findObjectField(FieldSearchHelper.kt:79)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.FieldSearchHelper.findFinalFieldWithOwner$lincheck(FieldSearchHelper.kt:39)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.findOwnerName(ManagedStrategy.kt:1305)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.createBeforeMethodCallTracePoint(ManagedStrategy.kt:1166)
[...]

PR #367, opened 2024-08-22
In `FieldSearchHelper` skip protected or private fields that cannot be read by `Unsafe`

eupp:


eupp:
Should fix #340

ndkoval (changes requested):

ndkoval:
@eupp could you please check the red builds?

eupp:
It turns out `Class::isHidden()` and `Class::isRecord()` were introduced only in Java 15. 
So I rollbacked to the exception-catching solution.

ndkoval:
@eupp, please add a comment to the code explaining the logic.

eupp:
> @eupp, please add a comment to the code explaining the logic.

@ndkoval, done

ndkoval (approved):

PR #406, opened 2024-09-30
Creating a thread actor with `blocking = true`

de-shyt:


ndkoval:
@de-shyt please include only your changes in the PR

Issue #404, opened 2024-09-30
"Count 'n' must be non-negative, but was -4"


de-shyt:
When `invocationsPerIteration` value in the ModelChecking test is between 13000 and 15000, the `IllegalArgumentException` error occurs:

```
Count 'n' must be non-negative, but was -4.
java.lang.IllegalArgumentException: Count 'n' must be non-negative, but was -4.
	at kotlin.text.StringsKt__StringsJVMKt.repeat(StringsJVM.kt:788)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceNodePrefixFactory.prefixForNode$lambda$5(TraceReporter.kt:628)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceNode.getPrefix(TraceReporter.kt:354)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.CallNode.addRepresentationTo(TraceReporter.kt:461)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.traceGraphToRepresentationList(TraceReporter.kt:343)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.appendShortTrace(TraceReporter.kt:43)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.appendTrace(TraceReporter.kt:31)
	at org.jetbrains.kotlinx.lincheck.ReporterKt.appendFailure(Reporter.kt:392)
	at org.jetbrains.kotlinx.lincheck.strategy.LincheckFailure.toString(LincheckFailure.kt:22)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.tryCollectTrace(ManagedStrategy.kt:269)
	at org.jetbrains.kotlinx.lincheck.strategy.StrategyKt.verify(Strategy.kt:127)
	at org.jetbrains.kotlinx.lincheck.strategy.StrategyKt.runIteration(Strategy.kt:108)
	at org.jetbrains.kotlinx.lincheck.LinChecker.run(LinChecker.kt:147)
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl(LinChecker.kt:104)
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl$lincheck(LinChecker.kt:65)
	at org.jetbrains.kotlinx.lincheck.LinChecker.check(LinChecker.kt:47)
[...]

PR #411, opened 2024-10-07
API for the general-purpose model checker

dmitrii-artuhov:
### Description:
PR adds function for running lincheck model checker algorithm on arbitrary code via lambda body.
Usage example:
```kotlin
class JunitTest {
  val invocations = 100
  @Test
  fun test() = runConcurrentTest(invocations) {
    var counter = 0
    val t1 = thread { counter++ }
    val t2 = thread { counter++ }
    t1.join(); t2.join()
    check(counter == 2) // will detect an error an throw an exception
  }
}
```

#### Additional improvements to the PR:
- [x] make a different default invocations count (more than lincheck does for data-structures testing)
- [x] make `runConcurrentTest` throw `AssertionError` instead of returning it
- [ ] tests with simple coroutines usage
- [ ] tests with thread pool (created inside a lambda body): see https://github.com/JetBrains/lincheck/pull/447
- [ ] tests on java - bug arised ⚰️: see https://github.com/JetBrains/lincheck/issues/445


src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+                listOf(
+                    Actor(
+                        Wrapper<R>::run.javaMethod!!,
+                        listOf(block)
+                    )
+                )
+            ),
+            emptyList(),
+            null
+        )
```
eupp, line 43: Please use [`CustomScenarioDSL.kt`](https://github.com/JetBrains/lincheck/blob/master/src/jvm/main/org/jetbrains/kotlinx/lincheck/CustomScenarioDSL.kt) instead.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+            val verifier = testCfg.createVerifier()
+
+            for (i in 1..invocations) {
+                if (!strategy.nextInvocation()) {
+                    break
+                }
+
+                val result = strategy.runInvocation()
+                val failure = strategy.verify(result, verifier)
+
```
eupp, line 66: Why so many empty lines for just a few lines of code ?)

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingOptions
+import org.jetbrains.kotlinx.lincheck.strategy.verify
+import org.jetbrains.kotlinx.lincheck.transformation.withLincheckJavaAgent
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import kotlin.reflect.jvm.javaMethod
+
+
+object Lincheck {
+    @JvmStatic
+    fun <R> verifyWithModelChecker(
```
eupp, line 27: Please add a kdoc explaining the usage of this function. In particular, mention that this function looks for exceptions being thrown inside the user provided lambda `block`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+                val failure = strategy.verify(result, verifier)
+
+                if (failure != null) {
+                    return failure
+                }
+            }
+        }
+
+        return null
+    }
```
eupp, line 74: Please add tests for this function. 
For start, add few tests in `representation` sub-module and check that the trace is correctly collected.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+            .addGuarantee(forClasses(Lincheck::class).allMethods().ignore())
+            .addGuarantee(forClasses(Wrapper::class).allMethods().ignore())
+            .verifier(ExecutionExceptionsVerifier::class.java)
+
+        val testCfg = options.createTestConfigurations(Wrapper::class.java)
+
+        withLincheckJavaAgent(testCfg.instrumentationMode) {
+            val strategy = testCfg.createStrategy(Wrapper::class.java, scenario, null, null)
+            val verifier = testCfg.createVerifier()
+
```
eupp, line 58: BTW, I think somewhere here you need to enforce the bytecode transformation of the `block::class.java` (i.e. the class of the passed lambda). Otherwise the events inside the lambda would not be tracked. 

I think once you'll add the tests you will see the problem.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    internal class Wrapper<R>() {
+        fun run(block: () -> R) = block()
+    }
+
+    private class ExecutionExceptionsVerifier(@Suppress("UNUSED_PARAMETER") sequentialSpecification: Class<*>) : Verifier {
+        override fun verifyResults(scenario: ExecutionScenario?, results: ExecutionResult?): Boolean {
+            if (results == null) return true
+            // User `check(...)`'s will fail and put their exceptions as an execution result, we must track that and disallow such outcomes.
```
eupp, line 83: Please move this explanation to the kdoc of `ExecutionExceptionsVerifier`.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,93 @@
...
+     * and track exceptions thrown from it (threads are created by the user in the [block]).
+     *
+     * @param invocations number of different interleavings of code in the [block]
+     * that should be explored.
+     * @param block lambda which body will be a target for the interleavings exploration.
+     */
+    @JvmStatic
+    fun <R> verifyWithModelChecker(
+        invocations: Int = ManagedCTestConfiguration.DEFAULT_INVOCATIONS,
+        verifierClass: Class<out Verifier>? = null,
```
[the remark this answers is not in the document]
dmitrii-artuhov: I need to somehow allow lincheck to differentiate failed `check(...)` invocation from lincheck source code and from the `runWithModelChecker` function. I can add a new internal custom execption for InternalLincheckError and substitute all internal `check(...)` with `checkInternal(...)` or something like that. Are there any better solutions?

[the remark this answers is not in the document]
eupp: You can either:
- omit this problem for now;
- try to use `isInternalLincheckBug()` function: 
https://github.com/JetBrains/lincheck/blob/a3523c94e4c08ec30d8df860a91525c7ddd54133/src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt#L616

Also please have a look at the following issue: https://github.com/JetBrains/lincheck/issues/440

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,91 @@
...
+object Lincheck {
+    /**
+     * This method will explore different interleavings of the [block] body and all the threads created within it,
+     * searching for the first raised exception.
+     *
+     * @param invocations number of different interleavings of code in the [block] that should be explored.
+     * @param block lambda which body will be a target for the interleavings exploration.
+     */
+    @JvmStatic
+    fun <R> verifyWithModelChecker(
```
eupp, line 36: As discussed with @ndkoval . 

For now, lets rename this function into `runConcurrentTest`, and make it top-level function (no `Lincheck` object needed). This is not a final name, we will discuss API of this function (including its name) in more detail before public release of this feature.

src/jvm/test-jdk11/org/jetbrains/kotlinx/lincheck_test/representation/RunWithLambdaRepresentationTests.kt
```diff
@@ -0,0 +1,154 @@
...
+    abstract fun block()
+
+    @Test
+    fun testRunWithModelChecker() {
+        val failure = Lincheck.verifyWithModelChecker {
+            block()
+        }
+        failure.checkLincheckOutput(outputFileName)
+    }
+}
```
eupp, line 39: There are also tests in `CustomThreadsRepresentationTest.kt` file that I have added when I was working on supporting custom threads. Could you please move them in this file, rewrite them to use `verifyWithModelChecker` function and check that they would still work?

eupp (changes requested):

dmitrii-artuhov:
I added the `testClassInstance` parameter to the `runConcurrentTest` function, because right now I utilize a specifal wrapper class which calls the lambda and only this wrapper is recreated between iterations, not the user class which might contain fields that are modified. Since on trace collection there are in total 2 iterations, then I need to restore the fields of the actual user test class. Required functionality was added to the `SnapshotTracker` class, so that it restores the user test class instance between iterations.

eupp:
> I added the testClassInstance parameter to the runConcurrentTest function

Let's please revert this change. The API should be as simple as it could be, that is the point why we adding it. It should only take the lambda to run and that's it. 

> I utilize a specifal wrapper class which calls the lambda and only this wrapper is recreated between iterations, not the user class which might contain fields that are modified.

To make the test work with only the lambda, please remove any state of the test class by either:
- making the respective fields static, or
- promoting the fields to regular variables inside lambda. 

Use one method or another, depending on which makes more sense for a particular test. 

This way we will eliminate any "external" state outside lambda (except static state, but it will be handled by the snapshot tracker).


In general, there is indeed a problem here --- the problem is that the lambda that we pass to the model checker should not capture any state, in order to be deterministically reproducible. 
I will add this problem to the design document, and we will discuss it and possible solutions separately. 

One of possible solutions is similar to what you propose --- to force the user to explicitly mention all captured objects as method parameters. But we need to discuss this solution and how to express it in the API in more detail before implementing it.

eupp:
The task to add support for thread pools turned out to be more challenging than anticipated, and it required non-trivial changes. I transferred it into separate PR so we can review and discuss that code separately: https://github.com/JetBrains/lincheck/pull/447

eupp:
I've tried to add tests with coroutines, using custom thread pool as a dispacther (using `ExecutorService.asCoroutineDispatcher()`: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/as-coroutine-dispatcher.html). 

But it seems I run into bug #404 with this test, so I am going to try debug and fix it first.

eupp (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+internal class GeneralPurposeMCWrapper<R>() {
+    fun run(block: () -> R) = block()
+}
+
+/**
+ * [ExecutionExceptionsVerifier] makes sure that in case if user `check(...)`'s
+ * fail and store their exceptions as an execution result (or just any exception is thrown),
+ * such outcomes will be disallowed and the execution will fail.
+ */
+private class ExecutionExceptionsVerifier(@Suppress("UNUSED_PARAMETER") sequentialSpecification: Class<*>) : Verifier {
```
ndkoval, line 80: Let's rename it `NoExceptionVerifier` and make it an internal class at the top level. We can then eliminate the similar verifier in tests.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+    override fun verifyResults(scenario: ExecutionScenario?, results: ExecutionResult?): Boolean {
+        if (results == null) return true
```
ndkoval, line 82: How can `scenario` and `results` be `null`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+        .verifier(ExecutionExceptionsVerifier::class.java)
+
+    val testCfg = options.createTestConfigurations(GeneralPurposeMCWrapper::class.java)
+
+    withLincheckJavaAgent(testCfg.instrumentationMode) {
+        ensureObjectIsTransformed(block)
+        val strategy = testCfg.createStrategy(GeneralPurposeMCWrapper::class.java, scenario, null, null)
+        val verifier = testCfg.createVerifier()
+
+        for (i in 1..invocations) {
```
ndkoval, line 58: Why do we need to implement this logic again? Can't we just call `options.checkImpl` instead?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * This method will explore different interleavings of the [block] body and all the threads created within it,
+ * searching for the first raised exception.
+ *
+ * @param invocations number of different interleavings of code in the [block] that should be explored.
+ * @param block lambda which body will be a target for the interleavings exploration.
+ */
+fun <R> runConcurrentTest(
+    invocations: Int = DEFAULT_INVOCATIONS_COUNT,
+    block: () -> R
```
ndkoval, line 36: Please add a test on Java, with and without `invocations`,  to check the API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck
+
+import org.jetbrains.kotlinx.lincheck.execution.ExecutionResult
+import org.jetbrains.kotlinx.lincheck.execution.ExecutionScenario
+import org.jetbrains.kotlinx.lincheck.execution.parallelResults
+import org.jetbrains.kotlinx.lincheck.strategy.LincheckFailure
```
ndkoval, line 16: unused imports -- check everywhere, please

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+ * such outcomes will be disallowed and the execution will fail.
+ */
+private class ExecutionExceptionsVerifier(@Suppress("UNUSED_PARAMETER") sequentialSpecification: Class<*>) : Verifier {
+    override fun verifyResults(scenario: ExecutionScenario?, results: ExecutionResult?): Boolean {
+        if (results == null) return true
+        return results.parallelResults[0][0] !is ExceptionResult
+    }
+}
+
+private const val DEFAULT_INVOCATIONS_COUNT = 50_000
```
ndkoval, line 87: please add an empty line at the end

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -174,6 +174,7 @@ abstract class ManagedStrategy(
             useClocks = UseClocks.ALWAYS
         )
 
+
```
ndkoval, line 177: please revert

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1113,7 +1114,7 @@ abstract class ManagedStrategy(
      * Required as a trick to overcome issue with leaking this in constructors, see https://github.com/JetBrains/lincheck/issues/424.
      */
     override fun updateSnapshotBeforeConstructorCall(objs: Array<Any?>) = runInIgnoredSection {
-        staticMemorySnapshot.trackObjects(objs)
+        staticMemorySnapshot.trackObjectsEagerly(objs)
```
ndkoval, line 1117: please revert

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -80,9 +80,9 @@ class SnapshotTracker {
         }
     }
 
-    fun trackObjects(objs: Array<Any?>) {
+    fun trackObjectsEagerly(objs: Array<Any?>) {
```
ndkoval, line 83: please revert

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunWithLambdaRepresentationTests.kt
```diff
@@ -0,0 +1,309 @@
...
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicLongArray
+import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.AtomicReferenceArray
+import kotlin.concurrent.thread
+import kotlin.random.Random
+
+
+abstract class BaseRunWithLambdaRepresentationTest<R>(private val outputFileName: String) {
```
ndkoval, line 35: Please replace `RunWithLabmda` to something more consistent with the API (please remember about the corresponding file names)

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/CustomThreadsRepresentationTest.kt
```diff
@@ -1,219 +0,0 @@
...
-import org.jetbrains.kotlinx.lincheck_test.gpmc.*
-import org.jetbrains.kotlinx.lincheck_test.util.*
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater
-import kotlin.concurrent.thread
-import kotlin.reflect.KClass
-import kotlin.reflect.KFunction
-import org.junit.Test
-import java.util.concurrent.ConcurrentLinkedDeque
-
-class CustomThreadsRepresentationTest {
```
ndkoval, line 31: Should this test be restored?

src/jvm/test/resources/expected_logs/array_rw_run_with_lambda.txt
```diff
@@ -0,0 +1,37 @@
...
+
+---
+The number next to an exception name helps you find its stack trace provided after the interleaving section
+---
+
+The following interleaving leads to the error:
+| -------------------------------------- |
+|                Thread 1                |
+| -------------------------------------- |
+| run(() -> R): IllegalStateException #1 |
```
ndkoval, line 16: Can we not show `run(..)` here?

src/jvm/test/resources/expected_logs/array_rw_run_with_lambda.txt
```diff
@@ -0,0 +1,37 @@
...
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|                                                                                             Thread 1                                                                                              |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| run(() -> R): IllegalStateException #1                                                                                                                                                            |
+|   ArrayReadWriteRunWithLambdaTest#1.block(): threw IllegalStateException at BaseRunWithLambdaRepresentationTest$testRunWithModelChecker$result$1$1.invoke(RunWithLambdaRepresentationTests.kt:44) |
+|     block(): threw IllegalStateException at ArrayReadWriteRunWithLambdaTest.block(RunWithLambdaRepresentationTests.kt:58)                                                                         |
```
ndkoval, line 32: Shouldn't this be the first event in the test?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,87 @@
...
+
+
+/**
+ * This method will explore different interleavings of the [block] body and all the threads created within it,
+ * searching for the first raised exception.
+ *
+ * @param invocations number of different interleavings of code in the [block] that should be explored.
+ * @param block lambda which body will be a target for the interleavings exploration.
+ */
+fun <R> runConcurrentTest(
```
ndkoval, line 34: Let's use Opt-in to mark the functionality as experimental, and clearly indicate in the documentation (with CAPS) that this API is experimental and subject to change. 

https://kotlinlang.org/docs/opt-in-requirements.html

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
eupp: I implemented these changes. 
However, I haven't found `NoExceptionVerifier` or any similar class in tests.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
eupp: Fixed. That was just because of java interop nullability issues.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
eupp: Unfortunately, using this API from Java is currently unsupported due to technical limitations:
https://github.com/JetBrains/lincheck/issues/445

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunWithLambdaRepresentationTests.kt
eupp: Renamed to `RunConcurrentRepresentationTest`. Also, moved the corresponding trace `.txt` files into dedicated sub-folder.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/CustomThreadsRepresentationTest.kt
eupp: It is restored under the name `CustomThreadsRunConcurrentRepresentationTest`.

src/jvm/test/resources/expected_logs/array_rw_run_with_lambda.txt
eupp: I've implemented it, not without some crutches. 
We should really refactor `TraceReporter.kt` at some point, it becomes really painful to maintain and modify it...

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
eupp: Unfortunately, just using `options.checkImpl` currently does not work. 
The problem is that we need to manually trigger the instrumentation of the passed lambda class:
```kotlin
ensureObjectIsTransformed(block)
```

Otherwise, it is not triggered automatically, with the current implementation of `ensureObjectIsTransformed` and `ensureClassHierarchyIsTransformed`. This is because instead of the lambda itself, we pass the `GeneralPurposeModelCheckingWrapper` object to make the whole thing work with the scenario-based Lincheck API. So only this class is transformed. Because this class does not store `block` lambda in any of its fields, but instead is passed as actor argument, its instrumentation is not triggered automatically.

I propose we implement the required refactoring in a separate PR, because it would touch a lot of places in the code (`LinChecker.kt`, runners, managed strategy, `LincheckJavaAgent`, etc).

In a meantime, I changed the code to use `Strategy.runIteration`. It is not as good, but still simplifies the code and avoids unnecessary duplication of logic.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
ndkoval: Yes, so let's ignore the test. However, we need to see what the code looks like in Java.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
ndkoval: Would storing the lambda in `GeneralPurposeModelCheckingWrapper` solve the issue and simplify the logic?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
ndkoval: Sorry, my bad, I confused it with the verifier that always fails.

src/jvm/test/resources/expected_logs/run_concurrent_test/array_rw.txt
```diff
@@ -0,0 +1,35 @@
...
+
+---
+The number next to an exception name helps you find its stack trace provided after the interleaving section
+---
+
+The following interleaving leads to the error:
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|                                                                                                    Thread 1                                                                                                     |
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|   ArrayReadWriteRunConcurrentRepresentationTest#1.block(): threw IllegalStateException at BaseRunConcurrentRepresentationTest$testRunWithModelChecker$result$1$1.invoke(RunConcurrentRepresentationTests.kt:35) |
```
ndkoval, line 16: Please fix the indentation here, fewer whitespaces should be in the beginning

src/jvm/test/resources/expected_logs/run_concurrent_test/array_rw.txt
```diff
@@ -0,0 +1,35 @@
+= Invalid execution results =
+| -------------------------------------- |
+|                Thread 1                |
+| -------------------------------------- |
+| run(() -> R): IllegalStateException #1 |
```
ndkoval, line 5: In case of `runConcurrentTest`, we do not need to show the scenario. Instead, we can show the exception followed by the short and detailed interleavings.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -0,0 +1,74 @@
...
+
+import org.jetbrains.kotlinx.lincheck.execution.*
+import org.jetbrains.kotlinx.lincheck.strategy.runIteration
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingOptions
+import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.ensureObjectIsTransformed
+import org.jetbrains.kotlinx.lincheck.transformation.withLincheckJavaAgent
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+
+
+@RequiresOptIn(message = "The model checking API is experimental and could change in the future.")
```
ndkoval, line 21: will change :)

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
ndkoval: Note: as we are not working on the API now, we can postpone this question.

src/jvm/test/resources/expected_logs/run_concurrent_test/array_rw.txt
eupp: Done, please have a look at reports

src/jvm/test/resources/expected_logs/run_concurrent_test/kotlin_thread.txt
```diff
@@ -0,0 +1,42 @@
+= Concurrent test failed =
```
ndkoval, line 1: Do we need this line, or should we start the output with the exception?

eupp: I would leave it. 

1) It is a title of the whole report, explaining what happened. Starting with the exception might be confusing, as if the code itself thrown an exception.

2) Currently there are two possible lines:

```
= Concurrent test failed =
```


```
= Concurrent test has hung =
```

ndkoval (approved):

Issue #405, opened 2024-09-30
`IndexOutOfBoundsException` in the modelChecking test


de-shyt:
The complete output with the detailed stack trace is in the file [output.log](https://github.com/user-attachments/files/17195620/output.log)

```
Exception stacktrace:
java.lang.IllegalStateException: Trying to switch the execution to thread 2,
but only the following threads are eligible to switch: [0]
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.chooseThread(ModelCheckingStrategy.kt:306)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.doSwitchCurrentThread(ManagedStrategy.kt:524)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread(ManagedStrategy.kt:500)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread$default(ManagedStrategy.kt:493)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPoint(ManagedStrategy.kt:402)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPointOnAtomicMethodCall(ManagedStrategy.kt:964)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.beforeMethodCall(ManagedStrategy.kt:903)
	at sun.nio.ch.lincheck.Injections.beforeMethodCall(Injections.java:249)
	at kotlinx.coroutines.CancellableContinuationImpl.tryResume(CancellableContinuationImpl.kt:279)
	at kotlinx.coroutines.CancellableContinuationImpl.dispatchResume(CancellableContinuationImpl.kt:468)
	at kotlinx.coroutines.CancellableContinuationImpl.completeResume(CancellableContinuationImpl.kt:591)
	at com.deshyt.buffered.BufferedChannel.tryResumeRequest(BufferedChannel.kt:249)
	at com.deshyt.buffered.BufferedChannel.tryResumeRequest(BufferedChannel.kt:238)
	at com.deshyt.buffered.BufferedChannel.updateCellOnSend(BufferedChannel.kt:106)
	at com.deshyt.buffered.BufferedChannel.send(BufferedChannel.kt:74)
[...]

Issue #402, opened 2024-09-27
`UnsupportedOperationException` after adding a custom scenario


de-shyt:
The exception occurs when a custom scenario is added: 

```
java.lang.UnsupportedOperationException: can't get field offset on a record class: private final java.lang.reflect.Parameter[] java.lang.reflect.Executable$ParameterData.parameters
	at jdk.unsupported/sun.misc.Unsafe.objectFieldOffset(Unsafe.java:655)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.readField(ObjectTraverser.kt:201)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:74)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:63)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:63)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:105)
	at org.jetbrains.kotlinx.lincheck.ObjectTraverserKt.enumerateObjects(ObjectTraverser.kt:42)
	at org.jetbrains.kotlinx.lincheck.IdeaPluginKt.createObjectToNumberMapAsArray(IdeaPlugin.kt:147)
	at org.jetbrains.kotlinx.lincheck.IdeaPluginKt.visualize(IdeaPlugin.kt:129)
[...]

PR #414, opened 2024-10-09
Creating a thread actor with `blocking = true`

de-shyt:


ndkoval:
@de-shyt do you still need this functionality?

de-shyt:
@ndkoval the `blockingActor` method is used in several issues (for example, [this](https://github.com/JetBrains/lincheck/issues/405) and [this](https://github.com/JetBrains/lincheck/issues/402))

ndkoval:
Since the issues mentioned above are already closed, we do not have applications for the suggested API, and we are going to rewrite it, I'm closing the PR.

PR #413, opened 2024-10-07
Non-determinism fix: ClassLoaders are now transformed in the Model Checking mode

avpotapov00:
closes #412, #377, #405

Issue #412, opened 2024-10-07
ClassLoaders are not transformed in ModelChecking mode


avpotapov00:
ClassLoaders are not transformed in ModelChecking mode. That leads to non-determinism and potential deadlocks.

PR #415, opened 2024-10-11
Support custom threads

eupp:


eupp:
The only failing CI configuration is "Integration Test with kotlinx.coroutines". 
I believe it is because of #412, and once we merge #413 the problem should be fixed. 
I've create a separate test branch where I merged the #413 into my branch, and it seems that there all kotlinx.coroutines [tests are passed](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_IntegrationTestWithKotlinxCoroutinesMasterLinuxOnJava17/4838815?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true).

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/transformation/KotlinStdlibTransformationTest.kt
```diff
@@ -54,5 +54,6 @@ class KotlinStdlibTransformationTest : AbstractLincheckTest() {
 
     override fun <O : Options<O, *>> O.customize() {
         iterations(1)
+        invocationTimeout(20_000)
```
ndkoval, line 57: Why do you need this change?

eupp: That was a performance "bug" actually, already fixed it and reverted the change.

eupp:
@ndkoval while working on this I discovered a few problems, which are probably should be addressed in separate PRs.

One of the problems is related to the local objects tracking --- the current implementation does not work correctly with custom threads (see example  below and the comment in `DataStructuresTests::incorrectHashMap` test).
So the problem can be illustrated by the following example:

```kotlin
class Box(var x: Int)

fun test(): Int {
    val box = Box() // <- this object is incorrectly classified as a local object
    thread {
        box.x = 42 // the local object tracker does not detect here that the `box` object,
                   // stored in the local variable, escapes into another thread;
                   // thus it will not insert a switch point before accesses to this object fields
    }
    return box.x
}
```

I would propose that we can address this problem separately in another PR after we merge this one.
The reasons is that adding support of this case would require significant refactoring of local objects tracking algorithm,
and this PR is already big.

Alternatively, we can first perform the necessary refactoring of the local objects tracking algorithm in a separate PR,
and after this rebase and merge custom threads PR.

eupp:
Another small bug fix on which this PR relies on: #426

gradle.properties
```diff
@@ -15,7 +15,7 @@ version=2.35-SNAPSHOT
 inceptionYear=2019
 lastCopyrightYear=2023
 
-jdkToolchainVersion=17
+jdkToolchainVersion=8
```
ndkoval, line 18: please revert the change

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -45,6 +50,101 @@ public class Injections {
     @SuppressWarnings({"FieldCanBeLocal", "unused"})
     private static int currentEventId = -1;
 
+    // Thread local variable storing testing code and ignored section flags.
```
ndkoval, line 53: Please remove this line

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -45,6 +50,101 @@ public class Injections {
...
+     * However, because there is no `ConcurrentWeakIdentityHashMap` out of the box in
+     * the java standard library, we use a few tricks to achieve a similar effect.
+     *
+     * - We use `ConcurrentHashMap` to guarantee thread safety.
+     * - We use identity hash codes of the thread objects as keys.
+     * - To accommodate for potential hash code collisions, we store lists of thread descriptors as values.
+     * - These lists use copy-on-write strategy when a new descriptor is added to avoid race conditions.
+     * - Thread descriptors store weak references to thread objects, and thus do not prevent their garbage collection.
+     *
+     * TODO: although the garbage collection of thread objects is not prevented thanks to weak references,
```
ndkoval, line 78: You can do the cleanup on each 1000's (feel free choose any number here) access

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -45,6 +50,101 @@ public class Injections {
...
+        }
+        threadDescriptor.set(descriptor);
+    }
+
+    public static ThreadDescriptor getThreadDescriptor(Thread thread) {
+        if (thread instanceof TestThread) {
+            return ((TestThread) thread).descriptor;
+        }
+        int hashCode = System.identityHashCode(thread);
+        ArrayList<ThreadDescriptor> threadDescriptors = threadDescriptorsMap.get(hashCode);
```
ndkoval, line 110: Can we use a weak hash map instead?

src/jvm/test/resources/expected_logs/obstruction_freedom_synchronized.txt
```diff
@@ -10,19 +10,21 @@ The following interleaving leads to the error:
...
 | operation()                                                                                                              |                                                                                                                          |
-|   MONITORENTER at ObstructionFreedomSynchronizedRepresentationTest.operation(ObstructionFreedomRepresentationTest.kt:70) |                                                                                                                          |
+|   MONITORENTER at ObstructionFreedomSynchronizedRepresentationTest.operation(ObstructionFreedomRepresentationTest.kt:61) |                                                                                                                          |
+|   switch (reason: lock is already acquired)                                                                              |                                                                                                                          |
```
ndkoval, line 17: is there any switch? It seems the execution ends after this event.

src/jvm/test/resources/expected_logs/validation_function_failure.txt
```diff
@@ -19,8 +19,9 @@ All operations above the horizontal line | ----- | happen before those below the
 java.lang.IllegalStateException: Validation works!
 	at org.jetbrains.kotlinx.lincheck_test.representation.ValidationFunctionCallTest.validateWithError(ValidationFunctionTests.kt:44)
```
ndkoval, line 20: Should we shrink the stacktrace here? If so, let's make an issue for that.

src/jvm/test/resources/expected_logs/custom_threads_deque_trace.txt
```diff
@@ -0,0 +1,78 @@
...
+| incorrectConcurrentLinkedDeque(): IllegalStateException #1                                                                                     |                                                                                                                                                             |          |
+|   ConcurrentLinkedDeque#1.addLast(1) at CustomThreadsRepresentationTest.incorrectConcurrentLinkedDeque(CustomThreadsRepresentationTest.kt:112) |                                                                                                                                                             |          |
+|   THREAD_START(2)                                                                                                                              |                                                                                                                                                             |          |
```
ndkoval, line 18: Consider changing it to `start Thread 2` or similar

src/jvm/test/resources/expected_logs/custom_threads_deque_trace.txt
```diff
@@ -0,0 +1,78 @@
...
+|   switch                                                                                                                                       |                                                                                                                                                             |          |
+|                                                                                                                                                | run()                                                                                                                                                       |          |
+|                                                                                                                                                |   switch                                                                                                                                                    |          |
```
ndkoval, line 22: is it ok that no event has occurred here?

src/jvm/test/resources/expected_logs/custom_threads_deque_trace.txt
```diff
@@ -0,0 +1,78 @@
...
+|   switch (reason: waiting for thread to join)                                                                                                  |                                                                                                                                                             |          |
+|                                                                                                                                                |                                                                                                                                                             | run()    |
+|   switch (reason: waiting for thread to join)                                                                                                  |                                                                                                                                                             |          |
```
ndkoval, line 25: Here, the scheduler incorrectly switched the execution to a thread that cannot continue at the moment

src/jvm/test/resources/expected_logs/custom_threads_deque_trace.txt
```diff
@@ -0,0 +1,78 @@
...
+|                                                                                                                                                | run()                                                                                                                                                       |          |
+|                                                                                                                                                |   switch                                                                                                                                                    |          |
+|   switch (reason: waiting for thread to join)                                                                                                  |                                                                                                                                                             |          |
```
ndkoval, line 23: let's mention which thread the execution is calling `join` on

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/Utils.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.jetbrains.kotlinx.lincheck.strategy.*
+import org.jetbrains.kotlinx.lincheck.transformation.*
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import kotlin.reflect.*
+import org.junit.Assert
+
+internal fun modelCheckerTest(
+    testClass: KClass<*>,
+    testOperation: KFunction<*>,
+    outcomes: Set<Any?> = setOf(),
```
ndkoval, line 25: Are these the excepted outcomes?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -79,7 +79,8 @@ fun shouldReplayInterleaving(): Boolean {
  */
 @Suppress("UNUSED_PARAMETER")
 fun beforeEvent(eventId: Int, type: String) {
-    val strategy = (Thread.currentThread() as? TestThread)?.eventTracker ?: return
+    val strategy = Injections.getCurrentThreadDescriptor()?.eventTracker
```
ndkoval, line 82: Have you checked how custom threads work with the plugin?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Utils.kt
```diff
@@ -0,0 +1,47 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2024 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
```
ndkoval, line 11: Do we need to have to `Utils.kt` files?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -204,7 +204,7 @@ internal class LoopDetector(
         // Has the thread changed? Reset the counters in this case.
         check(lastExecutedThread == iThread) { "reset expected!" }
         // Ignore coroutine suspension code locations.
-        if (codeLocation == COROUTINE_SUSPENSION_CODE_LOCATION) return Decision.Idle
```
ndkoval, line 207: Why UNKNOWN? The documentation here and of the `UNKNOWN_CODE_LOCATION` field says it is the coroutine suspension code location.

ndkoval (changes requested):

eupp:
From what I can see, there is no significant changes in the time of CI builds between this branch and `develop`:
-`dynamic-threads`  [branch](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildAggregated/4917853?buildTab=dependencies&type=snapshot&mode=list)
- `develop` [branch](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildAggregated/4917203?buildTab=dependencies&type=snapshot&mode=list)

The reason is that in this PR I strive to preserve the old behavior whenever possible, by adding special treatment of `TestThread` class, see an example here:
https://github.com/JetBrains/lincheck/blob/3feb8a93a90644c85c76409bacd72dea01ee7da1/bootstrap/src/sun/nio/ch/lincheck/Injections.java#L91

Thus, all the code interacting with `TestThread`-s (i.e. threads create by Lincheck) should work almost the same as before.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
eupp: I am afraid no, because we need concurrent identity hash map with weak keys, and there is no such data structure out-of-the-box in the Java standard library. 

Please also have a look at the comment before `threadDescriptorsMap` field declaration --- it explains the reason and the proposed solution.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
eupp: It was `COROUTINE_SUSPENSION_CODE_LOCATION` previously, I renamed it into `UNKNOWN_CODE_LOCATION` because I re-used it for in a different context, but with a similar purpose (that is to represent an unknown code location). 

I fixed the comments in the code so they now mention `UNKNOWN_CODE_LOCATION`.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/Utils.kt
eupp: Renamed to `expectedOutcomes`.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
ndkoval: Would using `Collections.synchronizedMap(...)` solve the issue?

src/jvm/test/resources/expected_logs/custom_threads_deque_trace.txt
eupp: Done

src/jvm/test/resources/expected_logs/validation_function_failure.txt
eupp: What do you mean by "shrinking" the stack trace?
Do you mean that we need to remove the lines like this one (like we do for exceptions thrown in actors?):
```
at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.testThreadRunnable$lambda$10(FixedActiveThreadsExecutor.kt:172)
```

If so, then yes, there is a problem currently that validation function stack traces are handled differently compared to other stack traces printed by Lincheck.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Utils.kt
eupp: The idea is to eventually leave only one file, inside the `util` subpackage, and move the stuff from the top-level package `Utils.kt` file either into `util` subpackage or to other more suitable places.

However, I didn't want to clutter this PR with unnecessary changes, thus I haven't modified the top-level `Utils.kt` file.

src/jvm/test/resources/expected_logs/obstruction_freedom_synchronized.txt
eupp: Fixed

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
eupp: It doesn't. The code now looks like this:
```
        // Ignore unknown code locations.
        if (codeLocation == UNKNOWN_CODE_LOCATION) return Decision.Idle
```

bootstrap/src/sun/nio/ch/lincheck/Injections.java
eupp: `Collections.synchronizedMap(...)` would not fully solve the problem. It would make the collection thread-safe, but `WeakIdentityHashMap` is still not a thing in Java stdlib (remember we need thread-safe weak identity hash-map).

Also, there is a method to lookup for a descriptor of the given thread:
```
public static ThreadDescriptor getThreadDescriptor(Thread thread)
``` 

which uses `threadDescriptorsMap.get`. 

Forcing the descriptor lookup function to acquire a lock on each lookup (instead of using concurrent hash map), may in theory have a performance penalty.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
eupp: No it does not currently work. 

I found one problem in the plugin code that I've managed to fix by myself, but it still does not work. 
There are some other bugs, and I need help from someone on the plugin side to debug the remaining problems.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
eupp: Fixed, changed to `WeakIdentityHashMap` + `Collections.synchronizedMap(...)`

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
eupp: The plugin seems to work now, but it required a few commits to the plugin repo. We would need to sync the releases of the new version of lincheck and the plugin.

ndkoval (changes requested):

src/jvm/test/resources/expected_logs/validation_function_failure.txt
eupp: Here is the issue: https://github.com/JetBrains/lincheck/issues/438

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -328,12 +325,18 @@ internal open class ParallelThreadsRunner(
...
         } catch (e: TimeoutException) {
-            val threadDump = collectThreadDump(this)
-            return RunnerTimeoutInvocationResult(threadDump, collectExecutionResults())
+            return RunnerTimeoutInvocationResult()
         } catch (e: ExecutionException) {
+            // in case if invocation thrown an exception,
+            // make sure all threads terminate before continuing
+            // and handle possible hangings gracefully
+            runCatching { strategy.awaitAllThreads(timeout) }.onFailure { exception ->
+                if (exception is TimeoutException) return RunnerTimeoutInvocationResult()
```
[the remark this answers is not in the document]
eupp: Nothing should happen. This is a case when scenario execution raised some exception. 
In this case, we just want to wait for all threads to terminate. 

However, the threads may hung. In this case, `awaitAllThreads` should throw `TimeoutException`.
So I wrap `awaitAllThreads` call into `runCatching` to handle this timeout exception and return `RunnerTimeoutInvocationResult`. 

Otherwise, if threads terminate normally without hunging, the `UnexpectedExceptionInvocationResult` will be returned with the original exception thrown by the scenario.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Utils.kt
eupp: Renamed to `Ensure.kt`

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -447,7 +479,7 @@ internal open class ParallelThreadsRunner(
 
     override fun isCurrentRunnerThread(thread: Thread): Boolean = executor.threads.any { it === thread }
 
-    override fun onFinish(iThread: Int) {}
+    override fun onThreadFinish(iThread: Int) {}
 
     override fun onFailure(iThread: Int, e: Throwable) {}
```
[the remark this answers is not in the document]
eupp: Renamed

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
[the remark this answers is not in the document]
ndkoval: Thanks for the clarification! Could you please add a short version of this comment to the code?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedThreadScheduler.kt
```diff
@@ -0,0 +1,123 @@
...
+ * [ManagedThreadScheduler] is responsible for managing the execution of threads
+ * ensuring that at any given moment, only a single registered thread is allowed to run.
+ *
+ * To achieve this goal, the scheduler relies on cooperation from the [ManagedStrategy],
+ * in particular, that the strategy correctly places [awaitTurn] calls
+ * to artificially block a thread from running until it is scheduled.
+ *
+ * @see [ThreadScheduler]
+ */
+class ManagedThreadScheduler : ThreadScheduler() {
```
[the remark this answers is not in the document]
eupp: Let's keep them separate. 

Later on, I also want to move the coroutine managing logic from `ParallelThreadsRunner` to `ThreadScheduler` to keep the state of threads in a single place. In this case, `ThreadScheduler` will be used by both `StressStrategy` and `ManagedStrategy`.

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
```diff
@@ -0,0 +1,84 @@
...
+---
+The number next to an exception name helps you find its stack trace provided after the interleaving section
+---
+
+The following interleaving leads to the error:
+| ----------------------------------------------------------------------------------------------------------------------------------- |
+|                                              Thread 1                                              | Thread 2 | Thread 3 | Thread 4 |
+| ----------------------------------------------------------------------------------------------------------------------------------- |
+| basic(): IllegalStateException #1                                                                  |          |          |          |
+|   Thread#1.start() at CustomThreadsRepresentationTest.basic(CustomThreadsRepresentationTest.kt:47) |          |          |          |
```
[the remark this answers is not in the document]
eupp: In `Thread#1` the `#1` is the object numeration, while in `Thread 1`, `1` is the thread numeration. 
The two numerations can differ, unfortunately. 

It is not clear what to do about this. 
Even if we use thread numeration in `Thread#N` string, what to do then with the thread objects not registered in managed strategy (for these objects the thread number would not exist)?
Moreover, threads are registered and get their thread number only after `start()` call, so what should be the number before the thread is started?

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
```diff
@@ -0,0 +1,84 @@
...
+|   Thread#1.join() at CustomThreadsRepresentationTest.basic(CustomThreadsRepresentationTest.kt:48)  |                                                                                                                       |                                                                                                                       |                                                                                                                       |
+|     join Thread 2                                                                                  |                                                                                                                       |                                                                                                                       |                                                                                                                       |
```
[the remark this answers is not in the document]
eupp: Do you mean `start Thread N` and `join Thread N` lines?

They correspond to `ThreadStartTracePoint` and `ThreadJoinTracePoint` objects.
If I don't add these objects to the trace, then no stack frames related to start/join are going to be printed at all (e.g. `Thread#2.start()` or `Thread#2.join()` are going to be printed). 
This is because currently, method calls are only added in the trace if there are some trace points inside.

src/jvm/test/resources/expected_logs/custom_threads_kotlin_trace_jdk8.txt
```diff
@@ -0,0 +1,52 @@
...
+---
+The number next to an exception name helps you find its stack trace provided after the interleaving section
+---
+
+The following interleaving leads to the error:
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|                                                                                         Thread 1                                                                                         | Thread 2 |
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| kotlinThread(): IllegalStateException #1                                                                                                                                                 |          |
+|   ThreadsKt.thread$default(false,false,null,null,0,kotlinThread$t$1#1,31,null): thread$thread$1#1 at CustomThreadsRepresentationTest.kotlinThread(CustomThreadsRepresentationTest.kt:61) |          |
```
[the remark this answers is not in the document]
eupp: In `kotlinThread$t$1#1`
- `$1` is some lambda-related identifier generated by the Kotlin compiler, it is a part of the class name
- `#1` is the object numeration

And in `Thread 2`, the number `2` is the thread numeration, separate from the object numeration.

src/jvm/test/resources/expected_logs/custom_threads_hashmap_trace.txt
```diff
@@ -0,0 +1,107 @@
...
+|                                                                                                                                                                                                                         |          |       ObjectRef#2.element.WRITE(null) at CustomThreadsRepresentationTest$incorrectHashMap$t2$1.invoke(CustomThreadsRepresentationTest.kt:142) |
+|                                                                                                                                                                                                                         |          |   result: void                                                                                                                                |
+|   thread$1.join() at CustomThreadsRepresentationTest.incorrectHashMap(CustomThreadsRepresentationTest.kt:144)                                                                                                           |          |                                                                                                                                               |
```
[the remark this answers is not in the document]
eupp: In `thread$1`, `$1` is some lambda-related identifier generated by the Kotlin compiler, it is a part of the class name.

src/jvm/test/resources/expected_logs/custom_threads_kotlin_trace_jdk8.txt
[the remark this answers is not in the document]
eupp: What exactly should we filter?
This is just how Kotlin's function call with the default parameters look like in the trace. 
Any other similar Kotlin function (with default arguments) would have similar representation in the trace.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
```diff
@@ -307,6 +302,8 @@ internal open class ParallelThreadsRunner(
             // Execute the parallel part.
             beforePart(PARALLEL)
             timeout -= executor.submitAndAwait(parallelPartExecutions, timeout)
```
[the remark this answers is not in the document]
eupp: Yes. 

- `executor.submitAndAwait(parallelPartExecutions, timeout)` submits the tasks and awaits for their compeltion. 
- `strategy.awaitAllThreads(timeout)` awaits for the completion of all threads registered in the strategy (including user custom threads).

The only reason why `strategy.awaitAllThreads(timeout)` is needed now is because now we also have custom threads, so we also need to wait for their completion. 

Previously, number of threads was equal to the number of submitted tasks to the executor, so no additional waiting was required after `executor.submitAndAwait(parallelPartExecutions, timeout)`.

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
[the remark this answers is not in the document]
ndkoval: What I find important is the thread objects enumeration to match the one in the thread columns

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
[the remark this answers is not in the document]
ndkoval: Is it possible to process these method calls similarly to the atomic ones?

src/jvm/test/resources/expected_logs/custom_threads_kotlin_trace_jdk8.txt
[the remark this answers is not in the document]
ndkoval: Ok.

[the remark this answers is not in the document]
ndkoval: Let's make the thread numeration consistent with the thread object numeration

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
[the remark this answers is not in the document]
ndkoval: Can we keep `strategy.awaitAllThreads(timeout)` and remove `executor.submitAndAwait(parallelPartExecutions, timeout)`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedThreadScheduler.kt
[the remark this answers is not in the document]
ndkoval: Got it

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -68,6 +69,17 @@ abstract class Strategy protected constructor(
...
 
+    /**
+     * Waits for all threads in the current invocation to complete or suspend within the given timeout.
+     *
+     * @param timeoutNano The maximum time to wait in nanoseconds.
+     * @return The elapsed time in nanoseconds if all threads finish within the timeout.
+     * @throws TimeoutException if more than [timeoutNano] is passed.
+     */
+    open fun awaitAllThreads(timeoutNano: Long): Long {
+        return 0L
```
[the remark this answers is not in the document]
ndkoval: > currently, creation of custom threads is not intercepted in the StressStrategy.

Should it be intercepted?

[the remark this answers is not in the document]
ndkoval: Also, it is unexpected that `StressStrategy. awaitAllThreads` does nothing. Can we do something about it?

bootstrap/src/sun/nio/ch/lincheck/ThreadRegistry.java
```diff
@@ -0,0 +1,137 @@
...
+ * Moreover, the registry also maintains a global map
+ * to store a mapping from each registered thread to its descriptor
+ * in a thread-safe and garbage collection-friendly manner.
+ * The thread-local variable is used for quick retrieval of the descriptor from within the thread itself,
+ * while the global map is used to access a descriptor of the thread from another thread.
+ *
+ * <p>
+ * This class provides only static members and is not meant to be instantiated.
+ */
+public final class ThreadRegistry {
```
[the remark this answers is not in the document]
eupp: Moved

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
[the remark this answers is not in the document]
eupp: Ok, I can try the following:
- in `Thread#N` use the thread number if the thread is registered
- if the thread is not registered, use some other number (perhaps based on object numeration?), which should not collide with the numbers of registered threads

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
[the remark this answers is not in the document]
eupp: I can try this approach.

src/jvm/test/resources/expected_logs/custom_threads_kotlin_trace_jdk8.txt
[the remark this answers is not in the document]
eupp: Ok, I can try to fix `#1` numeration, but with `$1` I am afraid we cannot do much. 
The `$1` is not a thread number, not an object number, it is just number of lambda class generated by the Kotlin compiler.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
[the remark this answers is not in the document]
eupp: > Should it be intercepted or should we keep the logic as is?

This is a complicated question. Perhaps, in the future we should revisit what events are intercepted in the stress strategy in general. I remember there was also a need to intercept `park` events to support blocking operations. 
For now, I propose to not modify the `StressStrategy`. 

> Also, it is unexpected that StressStrategy. awaitAllThreads does nothing. Can we do something about it?

I can rename `awaitAllThreads` to `awaitUserThreads` (or `awaitCustomThreads`), and add a comment that stress strategy does not track user threads currently, to make the situation less confusing.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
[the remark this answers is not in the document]
eupp: We cannot remove `executor.submitAndAwait(parallelPartExecutions, timeout)` --- its both submits and awaits the tasks. We need to at least submit the tasks : )

I can rename `strategy.awaitAllThreads(timeout)` to `strategy.awaitUserThreads(timeout)` to make the code less confusing.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
```diff
@@ -0,0 +1,392 @@
...
+     *   -1 if the timeout is reached.
+     */
+    fun awaitThreadFinish(threadId: ThreadId, timeoutNano: Long): Long {
+        val threadData = threads[threadId]!!
+        // special handling of Lincheck test threads
+        if (threadData.descriptor.thread is TestThread) {
+            val elapsedTime = threadData.spinner.spinWaitTimedUntil(timeoutNano) {
+                threadData.state == ThreadState.FINISHED ||
+                // TODO: due to limitations of current implementation,
+                //   Lincheck test threads sometime end up in ABORTED state,
```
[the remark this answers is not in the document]
eupp: If I remember correctly, `onThreadFinish` is not guaranteed to be called at the end of the thread execution. 
This can happen in case of timeout, or in case if some exception was thrown from the code and not properly handled. 

To fix this, we will need to modify `TestThreadExecutionGenerator` class, to modify the generated bytecode. Wrap the code in try-catch, and put `onThreadFinish` into `finally` block to guarantee it is always called at the end.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ThreadTransformer.kt
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private fun isThreadStartMethod(methodName: String, desc: String): Boolean =
+        methodName == "start" && desc == VOID_METHOD_DESCRIPTOR && isThreadSubClass(className)
+
+    private fun isThreadRunMethod(methodName: String, desc: String): Boolean =
+        methodName == "run" && desc == VOID_METHOD_DESCRIPTOR && isThreadSubClass(className)
+
+    // TODO: add support for thread joins with time limit
```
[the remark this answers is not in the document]
eupp: Done

src/jvm/test/resources/expected_logs/custom_threads_trace.txt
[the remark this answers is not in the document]
eupp: The approach with atomic methods did not work out. Because if we treat start/join as atomic methods, the managed strategy also starts to insert switch points before them (and previously we agreed to get rid of exactly this kind of redundant switch points before start/join). 

However, I've implemented alternative approach relatively easy. 
I still add `ThreadStartTracePoint` and `ThreadJoinTracePoint` to the trace, but then I'm simply do not print them.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ParallelThreadsRunner.kt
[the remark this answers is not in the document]
eupp: Renamed `awaitAllThreads` to `awaitUserThreads`.

ndkoval (approved):

PR #416, opened 2024-10-16
Support Java 8

zhelenskiy:


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1190,7 +1190,7 @@ abstract class ManagedStrategy(
 
     private fun initializeVarHandleMethodCallTracePoint(
         tracePoint: MethodCallTracePoint,
-        varHandle: VarHandle,
+        varHandle: Any?, // for Java 8 VarHandle class does not exist
```
ndkoval, line 1193: Is it possible to make it `Any` (without `?`)? The same question applies to other `VarHandle -> Any?` type changes.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/VarHandleNames.kt
```diff
@@ -20,7 +20,7 @@ import java.lang.reflect.Field
 /**
  * Helper object to provide the field name and the owner of the VarHandle method call.
  */
-@Suppress("SameParameterValue")
+@Suppress("SameParameterValue", "Since15")
```
ndkoval, line 23: Why is `Since15` necessary here?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
```diff
@@ -123,7 +123,7 @@ internal fun isAtomicFieldUpdaterMethod(className: String, methodName: String) =
     isAtomicFieldUpdaterClass(className) && methodName in atomicFieldUpdaterMethods
 
 internal fun isVarHandle(obj: Any?) =
-    obj is VarHandle
+    obj != null && generateSequence(obj::class.java) { it.superclass }.any { isVarHandleClass(it.canonicalName ?: "") }
```
ndkoval, line 126: Could you please add a text explanation here? The code is non-trivial to read.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Spinner.kt
```diff
@@ -47,7 +47,7 @@ internal class Spinner(val nThreads: Int = -1) {
         var counter = 0
         val yieldLimit = 1 + if (shouldSpin) SPIN_CYCLES_BOUND else 0
         while (!condition()) {
-            Thread.onSpinWait()
+            threadOnSpinWait?.invoke(null)
```
ndkoval, line 50: Invoking `Thread.onSpinWait` via reflection is not a good idea. Let's check how the performance changes with `Thread.onSpinWait` being omitted.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/VarHandleRepresentationTests.kt
```diff
@@ -8,6 +8,8 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+@file:Suppress("Since15")
```
ndkoval, line 11: Is this suppression necessary?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/transformation/LocalObjectsTests.kt
```diff
@@ -243,6 +243,7 @@ class UnsafeLocalObjectsTest {
  * If we hadn't such check, this test would hang due to infinite spin-loop on a local object operations with
  * no chances to detect a cycle and switch.
  */
+@Suppress("Since15")
```
ndkoval, line 246: Is this suppression necessary?

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt
```diff
@@ -213,7 +213,12 @@ internal class FixedActiveThreadsExecutor(private val testName: String, private
 
 }
 
-private val majorJavaVersion = Runtime.version().version()[0]
+private val majorJavaVersion = runCatching {
```
ndkoval, line 216: Would `System.getProperty("java.specification.version")` work as a replacement?

ndkoval (changes requested):
@zhelenskiy, thanks for the change! I've left several comments, could you please take a look?

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt
zhelenskiy: Do you mean providing it explicitly as an argument?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/VarHandleNames.kt
zhelenskiy: To suppress Idea warnings about usage of `since JDK 9` API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Spinner.kt
zhelenskiy: I'll try method handles.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/VarHandleRepresentationTests.kt
zhelenskiy: https://github.com/JetBrains/lincheck/pull/416#discussion_r1803876177

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/transformation/LocalObjectsTests.kt
zhelenskiy: https://github.com/JetBrains/lincheck/pull/416#discussion_r1803876177

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/VarHandleNames.kt
ndkoval: How can `since JDK 9 API` usages be in the code with the source level being Java 8?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Spinner.kt
ndkoval: I don't have ones in Java 8. Just check what changes without `Thread.onSpinWait`

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/FixedActiveThreadsExecutor.kt
ndkoval: As far as I understand, it should be provided automatically by the JVM.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
eupp: By the way, I think there is a finite list of class inheriting from `VarHandle` (?), so instead of traversing the class hierarchy, we can just check against a set of class name strings. 

You can have a look at these class names here:
https://github.com/JetBrains/lincheck/blob/800f590ef453c3eab5c87b7264c29e580060cec4/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/VarHandleNames.kt#L29

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Spinner.kt
zhelenskiy: [It is even in Java 7](https://docs.oracle.com/javase/7/docs/api/java/lang/invoke/MethodHandle.html)

But I think we can just remove the method call. Is it OK?

ndkoval:
@zhelenskiy Please also run the tests on the following build configurations:
[Build on Java 8](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava8?branch=develop&buildTypeTab=overview&mode=builds)
[Build on Java 9](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava9?branch=develop&buildTypeTab=overview&mode=builds)

zhelenskiy:
Manually run build for 8 [passes](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava8/4883819?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) the tests.
The one for Java 9 [does not](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava9/4883820?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true&expandBuildDeploymentsSection=false). But the only failed test is a representation test where the difference is in line numbers in the exception message. Unfortunately, I am not able to fix it because I cannot test it locally. The reason for that is that JDK is not supported for almost 7 years, and I cannot install it on my machine. Furthermore, I don't even see the reason to support it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
```diff
@@ -122,8 +123,92 @@ internal fun isAtomicFieldUpdaterClass(className: String) =
...
+    "java.lang.invoke.VarHandleShorts${'$'}Array",
+    "java.lang.invoke.VarHandleShorts${'$'}FieldInstanceReadOnly",
+    "java.lang.invoke.VarHandleShorts${'$'}FieldInstanceReadWrite",
+    "java.lang.invoke.VarHandleShorts${'$'}FieldStaticReadOnly",
+    "java.lang.invoke.VarHandleShorts${'$'}FieldStaticReadWrite",
+)
+
+@OptIn(ExperimentalContracts::class)
+internal fun isVarHandle(obj: Any?): Boolean {
+    contract { returns(true) implies (obj != null) }
```
ndkoval, line 209: Can we try something similar to the code below? 
```
    VAR_HANDLE_CLASS?.isAssignableFrom(obj)
```

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AtomicMethods.kt
eupp: BTW, if the concern is the matching against a hard-coded set of strings of var-handle class names, then we already have a similar code here:
https://github.com/JetBrains/lincheck/blob/800f590ef453c3eab5c87b7264c29e580060cec4/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/VarHandleNames.kt#L26

zhelenskiy: I also agree + I benchmarked and found that it is slower than mine.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -344,12 +344,17 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
...
     override fun transform(
         loader: ClassLoader?,
-        internalClassName: String,
+        internalClassName: String?,
         classBeingRedefined: Class<*>?,
         protectionDomain: ProtectionDomain?,
         classBytes: ByteArray
     ): ByteArray? = runInIgnoredSection {
         // If the class should not be transformed, return immediately.
+        if (classBeingRedefined == null) {
```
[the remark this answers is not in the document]
ndkoval: Could you please explain what `if (classBeingRedefined == null)` means?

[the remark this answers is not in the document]
zhelenskiy: That the class, being instrumented is null.
Here is the source of the failure. It happens because instrumentation does not a class.
image
image

Issue #389, opened 2024-09-25
Improve support of static initialization blocks and statically-reachable memory


eupp:
Currently, if a Lincheck test execution both:
- reads some values from statically-reachable memory, and
- modifies the content of the same memory during the test

it can lead to non-determinism error (the Lincheck would not be able to replay the execution trace, because the content of the static memory changed). 

[Above by statically-reachable memory we mean any mutable state reachable from static variables.]

Note that during the test statically-reachable memory can be modified both by 
- static initialization blocks (`<clinit>`),
- and the regular functions from the test.

Thus, in order to fix possible non-determinism errors, we need to track modifications performed by `<clinit>` blocks, save them into trace, and manually re-perform them during replay (in the same order as in the original execution).

dmitrii-artuhov:
Todo list for proper support of the static memory restoring:
- [x] https://github.com/JetBrains/lincheck/pull/418 the `leaking this` problem, which is not fully solved in this PR: currently the code below will not work with the lazy snapshot tracking, because my detection of `this`'s fields modifications in constructors is implemented incorrectly (I merely check if `methodName == "<init>" && fieldOwnerClass == transformingClass` then the bytecode instructions are ignored. The correct way would be to check what `obj` owns the modified fields. And if `obj == this` we should only then skip these instructions, otherwise track them. However, I don't see how that can be achieved right now, because of #424
   ```kotlin
   class LincheckTest {
      class Node(var a: Int) {
          constructor(other: Node) : this(2) { // constructor accepting the same type
              other.a = 0 // in case if other is static variable, the change should be tracked
                          // because `other != this` here, but in my solution it will be missed.
          }
      }
      
      companion object {
          val staticNode = Node(1)
      }
      
      @Operation
      fun modify() {
          val localNode = Node(staticNode) // here will be a write to staticNode.a = 0, which is missed
      }
    }

   ```
- [x] https://github.com/JetBrains/lincheck/pull/418 in the constructor calls instrumentation there should be check for sub-classing and not the pure equals, see [TODO](https://github.com/JetBrains/lincheck/pull/418/files#diff-929e89c2c95100f180810601c429004d61b640aa58d84075bc001c2f75616654R50). Right now there is no obvious way of doing so during class transformation with loading the class hierarchy (which is discouraged during transformation stage).
- [ ] `System.arraycopy(...)` this call is not considered as modification, so overridden values will not be remembered. This call is used in `ArrayList::remove(...)` for example
[...]

PR #418, opened 2024-10-17
Static memory snapshot collecting and restoring

dmitrii-artuhov:
This PR partly resolves issue #389 

Here we introduced a lazy-tracking feature which aims to only track and restore variables reachable from static memory iff the variables were accessed perviously via read or write operations. Yet, there are still problems (see issue) that are left to be tackled in separate PRs.  

Proposed solutions in this PR:
- lazy tracking for most of the variables
- snapshot of collections are stored enegetically (meaning that the whole object graph is snapshotted). There are several APIs that are not handled yet, which does not allow a correct lazy snapshot building (e.g. `System.arraycopy(...)`)
- leaking this problem: I added a fix, which allows the test mentioned in the issue to pass, [see](https://github.com/JetBrains/lincheck/pull/418/files#diff-929e89c2c95100f180810601c429004d61b640aa58d84075bc001c2f75616654R30-R48). The TODO (calculating common superclass without loading additional classes to the VM) was resolved by leveraging `SafeClassWriter::getCommonSuperClass(...)`. Thanks to @ndkoval for help!


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingCTest.java
```diff
@@ -111,6 +111,11 @@
     @Deprecated(message = "Does nothing, because equals/hashcode don't always improve performance of verification")
     boolean requireStateEquivalenceImplCheck() default false;
 
+    /**
+     * Should the managed strategy use the static memory tracking and restoring algorithm.
+     */
+    boolean restoreStaticMemory() default DEFAULT_RESTORE_STATIC_MEMORY;
```
ndkoval, line 117: Should we remove this parameter and always restore the memory?

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -32,6 +32,10 @@ public interface EventTracker {
     void beforeNewObjectCreation(String className);
     void afterNewObjectCreation(Object obj);
 
+    void updateSnapshotOnFieldAccess(Object obj, String className, String fieldName, int codeLocation);
+    void updateSnapshotOnArrayElementAccess(Object array, int index, int codeLocation);
+    void updateSnapshotWithEnergeticTracking(Object[] objs);
```
eupp, line 37: Let's not expose `updateSnapshot` functions on `EventTracker` level, and instead call them from existing functions like `beforeReadField`, etc. 

The reasons for that are as follows. 

1) Snapshot collection and restoring logic is an implementation detail (of `ManagedStrategy`). Different strategies may either not need it at all, or implement their own different algorithms. 

2) The signature of `updateSnapshotOnFieldAccess` and `beforeReadField` is very similar. 

3) The code of `SharedMemoryAcessTransformer` and `SnapshotTrackerTransformer` that handles instructions like `GETFIELD`, `PUTFIELD`, etc, is very similar. In the long run, it would be more difficult to maintain two similar versions of APIs and transformer implementations.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,235 @@
...
+                        pushNull()
+                        push(owner.canonicalClassName)
+                        push(fieldName)
+                        loadNewCodeLocationId()
+                        // STACK: [<empty> | value], null, className, fieldName, codeLocation
+                        invokeStatic(Injections::updateSnapshotOnFieldAccess)
+                        // STACK: [<empty> | value]
+                        visitFieldInsn(opcode, owner, fieldName, desc)
+                        // STACK: [<empty> | value]
+                    }
```
eupp, line 129: This code is very similar to the code in `SharedMemoryAcessTransformer`. Let's unify the two and use only `SharedMemoryAccessTransformer`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,235 @@
...
+                code = {
+                    // STACK: args
+                    val arguments = storeArguments(desc)
+                    val matchedLocals = arguments.filterIndexed { index, _ -> matchedArguments.contains(index) }
+                    // STACK: <empty>
+                    push(matchedLocals.size)
+                    // STACK: length
+                    visitTypeInsn(ANEWARRAY, "java/lang/Object")
+                    // STACK: array
+                    matchedLocals.forEachIndexed { index, local ->
```
eupp, line 79: There is a similar function for this purpose in `TransformationUtils.kt` --- `storeLocals`, please use it instead.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,235 @@
...
+     * But sometimes `GETFIELD`/`PUTFIELD` are invoked on fields with owner objects are the same type as a constructor owner,
+     * but which are not `this`. This can happen when:
+     * - a non-static object is accessed which was passed to constructor arguments
+     * - a non-static object is accessed which was created locally in constructor
+     *
+     * We only case about the 1nd case, because for the 3nd case when dealing with
+     * a locally constructed object it can't be reached from static memory, so no need to track its fields.
+     * The 2nd case where such object is passed via constructor argument is handled here.
+     */
+    override fun visitMethodInsn(opcode: Int, owner: String, name: String, desc: String, itf: Boolean) = adapter.run {
```
eupp, line 49: This part of the code can be left in `SnapshotTrackerTransformer`, the rest should be unified with `SharedMemoryAccessTransformer`. Then we probably would want to rename `SnapshotTrackerTransformer` into something like `ConstructorArgumentsSnapshotTrackerTransformer`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,235 @@
...
+                )
+            }
+
+            else -> {
+                visitInsn(opcode)
+            }
+        }
+    }
+
+    private fun isArray(type: Type): Boolean = type.sort == Type.ARRAY
```
eupp, line 221: Please move `isArray` and `isPrimitive` into `TransformationUtils.kt` and add kdoc.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,235 @@
...
+
+    private fun isPrimitive(type: Type): Boolean {
+        val sort = type.sort
+        return sort == Type.BOOLEAN ||
```
eupp, line 225: Please use `when` here instead of chain of `||`

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SharedMemoryAccessTransformer.kt
```diff
@@ -270,55 +268,4 @@ internal class SharedMemoryAccessTransformer(
...
-    */
-    private fun getArrayAccessTypeFromStack(position: Int): Type? {
-        if (analyzer.stack == null) return null
-        val arrayDesc = analyzer.stack[analyzer.stack.size - position]
-        check(arrayDesc is String)
-        val arrayType = getType(arrayDesc)
-        check(arrayType.sort == ARRAY)
-        check(arrayType.dimensions > 0)
-        return getType(arrayDesc.substring(1))
-    }
```
eupp, line 323: Please unify `SharedMemoryAccessTransformer` and `SnapshotTrackerTransformer` code and then rollback these changes.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
```diff
@@ -41,11 +41,11 @@ import java.util.*
...
  *   Returns an object to be traversed next.
  */
-internal inline fun traverseObjectGraph(
+internal fun traverseObjectGraph(
     root: Any,
     traverseStaticFields: Boolean = false,
-    onField: (obj: Any, field: Field, value: Any?) -> Any?,
-    onArrayElement: (array: Any, index: Int, element: Any?) -> Any?,
+    onField: ((obj: Any, field: Field, value: Any?) -> Any?)?,
+    onArrayElement: ((array: Any, index: Int, element: Any?) -> Any?)?,
```
eupp, line 48: Why `onField` and `onArrayElement` can be `null` now?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -213,6 +217,96 @@ internal val Throwable.text: String get() {
...
+fun getFieldOffset(field: Field): Long {
+    return if (Modifier.isStatic(field.modifiers)) {
+        UnsafeHolder.UNSAFE.staticFieldOffset(field)
+    }
+    else {
+        UnsafeHolder.UNSAFE.objectFieldOffset(field)
+    }
+}
+
+internal fun getArrayElementOffset(arr: Any, index: Int): Long {
```
eupp, line 277: Isn't it a duplicate of `getArrayElementOffsetViaUnsafe`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -90,6 +90,11 @@ abstract class Strategy protected constructor(
     override fun close() {
         runner.close()
     }
+
+    /**
+     * Restores recorded values of all memory reachable from static state.
+     */
+    open fun restoreStaticMemorySnapshot() {}
```
eupp, line 97: Let's keep it only on the level of `ManagedStrategy` and do not expose on `Stategy` API level.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -101,15 +106,20 @@ abstract class Strategy protected constructor(
...
-        val failure = verify(result, verifier)
+        restoreStaticMemorySnapshot()
+        failure = verify(result, verifier)
         if (failure != null)
-            return failure
+            break
     }
-    return null
+
+    restoreStaticMemorySnapshot()
```
eupp, line 121: Should be called from `ManagedStrategy`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/UnsafeHolder.kt
```diff
@@ -10,6 +10,7 @@
 
 package org.jetbrains.kotlinx.lincheck.util
 
+import org.jetbrains.kotlinx.lincheck.getArrayElementOffset
```
eupp, line 13: Please remove this import

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -213,6 +217,96 @@ internal val Throwable.text: String get() {
...
+
+        // Move up the hierarchy
+        clazz = clazz.superclass
+    }
+
+    throw NoSuchFieldException("Class '${this.name}' does not have field '$fieldName'")
+}
+
+@Suppress("DEPRECATION")
+fun getFieldOffset(field: Field): Long {
```
eupp, line 268: Please rename to `getFieldOffsetViaUnsafe` and move into `UnsafeHolder.kt`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+        class ArrayCellDescriptor(val index: Int) : Descriptor()
+    }
+
+    private sealed class MemoryNode(
+        val descriptor: Descriptor,
+        val initialValue: Any?
+    ) {
+        class RegularFieldNode(descriptor: FieldDescriptor, initialValue: Any?) : MemoryNode(descriptor, initialValue)
+        class StaticFieldNode(descriptor: FieldDescriptor, initialValue: Any?) : MemoryNode(descriptor, initialValue)
+        class ArrayCellNode(descriptor: ArrayCellDescriptor, initialValue: Any?) : MemoryNode(descriptor, initialValue)
```
eupp, line 49: Do you need different subclasses here? I think the type of node is fully defined by the `Descriptor` (i.e. `FieldDescriptor`, `ArrayCellDescriptor`, etc).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    fun trackField(obj: Any?, className: String, fieldName: String) {
+        if (obj != null && obj !in trackedObjects) return
+
+        val clazz: Class<*> = Class.forName(className).let {
+            if (obj != null) it
+            else it.findField(fieldName).declaringClass
+        }
+        val field = clazz.findField(fieldName)
+        val readResult = runCatching { readFieldViaUnsafe(obj, field) }
```
eupp, line 60: Please re-use similar logic for reading field from `ObjectGraph.kt`. Extract it into a function.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+        class ArrayCellNode(descriptor: ArrayCellDescriptor, initialValue: Any?) : MemoryNode(descriptor, initialValue)
+    }
+
+    fun trackField(obj: Any?, className: String, fieldName: String) {
+        if (obj != null && obj !in trackedObjects) return
+
+        val clazz: Class<*> = Class.forName(className).let {
+            if (obj != null) it
+            else it.findField(fieldName).declaringClass
+        }
```
eupp, line 58: Please extract this code into a function and give it a meaningful name.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+        val field = clazz.findField(fieldName)
+        val readResult = runCatching { readFieldViaUnsafe(obj, field) }
+
+        if (readResult.isSuccess) {
+            val fieldValue = readResult.getOrNull()
+            trackSingleField(obj, clazz, field, fieldValue) {
+                if (shouldTrackEnergetically(fieldValue)) {
+                    trackHierarchy(fieldValue!!)
```
eupp, line 66: Please rename into `trackReachableObjectSubgraph`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+            if (obj != null) it
+            else it.findField(fieldName).declaringClass
+        }
+        val field = clazz.findField(fieldName)
+        val readResult = runCatching { readFieldViaUnsafe(obj, field) }
+
+        if (readResult.isSuccess) {
+            val fieldValue = readResult.getOrNull()
+            trackSingleField(obj, clazz, field, fieldValue) {
+                if (shouldTrackEnergetically(fieldValue)) {
```
eupp, line 65: I think, a more widely used (in programming) antonym of `lazy` is `eager`, not `energetic`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    private fun trackSingleField(
+        obj: Any?,
+        clazz: Class<*>,
+        field: Field,
+        fieldValue: Any?,
+        callback: (() -> Unit)? = null
+    ) {
+        val nodesList =
+            if (obj != null) trackedObjects[obj]
```
eupp, line 110: Shouldn't it also be `trackedObjects.getOrPut(obj) { ... }` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+        val childNode = createMemoryNode(
+            obj,
+            FieldDescriptor(field, getFieldOffset(field)),
+            fieldValue
+        )
+
+        nodesList.add(childNode)
+        if (isTrackableObject(childNode.initialValue)) {
+            trackedObjects.putIfAbsent(childNode.initialValue, mutableListOf<MemoryNode>())
+            callback?.invoke()
```
eupp, line 130: Let's just return `Boolean` instead of passing `callback`, it would simplify the code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,231 @@
...
+                else if (!Modifier.isFinal((node.descriptor as FieldDescriptor).field.modifiers)) {
+                    writeFieldViaUnsafe(
+                        if (node is StaticFieldNode) null else obj,
+                        node.descriptor.field,
+                        node.initialValue
+                    )
+                }
+
+                if (isTrackableObject(node.initialValue)) {
+                    restoreValues(node.initialValue!!, visitedObjects)
```
eupp, line 208: Could we rewrite this DFS to use stack/queue instead of recursion?

src/jvm/test/resources/expected_logs/var_handle/varhandle_boolean_representation.txt
```diff
@@ -13,19 +13,19 @@ The following interleaving leads to the error:
...
+| operation()                                                                                                                                                            |
+|   number.READ: false at VarHandleBooleanRepresentationTest.operation(VarHandleRepresentationTests.kt:216)                                                              |
+|   number.compareAndSet(false,false): true at VarHandleBooleanRepresentationTest.operation(VarHandleRepresentationTests.kt:216)                                         |
+|   number.set(true) at VarHandleBooleanRepresentationTest.operation(VarHandleRepresentationTests.kt:217)                                                                |
+|   VarHandleBooleanRepresentationTest.staticNumber.READ: true at VarHandleBooleanRepresentationTest.operation(VarHandleRepresentationTests.kt:219)                      |
+|   VarHandleBooleanRepresentationTest.staticNumber.compareAndSet(true,false): true at VarHandleBooleanRepresentationTest.operation(VarHandleRepresentationTests.kt:219) |
```
eupp, line 24: Why the traces in these tests have changed?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/modelchecking/snapshot/AtomicFUSnapshotTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        //  Because of this, methods, such as `AtomicInt::getAndIncrement()`, will not be tracked as modification of `value` field.
+        //  Tracking of atomicfu classes should be implemented energetically, the same way as for java atomics
+        //  in order to handle modification that do not reference `value` field directly.
+        //  If energetic traversal does not help/not possible to reach `value` field, then such indirect methods should be handled separately,
+        //  the same way as reflexivity, var-handles, and unsafe by tracking the creation of java afu and retrieving the name of the associated field.
+        private val atomicFUInt = kotlinx.atomicfu.atomic(1)
+        private val atomicFURef = kotlinx.atomicfu.atomic<Wrapper>(Wrapper(1))
+
+        private val atomicFUIntArray = kotlinx.atomicfu.AtomicIntArray(3)
+        private val atomicFURefArray = kotlinx.atomicfu.atomicArrayOfNulls<Wrapper>(3)
```
eupp, line 33: Could you import `kotlinx.atomicfu` to shorten the names?

eupp (changes requested):

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
dmitrii-artuhov: The `updateSnapshotWithEnergeticTracking` should be exposed, since I need to call it from the `Injections` class and there I cannot check for `getEventTracker() instanceof ManagedStrategy`, thus, I need to have this method in the `EventTracker` interface

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
dmitrii-artuhov: Optimization for methods which do not need one or the other callback. So that I do not traverse arrays when the `onArrayElement` call is not needed (specifically, for `FieldSearchHelper` class)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingCTest.java
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -216,12 +219,20 @@ abstract class ManagedStrategy(
...
+        }
+    }
+
     /**
      * Runs the current invocation.
      */
     override fun runInvocation(): InvocationResult {
         while (true) {
             initializeInvocation()
+            restoreStaticMemorySnapshot()
```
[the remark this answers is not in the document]
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: I removed Descriptors and left onnly MemoryNodes

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: done

src/jvm/test/resources/expected_logs/var_handle/varhandle_boolean_representation.txt
dmitrii-artuhov: `staticNumber` is now restored and multiple `staticNumber.compareAndSet(...)` calls on it return the same value as the 1st invocation (which wasn't the case before this PR)

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SnapshotTrackerTransformer.kt
dmitrii-artuhov: `pushArray(matchedLocals)` you probably meant
done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -213,21 +212,63 @@ internal val Throwable.text: String get() {
...
+        // Move up the hierarchy
+        clazz = clazz.superclass
     }
-    return null // Field not found
+
+    throw NoSuchFieldException("Class '${this.name}' does not have field '$fieldName'")
+}
+
+/**
+ * Reads a [field] of the owner object [obj] via Unsafe, and falls back onto the reflexivity in case of failure.
```
eupp, line 263:
```suggestion
 * Reads a [field] of the owner object [obj] via Unsafe,  in case of failure fallbacks into reading the field via reflection.
```

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -32,6 +32,7 @@ public interface EventTracker {
     void beforeNewObjectCreation(String className);
     void afterNewObjectCreation(Object obj);
 
+    void updateSnapshotWithEagerTracking(Object[] objs);
```
eupp, line 35: Please add a newline to separate `updateSnapshotWithEagerTracking` from `beforeReadXXX` methods.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedCTestConfiguration.kt
```diff
@@ -52,6 +52,7 @@ abstract class ManagedCTestConfiguration(
         const val DEFAULT_CHECK_OBSTRUCTION_FREEDOM = false
         const val DEFAULT_HANGING_DETECTION_THRESHOLD = 101
         const val LIVELOCK_EVENTS_THRESHOLD = 10001
+        const val DEFAULT_RESTORE_STATIC_MEMORY = true
```
eupp, line 55: Do you still need this new constant?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedOptions.kt
```diff
@@ -50,6 +52,13 @@ abstract class ManagedOptions<OPT : Options<OPT, CTEST>, CTEST : CTestConfigurat
         this.hangingDetectionThreshold = hangingDetectionThreshold
     }
 
+    /**
+     * Set to `true` to activate the static memory snapshot tracking and restoring algorithm.
+     */
+    fun restoreStaticMemory(restoreStaticMemory: Boolean): OPT = applyAndCast {
+        this.restoreStaticMemory = restoreStaticMemory
+    }
```
eupp, line 60: I thought it was decided to remove this parameter?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
eupp: Ok

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -145,8 +152,10 @@ internal class LincheckClassVisitor(
         // which should be put in front of the byte-code transformer chain,
         // so that it can correctly analyze the byte-code and compute required type-information
         mv = run {
-            val sv = SharedMemoryAccessTransformer(fileName, className, methodName, mv.newAdapter())
+            val st = ConstructorArgumentsSnapshotTrackerTransformer(fileName, className, methodName, mv.newAdapter(), classVisitor::isInstanceOf)
```
eupp, line 155: I think in this case we know `method != "<init>"` so there is no need to insert `ConstructorArgumentsSnapshotTrackerTransformer`

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SharedMemoryAccessTransformer.kt
```diff
@@ -33,7 +33,12 @@ internal class SharedMemoryAccessTransformer(
     lateinit var analyzer: AnalyzerAdapter
 
     override fun visitFieldInsn(opcode: Int, owner: String, fieldName: String, desc: String) = adapter.run {
-        if (isCoroutineInternalClass(owner) || isCoroutineStateMachineClass(owner)) {
+        if (
+            isCoroutineInternalClass(owner) ||
+            isCoroutineStateMachineClass(owner) ||
+            // when initializing our own fields in constructor, we do not want to track that
```
eupp, line 39: please add to the comment that it will result in bytecode `VerificationError`

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
```diff
@@ -41,11 +41,11 @@ import java.util.*
  * @param onArrayElement callback for array elements traversal, accepts `(array, index, elementValue)`.
  *   Returns an object to be traversed next.
  */
-internal inline fun traverseObjectGraph(
```
eupp, line 44: why `inline` was removed?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
eupp: Let's then add typealiases: 

```
internal typealias FieldCallback = ((obj: Any, field: Field, value: Any?) -> Any?)
internal typealias ArrayElementCallback = ((obj: Any, field: Field, value: Any?) -> Any?)
```

and then use `FieldCallback?` and `ArrayElementCallback?` as argument types.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/UnsafeHolder.kt
```diff
@@ -79,4 +79,81 @@ internal fun getArrayElementOffsetViaUnsafe(arr: Any, index: Int): Long {
...
+    return if (Modifier.isStatic(field.modifiers)) {
+        UnsafeHolder.UNSAFE.staticFieldOffset(field)
+    }
+    else {
+        UnsafeHolder.UNSAFE.objectFieldOffset(field)
+    }
+}
+
+@Suppress("DEPRECATION")
+internal fun findFieldNameByOffset(targetType: Class<*>, offset: Long): String? {
```
eupp, line 146: Shouldn't be `findFieldNameByOffsetViaUnsafe` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
eupp: But do you need additional call to `restoreStaticMemorySnapshot` here, if there is already a call to it in `ManagedStrategy::run` ?

eupp (commented):
@dmitrii-artuhov Please fix a few remaining minor issues, otherwise looks good to me.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: done

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SharedMemoryAccessTransformer.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/ObjectGraph.kt
dmitrii-artuhov: otherwise callbacks are non-nullable

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/UnsafeHolder.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedCTestConfiguration.kt
dmitrii-artuhov: remove, I forgot about these parts(

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedOptions.kt
dmitrii-artuhov: done

dmitrii-artuhov:
> But do you need additional call to `restoreStaticMemorySnapshot` here, if there is already a call to it in `ManagedStrategy::run`?
@eupp I've rewritten it, see the modified `Strategy.runIteration(...)` function, please

eupp (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ConstructorArgumentsSnapshotTrackerTransformer.kt
```diff
@@ -0,0 +1,85 @@
...
+     */
+    override fun visitMethodInsn(opcode: Int, owner: String, name: String, desc: String, itf: Boolean) = adapter.run {
+        if (name == "<init>") {
+            val matchedArguments = getArgumentTypes(desc).toList()
+                .mapIndexed { index, type ->
+                    if (
+                        !isArray(type) &&
+                        !isPrimitive(type) &&
+                        isInstanceOf(type.className.replace('.', '/'), owner)
+                    ) index
```
ndkoval, line 55: Please add `{ .. }` for both branches when the condition is multiline

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -0,0 +1,255 @@
...
+
+
+/**
+ * Manages a snapshot of the global static state.
+ *
+ * This class only tracks static memory and memory reachable from it,
+ * referenced from the test code. So the whole static state is not recorded, but only a subset of that,
+ * which is named *snapshot*.
+ */
+class SnapshotTracker {
```
ndkoval, line 36: Should it be `SnapshotTracker` or `MemoryChangesTracker`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -216,13 +219,26 @@ abstract class ManagedStrategy(
...
      */
     override fun runInvocation(): InvocationResult {
         while (true) {
             initializeInvocation()
-            val result = runner.run()
+            val result: InvocationResult
+            try {
+                result = runner.run()
+            }
+            finally {
```
ndkoval, line 239: Please place `finally` on the previous line

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -216,13 +219,26 @@ abstract class ManagedStrategy(
...
+
     /**
      * Runs the current invocation.
      */
     override fun runInvocation(): InvocationResult {
         while (true) {
             initializeInvocation()
-            val result = runner.run()
+            val result: InvocationResult
+            try {
```
ndkoval, line 236: `val result = try { ... }`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -909,6 +929,29 @@ abstract class ManagedStrategy(
...
+     * or one this objects which contains it is already stored.
+     */
+    fun updateSnapshotOnFieldAccess(obj: Any?, className: String, fieldName: String) = runInIgnoredSection {
+        staticMemorySnapshot.trackField(obj, className, fieldName)
+    }
+
+    /**
+     * Tracks a specific [array] element at [index], if the [array] is already tracked.
+     */
+    fun updateSnapshotOnArrayElementAccess(array: Any, index: Int) = runInIgnoredSection {
```
ndkoval, line 943: "access" or "update"? (the same for other functions)

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/SafeClassWriter.java
```diff
@@ -53,6 +54,10 @@ public SafeClassWriter(ClassReader cr, ClassLoader loader, final int flags) {
         this.loader = loader != null ? loader : ClassLoader.getSystemClassLoader();
     }
 
+    public boolean isInstanceOf(final String actualType, final String expectedSuperType) {
```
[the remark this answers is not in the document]
dmitrii-artuhov: Why current names do not fit? What names do you suggest?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: We need to invoke this method for both reads and writes to the array elements (same for other functions), so the correct name would be `*Access(...)`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: This should be tracker, because its methods are invoked even if no chnages to static memory were made (a read, for example, which we must track, but there is no immidiate change in memory)

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -91,6 +90,14 @@ internal class LincheckClassVisitor(
         }
         if (methodName == "<init>") {
             mv = ObjectCreationTransformer(fileName, className, methodName, mv.newAdapter())
+            mv = run {
```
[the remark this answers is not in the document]
dmitrii-artuhov: I did not try including `<init>` methods in the whole transformation chain. Right now I only require tracking fields/arrays accesses and other constructor invocations inside `<init>` method bodies. 
Maybe the feature of integrating constructors in full transformation chain should be a separate PR, since there are might be tricky spots where to account for https://github.com/JetBrains/lincheck/issues/424

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -101,15 +102,26 @@ abstract class Strategy protected constructor(
...
     for (invocation in 0 until invocations) {
-        if (!nextInvocation())
-            return null
+        if (!nextInvocation()) break
         val result = runInvocation()
-        val failure = verify(result, verifier)
-        if (failure != null)
-            return failure
+
+        failure =
```
ndkoval, line 111: Can you put `try` on the same line, please?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -101,15 +102,26 @@ abstract class Strategy protected constructor(
  * @return the failure, if detected, null otherwise.
  */
 fun Strategy.runIteration(invocations: Int, verifier: Verifier): LincheckFailure? {
+    var failure: LincheckFailure? = null
```
ndkoval, line 105: Let's keep the previous approach with `return`s in the loop

ndkoval (changes requested):

ndkoval (approved):

PR #420, opened 2024-10-22
Utility function for unified object hierarchy traversing

### LINKED ISSUE #402: `UnsupportedOperationException` after adding a custom scenario

dmitrii-artuhov:
- Adds a utility function for proper object graph traversing and rewrites all places where such functionality was used
- Fixes #402

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
@@ -49,70 +49,65 @@ private fun enumerateObjects(obj: Any, objectNumberMap: MutableMap<Any, Int>) {
     if (obj is Class<*> || obj is ClassLoader) return
     objectNumberMap[obj] = getObjectNumber(obj.javaClass, obj)
 
-    val processObject: (Any?) -> Any? = processObject@{ value: Any? ->
-        if (value == null || value is Class<*> || value is ClassLoader) return@processObject null
+    val processObject: (Any?) -> Any? = { value: Any? ->
+        if (value == null || value is Class<*> || value is ClassLoader) null
+        else {
+            // We jump through most of the atomic classes
+            var jumpObj: Any? = value
 
-        // We jump through most of the atomic classes
-        var jumpObj: Any? = value
-
-        // Special treatment for java atomic classes, because they can be extended but user classes,
-        // in case if a user extends java atomic class, we do not want to jump through it.
-        while (jumpObj?.javaClass?.name != null && isAtomicJavaClass(jumpObj.javaClass.name)) {
-            jumpObj = jumpObj.javaClass.getMethod("get").invoke(jumpObj)
-        }
-
-        if (isAtomicFU(jumpObj)) {
-            val readNextJumpObjectByFieldName = { fieldName: String ->
-                readFieldViaUnsafe(jumpObj, jumpObj?.javaClass?.getDeclaredField(fieldName)!!)
-            }
-
-            while (jumpObj is kotlinx.atomicfu.AtomicRef<*>) {
-                jumpObj = readNextJumpObjectByFieldName("value")
+            // Special treatment for java atomic classes, because they can be extended but user classes,
+            // in case if a user extends java atomic class, we do not want to jump through it.
+            while (jumpObj?.javaClass?.name != null && isAtomicJavaClass(jumpObj.javaClass.name)) {
+                jumpObj = jumpObj.javaClass.getMethod("get").invoke(jumpObj)
             }
 
             if (isAtomicFU(jumpObj)) {
-                jumpObj =
-                    if (jumpObj is kotlinx.atomicfu.AtomicBoolean) readNextJumpObjectByFieldName("_value")
-                    else readNextJumpObjectByFieldName("value")
+                val readNextJumpObjectByFieldName = { fieldName: String ->
+                    readFieldViaUnsafe(jumpObj, jumpObj?.javaClass?.getDeclaredField(fieldName)!!)
+                }
+
+                while (jumpObj is kotlinx.atomicfu.AtomicRef<*>) {
+                    jumpObj = readNextJumpObjectByFieldName("value")
+                }
+
+                if (isAtomicFU(jumpObj)) {
+                    jumpObj =
+                        if (jumpObj is kotlinx.atomicfu.AtomicBoolean) readNextJumpObjectByFieldName("_value")
+                        else readNextJumpObjectByFieldName("value")
+                }
             }
-        }
 
-        if (jumpObj != null) {
-            objectNumberMap[jumpObj] = getObjectNumber(jumpObj.javaClass, jumpObj)
-            return@processObject if (shouldAnalyseObjectRecursively(jumpObj, objectNumberMap)) jumpObj else null
+            if (jumpObj != null) {
+                objectNumberMap[jumpObj] = getObjectNumber(jumpObj.javaClass, jumpObj)
+                if (shouldAnalyseObjectRecursively(jumpObj, objectNumberMap)) jumpObj else null
+            }
+            else null
         }
-
-        return@processObject null
     }
 
     traverseObjectGraph(
         obj,
         onArrayElement = { _, _, value ->
-            if (value?.javaClass?.isEnum == true) return@traverseObjectGraph null
-
-            try {
-                return@traverseObjectGraph processObject(value)
-            }
-            catch (e: Throwable) {
+            if (value?.javaClass?.isEnum == true) null
+            else try {
+                processObject(value)
+            } catch (e: Throwable) {
                 e.printStackTrace()
+                null
             }
-
-            return@traverseObjectGraph null
         },
         onField = { _, f, value ->
             if (
                 Modifier.isStatic(f.modifiers) ||
                 f.isEnumConstant ||
                 f.name == "serialVersionUID"
-            ) return@traverseObjectGraph null
-
-            try {
-                return@traverseObjectGraph processObject(value)
+            ) null
+            else try {
+                processObject(value)
             } catch (e: Throwable) {
                 e.printStackTrace()
+                null
             }
-
-            return@traverseObjectGraph null
         }
     )
 }
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
@@ -52,11 +52,13 @@ private fun enumerateObjects(obj: Any, objectNumberMap: MutableMap<Any, Int>) {
     val processObject: (Any?) -> Any? = processObject@{ value: Any? ->
         if (value == null || value is Class<*> || value is ClassLoader) return@processObject null
 
-        // we jump through most of the atomic classes
+        // We jump through most of the atomic classes
         var jumpObj: Any? = value
 
-        if (isAtomic(jumpObj)) {
-            jumpObj = jumpObj?.javaClass?.getMethod("get")?.invoke(jumpObj)
+        // Special treatment for java atomic classes, because they can be extended but user classes,
+        // in case if a user extends java atomic class, we do not want to jump through it.
+        while (jumpObj?.javaClass?.name != null && isAtomicClass(jumpObj.javaClass.name)) {
+            jumpObj = jumpObj.javaClass.getMethod("get").invoke(jumpObj)
         }
 
         if (isAtomicFU(jumpObj)) {
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
@@ -21,9 +21,7 @@
 package org.jetbrains.kotlinx.lincheck
 
 import org.jetbrains.kotlinx.lincheck.strategy.managed.ObjectLabelFactory.getObjectNumber
-import org.jetbrains.kotlinx.lincheck.util.isAtomic
-import org.jetbrains.kotlinx.lincheck.util.isAtomicFU
-import org.jetbrains.kotlinx.lincheck.util.readFieldViaUnsafe
+import org.jetbrains.kotlinx.lincheck.util.*
 import java.lang.reflect.Modifier
 import java.math.BigDecimal
 import java.math.BigInteger
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
@@ -21,13 +21,10 @@
 package org.jetbrains.kotlinx.lincheck
 
 import org.jetbrains.kotlinx.lincheck.strategy.managed.ObjectLabelFactory.getObjectNumber
-import org.jetbrains.kotlinx.lincheck.util.readFieldViaUnsafe
-import sun.misc.Unsafe
-import java.lang.reflect.Field
+import org.jetbrains.kotlinx.lincheck.util.readField
 import java.lang.reflect.Modifier
 import java.math.BigDecimal
 import java.math.BigInteger
-import java.util.*
 import java.util.concurrent.atomic.*
 import kotlin.coroutines.Continuation
 
@@ -39,8 +36,7 @@ import kotlin.coroutines.Continuation
  */
 internal fun enumerateObjects(obj: Any): Map<Any, Int> {
     val objectNumberMap = hashMapOf<Any, Int>()
-    enumerateObjects(obj, Collections.newSetFromMap(IdentityHashMap()), objectNumberMap)
-
+    enumerateObjects(obj, objectNumberMap)
     return objectNumberMap
 }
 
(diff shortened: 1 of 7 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,106 +36,47 @@ import kotlin.coroutines.Continuation
...
-                }
-                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicInt") {
-                    value = readField(value, value.javaClass.getDeclaredField("value"))
-                }
-                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicLong") {
-                    value = readField(value, value.javaClass.getDeclaredField("value"))
-                }
-                if (value?.javaClass?.canonicalName == "kotlinx.atomicfu.AtomicBoolean") {
-                    value = readField(value, value.javaClass.getDeclaredField("_value"))
-                }
```
eupp, line 91: Why does this special handling of atomic types was removed?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +232,83 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+ * set to `false`, then array fields will be traversed instead of elements.
+ *
+ * The [callback] is called for each tuple of `(fieldOwner, field, fieldValue)` of the [obj],
+ * the return value of the callback determines, what object should be traversed further
+ * (it can be something different from `fieldValue`).
+ *
+ * Parameters of [callback]:
+ * - `fieldOwner`: either `Any` - some class instance that owns the field, or `Class<*>` is case if `field` is static.
+ * - `field`: the field that currently is traversed.
+ * - `fieldValue`: the value that was read from the field.
```
eupp, line 248: Please document all the arguments of the function using KDoc syntax.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +232,83 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+                    }
+                }
+                emptyList<Field>()
+            }
+            else if (
+                (currentObj is AtomicReferenceArray<*> || currentObj is kotlinx.atomicfu.AtomicArray<*>) &&
+                treatArrayElementsAsFields
+            ) {
+                // Do the same as before, but for atomic arrays (with non-primitive types)
+                val getElementAt = currentObj.javaClass.getMethod("get", Int::class.javaPrimitiveType)
```
eupp, line 277: Instead of using reflection, you can just cast to `AtomicReferenceArray<*>` or `kotlinx.atomicfu.AtomicArray<*>` type, and use the required methods directly.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +232,83 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+                ) as Int
+
+                for (index in 0..length - 1) {
+                    getElementAt.invoke(currentObj, index)?.let { queue.add(it) }
+                }
+                emptyList<Field>()
+            }
+            else currentObj.javaClass.allDeclaredFieldWithSuperclasses
+
+        for (f in fields) {
```
eupp, line 290: Let's rename `f` to `field`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,106 +38,72 @@ import kotlin.coroutines.Continuation
...
+
+        try {
+            if (value == null || value is Class<*> || value is ClassLoader) return@traverseObjectHierarchy null
+            objectNumberMap[value] = getObjectNumber(value.javaClass, value)
+
+            var jumpValue: Any? = value
+
+            // we jump through most of the atomic classes
+            if (isAtomic(jumpValue)) {
+                jumpValue = jumpValue!!.javaClass.getDeclaredMethod("get").invoke(jumpValue)
```
eupp, line 70: Why don't just put this special handling of atomic types directly into `traverseObjectHierarchy` function?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +232,83 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+            // We wrap an unsafe read into `runCatching` to hande `UnsupportedOperationException`,
+            // which can be thrown, for instance, when attempting to read a field of
+            // a hidden class (starting from Java 15).
+            val result = runCatching { readField(currentObj, f) }
+            if (result.isFailure) continue // do not pass fields to the user, that are non-readable by Unsafe
+
+            val fieldValue = result.getOrNull()
+            val nextObject = callback(currentObj, f, fieldValue)
```
eupp, line 298: Why the semantics of `callback` is to return `nextObject`, instead of just boolean `shouldTraverse` flag as before?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -10,19 +10,21 @@
...
-import sun.nio.ch.lincheck.EventTracker
-import sun.nio.ch.lincheck.TestThread
 import org.jetbrains.kotlinx.lincheck.runner.*
 import org.jetbrains.kotlinx.lincheck.strategy.managed.*
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransformer
 import org.jetbrains.kotlinx.lincheck.util.UnsafeHolder
+import org.jetbrains.kotlinx.lincheck.util.readField
 import org.jetbrains.kotlinx.lincheck.verifier.*
+import sun.nio.ch.lincheck.EventTracker
+import sun.nio.ch.lincheck.TestThread
```
eupp, line 20: Unused imports?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +232,83 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+ *
+ * In case [treatArrayElementsAsFields] is
+ * set to `false`, then array fields will be traversed instead of elements.
+ *
+ * The [callback] is called for each tuple of `(fieldOwner, field, fieldValue)` of the [obj],
+ * the return value of the callback determines, what object should be traversed further
+ * (it can be something different from `fieldValue`).
+ *
+ * Parameters of [callback]:
+ * - `fieldOwner`: either `Any` - some class instance that owns the field, or `Class<*>` is case if `field` is static.
```
eupp, line 246: Isn't there a semantics collision when some `Class<*>` instance is passed as object, and its non-static field is being inspected?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/UnsafeHolder.kt
```diff
@@ -34,4 +34,21 @@ internal inline fun <T> readFieldViaUnsafe(obj: Any?, field: Field, getter: Unsa
         val offset = UnsafeHolder.UNSAFE.objectFieldOffset(field)
         return UnsafeHolder.UNSAFE.getter(obj, offset)
     }
+}
+
+internal fun readField(obj: Any?, field: Field): Any? {
```
eupp, line 39: Let's rename to `readFieldViaUnsafe` too (there should be no ambiguity, because the other function with the same name takes additional lambda argument).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/FieldSearchHelper.kt
```diff
@@ -52,50 +44,37 @@ internal object FieldSearchHelper {
...
-                    InstanceOwnerWithName(field.name, testObject)
-                }
-                continue
-            }
-            when (val result = findObjectField(fieldValue, value, visitedObjects)) {
-                is FieldName -> {
-                    if (fieldName != null) {
-                        return MultipleFieldsMatching
+            if (value === fieldValue && !isTraverseCompleted()) {
+                if (fieldName != null) traverseResult = MultipleFieldsMatching
```
eupp, line 61: Please use `when` instead of `if-else-if` chains.

eupp, line 61: Please put `traverseResult = MultipleFieldsMatching` into parentheses `{ traverseResult = MultipleFieldsMatching } ` for clarity.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
eupp: Or, event worse, when a static field of `Class<*>` object is being inspected?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: They are used in `runInIgnoredSection(...)` method

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
dmitrii-artuhov: I don't see why we would want that. The utility function is supposed to traverse the object graph (even if the object is an atomic class), but not to perform sophisticated logic for skipping through fields.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: In order to allow skipping through fields

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
eupp: Atomic objects are a bit special in this regard. 
In many aspects, `AtomicInteger` is very similar to `Integer`, it is essentially just an integer, accesses to which are atomic. Similarly, `AtomicReference` is just a reference to another object.

I don't see in what cases we might want not to skip atomic classes.
Also, probably make sense to skip boxed primitive types (e.g. `Integer`), and some other classes (at least `String`) --- basically any immutable value type.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: I made it so for atomic arrays, but for atomicfu array I keep getting errors of type `org.jetbrains.kotlinx.lincheck.UtilsKt: standalone invocation of kotlinx.atomicfu.AtomicArray::getSize that was not traced to previous field load` when I access `size` field. When I use `get(...)` method I get the same kind of error. This seems to be related to [KT-47749](https://youtrack.jetbrains.com/issue/KT-47749) which is marked as fixed. Right now I don't understand why it happens in our case

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: I will change the description in javadoc, but it seems that traversing `Class<*>` is not needed in any of usage of this utility function in lincheck. In snapshot replaying I also will not need to traverse `Class<*>`, because I am starting a traverse of static variables by exact reference to their instances, and not via `Class<*>` of the declaring class.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: changed back to return `true`/`false`)

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
dmitrii-artuhov: Seems reasonable 👍

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
eupp: Let's then explicitly skip any `Class` objects in the traversal function.

eupp:
@dmitrii-artuhov thank you for working on this!

While we are here, could you also please move `ObjectTraverserTest.kt` from `lincheck` to `lincheck_test` package (it was added to incorrect package by mistake), and extend it by adding few other test cases?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -230,6 +243,125 @@ internal val Class<*>.allDeclaredFieldWithSuperclasses get(): List<Field> {
...
+                        },
+                        { onArrayElement(currentObj, index, it) }
+                    )
+                }
+            }
+            else -> {
+                // we jump through most of the atomic classes
+                var jumpObj: Any? = currentObj
+
+                if (isAtomic(jumpObj) && jumpObj !is AtomicReference<*> && jumpObj !is kotlinx.atomicfu.AtomicRef<*>) {
```
dmitrii-artuhov, line 334: - absence of `jumpObj !is AtomicReference<*>` causes [ObstructionFreedomViolationEventsCutTest](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildAggregated/4868725?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true&expandBuildTestsSection=true&expandBuildProblemsSection=true) to fail
- `jumpObj !is kotlinx.atomicfu.AtomicRef<*>` added for symmetry

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
dmitrii-artuhov: Hm, now when I use `traverseObjectGraph` in static memory snapshot restoring, I need to track fields of `AtomicInteger`, `AtomicLong`, etc.
So, maybe jump skipping must be a local optimization for `enumerateObject` method and not a common optimization. Then, I have to return `nextObj` from the callbacks `onField` and `onArrayElement`

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
dmitrii-artuhov: Removed it after this [refactoring](https://github.com/JetBrains/lincheck/pull/420#discussion_r1826093690)

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,107 +37,84 @@ import kotlin.coroutines.Continuation
...
-            obj.forEach { element -> element?.let { enumerateObjects(it, processedObjects, objectNumberMap) } }
+    objectNumberMap[obj] = getObjectNumber(obj.javaClass, obj)
+
+    val processObject: (Any?) -> Any? = processObject@{ value: Any? ->
+        if (value == null || value is Class<*> || value is ClassLoader) return@processObject null
+
+        // we jump through most of the atomic classes
+        var jumpObj: Any? = value
+
+        if (isAtomic(jumpObj)) {
```
dmitrii-artuhov, line 60: TODO: replace with `isAtomicClass(...)` using string comparison with classname. Due to a possibility to extend from `AtomicReference<*>` for example. So if we have `Node : AtomicReference<*>()` we want to traverse node's fields

eupp: TODO: also handle nested atomic references case:  `AtomicReference<AtomicReference<...>>`

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,107 +35,86 @@ import kotlin.coroutines.Continuation
...
+        while (jumpObj?.javaClass?.name != null && isAtomicClass(jumpObj.javaClass.name)) {
+            jumpObj = jumpObj.javaClass.getMethod("get").invoke(jumpObj)
+        }
+
+        if (isAtomicFU(jumpObj)) {
+            val readNextJumpObjectByFieldName = { fieldName: String ->
+                readFieldViaUnsafe(jumpObj, jumpObj?.javaClass?.getDeclaredField(fieldName)!!)
+            }
+
+            if (jumpObj is kotlinx.atomicfu.AtomicRef<*>) {
```
eupp, line 69: shouldn't it also be `while` loop here?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/ObjectTraverserTest.kt
```diff
@@ -0,0 +1,108 @@
...
+            objectEnumeration.keys.containsAll(
+                listOf(
+                    myObject,
+                    // atomicfu transformers [are insane and] don't allow to compile direct reference to `myObject.array`
+                    myObject.javaClass.getDeclaredField("array").get(myObject),
+                    1, 2, 3
+                )
+            )
+        )
+    }
```
eupp, line 106: You can add a test case here for `AtomicReference<AtomicReference<...>>` case here (and similar for AFU)

dmitrii-artuhov: added a test with inner reference for java-atomic and atomicfu references

eupp (approved):

eupp:
@ndkoval please have a look

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,107 +35,86 @@ import kotlin.coroutines.Continuation
...
-    objectNumberMap[obj] = objectNumber
-
-    var clazz: Class<*>? = obj.javaClass
-    if (clazz!!.isArray) {
-        if (obj is Array<*>) {
-            obj.forEach { element -> element?.let { enumerateObjects(it, processedObjects, objectNumberMap) } }
+    objectNumberMap[obj] = getObjectNumber(obj.javaClass, obj)
+
+    val processObject: (Any?) -> Any? = processObject@{ value: Any? ->
+        if (value == null || value is Class<*> || value is ClassLoader) return@processObject null
```
ndkoval, line 53: Is it possible to organize the code flow without labels? The current version is hard to read.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/FieldSearchHelper.kt
```diff
@@ -52,50 +44,47 @@ internal object FieldSearchHelper {
...
+        }
 
-            if (fieldValue === value) {
-                if (fieldName != null) return MultipleFieldsMatching
-                if (!Modifier.isFinal(field.modifiers)) return FoundInNonFinalField
+        traverseObjectGraph(
+            testObject,
+            onArrayElement = { _, _, _ -> null }, // do not traverse array elements further
+            onField = { ownerObject, field, fieldValue ->
+                if (field.type.isPrimitive || fieldValue == null) return@traverseObjectGraph null
```
ndkoval, line 61: Can we avoid using labels, please?

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -39,107 +35,81 @@ import kotlin.coroutines.Continuation
...
+                null
+            }
+        },
+        onField = { _, f, value ->
+            if (
+                Modifier.isStatic(f.modifiers) ||
+                f.isEnumConstant ||
+                f.name == "serialVersionUID"
+            ) null
+            else try {
```
ndkoval, line 105: Please make it look like the code below, with `{ .. }` brackets. Please also check other places. 

```
if (...) {
 ...
} else {
  ...
}
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/FieldSearchHelper.kt
```diff
@@ -52,50 +44,46 @@ internal object FieldSearchHelper {
...
-                    if (fieldName != null) {
-                        return MultipleFieldsMatching
-                    } else {
-                        fieldName = result.field
+        traverseObjectGraph(
+            testObject,
+            onArrayElement = { _, _, _ -> null }, // do not traverse array elements further
+            onField = { ownerObject, field, fieldValue ->
+                if (field.type.isPrimitive || fieldValue == null) null
+                else if (value === fieldValue && !isTraverseCompleted()) {
```
ndkoval, line 62: Can we avoid repeating `else if` here and make the logic flow more straightforward? Please also add comments where necessary.

ndkoval (changes requested):

ndkoval (approved):

ndkoval:
@dmitrii-artuhov, thanks for the refactoring! Please fix the tests, and then the PR will be ready to merge.

Issue #403, opened 2024-09-30
`IllegalStateException`: Trying to switch the execution to thread 0


de-shyt:
```
java.lang.IllegalStateException: Trying to switch the execution to thread 0,
but only the following threads are eligible to switch: [2]
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.chooseThread(ModelCheckingStrategy.kt:306)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.doSwitchCurrentThread(ManagedStrategy.kt:524)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread(ManagedStrategy.kt:500)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread$default(ManagedStrategy.kt:493)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPoint(ManagedStrategy.kt:402)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPointOnAtomicMethodCall(ManagedStrategy.kt:964)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.beforeMethodCall(ManagedStrategy.kt:903)
	at sun.nio.ch.lincheck.Injections.beforeMethodCall(Injections.java:249)
	at com.deshyt.buffered.ChannelSegment.getState$kotlin_channels(ChannelSegment.kt:40)
	at com.deshyt.buffered.BufferedChannel.updateCellOnSend(BufferedChannel.kt:87)
	at com.deshyt.buffered.BufferedChannel.send(BufferedChannel.kt:74)
	at com.deshyt.ChannelTestBase.send(ChannelTestBase.kt:22)
	at org.jetbrains.kotlinx.lincheck.runner.TestThreadExecution1.run(Unknown Source)
	at org.jetbrains.kotlinx.lincheck.runner.FixedActiveThreadsExecutor.testThreadRunnable$lambda$10(FixedActiveThreadsExecutor.kt:172)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

The final part of the stack trace where the unsuccessful attempts to switch are:
```
[...]

PR #421, opened 2024-10-28
Prevent IllegalArgumentException

jbonzohln:
When running my tests I get this error:

```
java.lang.IllegalArgumentException: Count 'n' must be non-negative, but was -5.

	at kotlin.text.StringsKt__StringsJVMKt.repeat(StringsJVM.kt:775)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceNodePrefixFactory.prefixForNode$lambda$5(TraceReporter.kt:628)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceNode.getPrefix(TraceReporter.kt:354)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceLeafEvent.addRepresentationTo(TraceReporter.kt:415)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.traceGraphToRepresentationList(TraceReporter.kt:343)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.appendShortTrace(TraceReporter.kt:43)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.TraceReporterKt.appendTrace(TraceReporter.kt:31)
	at org.jetbrains.kotlinx.lincheck.ReporterKt.appendFailure(Reporter.kt:392)
	at org.jetbrains.kotlinx.lincheck.strategy.LincheckFailure.toString(LincheckFailure.kt:22)
	at java.base/java.lang.StringConcatHelper.stringOf(StringConcatHelper.java:467)
	at java.base/java.lang.StringConcatHelper.simpleConcat(StringConcatHelper.java:422)
	at org.jetbrains.kotlinx.lincheck.LincheckAssertionError.<init>(LincheckAssertionError.kt:17)
	at org.jetbrains.kotlinx.lincheck.LinChecker$check$1.invoke(LinChecker.kt:48)
	at org.jetbrains.kotlinx.lincheck.LinChecker$check$1.invoke(LinChecker.kt:47)
	at org.jetbrains.kotlinx.lincheck.LinChecker.checkImpl$lincheck(LinChecker.kt:67)
	at org.jetbrains.kotlinx.lincheck.LinChecker.check(LinChecker.kt:47)
	at org.jetbrains.kotlinx.lincheck.LinChecker$Companion.check(LinChecker.kt:195)
	at org.jetbrains.kotlinx.lincheck.LinChecker.check(LinChecker.kt)
```

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt
@@ -625,7 +625,7 @@ private class TraceNodePrefixFactory(nThreads: Int) {
             val arrowDepth = arrowDepth
             return PrefixProvider {
                 val extraPrefix = if (arrowDepth == 1) 0 else extraPrefixLength(iThread)
-                TRACE_INDENTATION.repeat(max(0, arrowDepth - 2 + extraPrefix)) + "| " + TRACE_INDENTATION.repeat(callDepth - arrowDepth + 1)
+                TRACE_INDENTATION.repeat(max(0, arrowDepth - 2 + extraPrefix)) + "| " + TRACE_INDENTATION.repeat(max(0, callDepth - arrowDepth + 1))
             }
         }
         return PrefixProvider { extraPrefixIfNeeded(iThread) + TRACE_INDENTATION.repeat(callDepth) }


eupp:
Could be related to #404

avpotapov00:
Hi, @jbonzohln 
Thanks for the PR!
Could you please provide the test that caused the failure? 
Let's fix it a little bit in a different way.

jbonzohln:
@avpotapov00 No, sorry it's company code I can't share. I believe the cause is a very deep call stack though.

ndkoval:
Unfortunately, we cannot accept such a PR without tests.

eupp:
Hi @ndkoval !
Can we please re-open and merge this PR?

I've tested it on #404 (same issue). I've tried to apply this commit on top of `master` (as of Lincheck-2.34), and with this commit the problem is no longer reproducible. 

Also, I've encountered the same bug while working on another feature --- supporting coroutines in GPMC API #411, and with this fix the problem is also fixed there. 

The fix in this PR is very simple, it just prevents the negative value to leak into a function not expecting negative value, by simply taking `max(0, x)`.  

As for the test, the problem is that the code affected by this fix is deep down the internal logic of constructing trace representation. It is hard to construct an artificial test to trigger this behavior (I've tried). 

We can try to port the reproducing test reported by @de-shyt in #404 from her repository. However, this test is quite large and complicated (it includes custom implementation of `Channel` interface). Also, this test is also affected by another issue #403, and when #403 is fixed the #404 no longer reproduces. 
So my point is that trying to adapt this complicated test for this simple one-liner fix probably does not worth the hustle.

ndkoval:
@eupp I've merged the PR but accidentally did that to `master`. Please port the fix to `develop` as well.

PR #422, opened 2024-11-07
Fix local objects detector

eupp:
This PR changes our approach to the local objects tracking. 

Current approach maintains a map `LocalObject -> List<LocalObject>` , which for each local object stores a list of objects immediately reachable from it (via its fields). This map is used to compute a set of local objects reachable from another given local object. Once a local object is assigned to a field of non-local object, all reachable objects are marked as non-local. 

The problem with this approach is that it is error prone and requires a lot of efforts to maintain the map in the correct state. If the map is not maintained in the correct state, some objects might be incorrectly classified as local (see new test added in this PR). 

So instead, in this PR we implement an alternative approach. We still maintain a *set* of local objects, however when we need to compute a set of objects reachable from given one, we instead traverse the object graph using the utility function. 
Thus there is no need to maintain a separate data structure approximating reachable objects.

This PR also should help with #415 , as it would be much more easier to just traverse all objects reachable from the given `Thread` objects once it starts, and mark them as non-local. 

Note this PR depends on PR #420 , so we first need to merge it, and then rebase.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
@@ -101,11 +101,7 @@ private fun enumerateObjects(obj: Any, objectNumberMap: MutableMap<Any, Int>) {
             }
         },
         onField = { _, f, value ->
-            if (
-                Modifier.isStatic(f.modifiers) ||
-                f.isEnumConstant ||
-                f.name == "serialVersionUID"
-            ) {
+            if (f.isEnumConstant || f.name == "serialVersionUID") {
                 null
             }
             else {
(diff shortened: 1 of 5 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/ObjectTraverser.kt
```diff
@@ -100,11 +100,8 @@ private fun enumerateObjects(obj: Any, objectNumberMap: MutableMap<Any, Int>) {
             return@traverseObjectGraph null
         },
         onField = { _, f, value ->
-            if (
-                Modifier.isStatic(f.modifiers) ||
-                f.isEnumConstant ||
-                f.name == "serialVersionUID"
-            ) return@traverseObjectGraph null
+            if (f.isEnumConstant || f.name == "serialVersionUID")
```
ndkoval, line 103: Please always use brackets when the `return` statement is not on the same line

ndkoval (approved):

ndkoval (changes requested):
The PR looks good, please address the only minor comment and rebase when #420 is merged

ndkoval (approved):

PR #427, opened 2024-11-26
Remove `InternalLincheckTestUnexpectedException`

eupp:


ndkoval (approved):

eupp:
TLDR: removed tests were intended to check for cases when exception is thrown from the Lincheck itself (thus called "UnexpectedException"), but because the logic of unexpected exceptions was abandoned some time ago (now almost any exception is considered to be normal actor result), the tests were not achieving their goal and were effectively useless.

PR #428, opened 2024-11-27
Remove atomic setter transformers

eupp:
Remove separate bytecode transformers which tracked methods like `.set`, `.compareAndSet`, etc, of `VarHandle`, `AFU`, and `Unsafe` APIs, to inform the `ObjectTracker` about changes in object graph topology (`registerObjectLink` listener). 

Instead, modify `ManagedStrategy::beforeMethodCall` to parse the method name and arguments to extract relevant information and pass it into `ObjectTracker`. 

Also, it looks like the new approach is more robust and handles more cases, which were not handled previously (e.g. atomic array APIs). As such, there are changes in some `representation` tests --- the reported traces now include more events, which were previously incorrectly classified as local object accesses.


ndkoval:
@eupp, do we have tests checking that the local object detection works correctly with updates via methods like `.set`, `.compareAndSet`, etc, of `VarHandle`, `AFU`, and `Unsafe` APIs

eupp:
> @eupp, do we have tests checking that the local object detection works correctly with updates via methods like .set, .compareAndSet, etc, of VarHandle, AFU, and Unsafe APIs

Apparently, there are tests for AFU, Unsafe, and VarHandle APIs, but it looks like there are no tests for Atomics and Atomic Arrays. I can add them. 

https://github.com/JetBrains/lincheck/blob/c4745c88a8cb5a37bed6a2af6206ed099abf42af/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/transformation/LocalObjectsTests.kt#L139

https://github.com/JetBrains/lincheck/blob/c4745c88a8cb5a37bed6a2af6206ed099abf42af/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/transformation/LocalObjectsTests.kt#L231

https://github.com/JetBrains/lincheck/blob/c4745c88a8cb5a37bed6a2af6206ed099abf42af/src/jvm/test-jdk11/org/jetbrains/kotlinx/lincheck_test/transformation/VarHandleLocalObjectsTest.kt#L29

eupp:
I have added new tests to check that local object detection works correctly with `AtomicReference` and `AtomicReferenceArray` classes. Tests for other APIs (AFU, VarHandle, Unsafe) were already there.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -937,6 +933,12 @@ abstract class ManagedStrategy(
             if (owner == null && atomicMethodDescriptor == null && guarantee == null) { // static method
                 LincheckJavaAgent.ensureClassHierarchyIsTransformed(className.canonicalClassName)
             }
+            if (atomicMethodDescriptor != null && atomicMethodDescriptor.kind.isSetter) {
```
ndkoval, line 936: Should the `isCasSetter` case also be covered here?

ndkoval, line 936: This code becomes more and more complicated. I would suggest splitting the logical blocks of this method with comments explaining what each of these logical blocks does.

ndkoval (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: `isSetter` includes `isCasSetter`:

```kotlin
internal val AtomicMethodKind.isSetter get() = when (this) {
    SET,
    GET_AND_SET,
    COMPARE_AND_SET,
    WEAK_COMPARE_AND_SET,
    COMPARE_AND_EXCHANGE
         -> true
    else -> false
}

internal val AtomicMethodKind.isCasSetter get() = when (this) {
    COMPARE_AND_SET,
    WEAK_COMPARE_AND_SET,
    COMPARE_AND_EXCHANGE
         -> true
    else -> false
}
```

Would you prefer to split the two, so that `isSetter` covers only `SET` and `GET_AND_SET` ?

eupp: Done

ndkoval: Got it, thanks. I think this is fine.

ndkoval (approved):

PR #432, opened 2024-12-06
Collect sources from the bootstrap module

zuevmaxim:


build.gradle.kts
```diff
@@ -116,6 +116,7 @@ tasks {
 
     replace("jvmSourcesJar", Jar::class).run {
         from(sourceSets["main"].allSource)
+        from(project(":bootstrap").file("src"))
```
eupp, line 119: Could we add a comment, explaining for what reason we add `boostrap` sources? 
Is it for the purposes to simplify debugging (i.e. being able to navigate to `bootstrap` sources in IDE)?

ndkoval (approved):

PR #444, opened 2025-01-17
Deterministic System.identityHashCode for Trace Debugger. 

ndkoval:
This is an adapted version of #430.


zhelenskiy (changes requested):
Rebase on top of new `develop`, please.

zhelenskiy (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
```diff
@@ -0,0 +1,112 @@
...
+    @OptIn(ExperimentalStdlibApi::class)
+    fun afterNewTrackedObjectCreation(obj: Any): Id {
+        if (!TRACE_DEBUGGER_MODE) return 0
+        val currentObjectId = getNextObjectId()
+        val initialIdentityHashCode = getInitialIdentityHashCode(
+            objectId = currentObjectId,
+            identityHashCode = System.identityHashCode(obj)
+        )
+        // ATTENTION: bizarre and crazy code below (might not work for all JVM implementations)
+        UnsafeHolder.UNSAFE.putInt(obj, IDENTITY_HASHCODE_OFFSET, initialIdentityHashCode)
```
qwwdfsad, line 54: You need a CAS here, this put races with GC allocation/monitor inflation etc.

Also, you may want to triple-check it with liliput and header shenanigans: https://wiki.openjdk.org/display/lilliput/Compact+Identity+Hashcode

qwwdfsad (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
zhelenskiy: I call identityHashCode in line 51, so by line 54, the hash-code of thee newly created object must be already calculated and stored. So, there would be no need for CAS.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/AbstractLincheckTest.kt
```diff
@@ -67,4 +67,4 @@ abstract class AbstractLincheckTest(
     }
 }
 
-private const val TIMEOUT = 10 * 60_000L // 10 min
\ No newline at end of file
+private const val TIMEOUT = 20 * 60_000L // 10 min
```
ndkoval, line 70: please revert

zhelenskiy: Otherwise, it fails because of time out on one of the tests

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -80,17 +80,19 @@ internal class LincheckClassVisitor(
...
             return mv
         }
-        if (methodName == "<init>") {
+        // Debugger implicitly evaluates toString for variables rendering
+        // We need to disable breakpoints in such a case, as the numeration will break.
+        // Breakpoints are disabled as we do not instrument toString and enter an ignored section,
+        // so there are no beforeEvents inside.
+        if (methodName == "<init>" || ideaPluginEnabled && methodName == "toString" && desc == "()Ljava/lang/String;") {
             mv = ObjectCreationTransformer(fileName, className, methodName, mv.newAdapter())
+            if (isInTraceDebuggerMode) {
```
ndkoval, line 93: Don't you always need to apply `DeterministicInvokeDynamicTransformer`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -148,7 +150,9 @@ internal class LincheckClassVisitor(
         mv = WaitNotifyTransformer(fileName, className, methodName, mv.newAdapter())
         mv = ParkingTransformer(fileName, className, methodName, mv.newAdapter())
         mv = ObjectCreationTransformer(fileName, className, methodName, mv.newAdapter())
-        mv = DeterministicHashCodeTransformer(fileName, className, methodName, mv.newAdapter())
+        if (isInTraceDebuggerMode) {
```
ndkoval, line 153: Don't you always need to apply `DeterministicInvokeDynamicTransformer`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
```diff
@@ -0,0 +1,110 @@
...
+         * @see <a href="https://shipilev.net/jvm/anatomy-quarks/26-identity-hash-code/#_hashcode_storage">JVM Anatomy Quark #26: Identity Hash Code</a>
+         */
+        private const val IDENTITY_HASHCODE_OFFSET: Long = 1L
+    }
+    private val initialHashCodes = ConcurrentHashMap<Id, Int>()
+    private val nextObjectId = AtomicLong(0)
+    private val objectIdAdvances = ConcurrentHashMap<Id, Id>()
+
+    /**
+     * This method substitutes identity hash code of the object in its header with the initial (from the first test execution) identity hashcode.
```
ndkoval, line 42: Now, this technique is applied to both Lincheck and Trace Debugger. Please do a pass over the documentation/comments in this PR and modify them accordingly where necessary.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/AbstractLincheckTest.kt
zhelenskiy: Here [it](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava21TraceDebugger/4969210?buildTab=tests&focusLine=NaN&name=SerializableJavaUtilParameterTest.testWithMod&expandedTest=build:(id:4969210),id:2000000297) is.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
zhelenskiy: Yes

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
zhelenskiy: Yes

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
zhelenskiy: > Now, this technique is applied to both Lincheck and Trace Debugger.
That's false.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
ndkoval: I see. Let's then leave a comment in the code explaining why this transformer is unnecessary.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
ndkoval: I mean, we always put the identity hash code value directly into the object now.

ndkoval: Still, this method injects the identity hash code in both cases.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/AbstractNativeCallTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import org.junit.Assume.assumeTrue
+import org.junit.Test
+import java.io.ByteArrayOutputStream
+import java.io.PrintStream
+
+/**
+ * Checks for absence of non-determinism and absence (or existence) of exceptions.
+ */
+abstract class AbstractNativeCallTest {
```
eupp, line 29: Consider renaming to `AbstractDeterministicCallTest`

eupp (changes requested):

eupp (approved):

eupp:
Merge is blocked without @qwwdfsad approval : )

@qwwdfsad should we wait for review or can we proceed?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectIdentityHashCodeTracker.kt
eupp: As @zhelenskiy mentioned, the hash-code computation is enforced just before this line, so the hash code should already be stored. 

Given that we currently run this tracker under `ManagedStrategy`, which controls thread scheduling, only single thread can have access to this object at the time, so no other thread may try to `synchronize` on this object and set its lock bits. 

So it looks like the only possibility is if the concurrent GC will decide to change some marking bits concurrently. 
I am not sure what exactly can happen in this case. Although the two concurrent writes (our hash code write and GC mark bits write) are to disjoint bits, they are still going to be rounded off to word-sized writes, and thus interfere with each other. Even worse, I am not sure would it be the case that `int`-sized write at offset of 1 byte can lead to word-tearing? 
 
On the other hand, we don't observe the bugs on our CI right now.  
And this is still an early prototype, and we might still decide to choose some alternative implementation strategy for deterministic hash-codes later.

So we left a comment and TODOs to investigate the problem further in the future.

eupp (approved):

PR #454, opened 2025-01-28
Add flag to enable the Trace Debugger mode

zhelenskiy:


build.gradle.kts
```diff
@@ -28,6 +28,22 @@ repositories {
     mavenCentral()
 }
 
+fun SourceDirectorySet.configureTestSources() {
+    srcDir("src/jvm/test-common")
+    val testInTraceDebuggerMode: String by project
+    if (testInTraceDebuggerMode.toBoolean().also(::println)) {
+        srcDir("src/jvm/test-trace-debugger")
+    } else {
+        srcDir("src/jvm/test")
```
ndkoval, line 37: In the Trace Debugger mode, we need to run all the tests that do not rely on the `LoopDetector` and `ObjectTracker` functionality

ndkoval (changes requested):

build.gradle.kts
eupp: I think the problem is that we don't know which tests would work (without `LoopDetector` and `ObjectTracker`), and which would not work. 

The only approach I see is to try run all tests, see what fail, and then disable them, but I guess it could take some time (there are ~400 tests in Lincheck currently).

build.gradle.kts
```diff
@@ -28,6 +28,22 @@ repositories {
     mavenCentral()
 }
 
+fun SourceDirectorySet.configureTestSources() {
+    srcDir("src/jvm/test-common")
```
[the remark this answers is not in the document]
zhelenskiy: No, for Kotlin/Multiplatform the common test suit is usually called commeonTest (aka jvmTest, wasmJsTest, nativeTest, iosTest, androidTest). But if you have some better namebin your mind, I would like to rename.

src/jvm/test-common/org/jetbrains/kotlinx/lincheck_test/util/JdkUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+package org.jetbrains.kotlinx.lincheck_test.util
+
+/**
+ * Indicates whether the current Java Development Kit (JDK) version is JDK 8.
+ *
+ * This property checks the system's Java specification version
+ * and determines if the major version corresponds to '8', signifying JDK 8.
+ */
+// java.specification.version is "1.$x" for Java prior to 8 and "$x" for the newer ones
+internal val isJdk8 = System.getProperty("java.specification.version").removePrefix("1.") == "8"
```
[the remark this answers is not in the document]
zhelenskiy: To make it common.

build.gradle.kts
zhelenskiy: Yes, I agree with Zhenya. I would have done that. But some tests when run with all others cause JVM crash. According to the logs, it happens only on when the test is run with JRE 8 (not depending on JVM target). The failure place seems unrelated to any changes done by me. I suspect some either minimisation or other similar stuff to be connected with the failure. According to my previous experience with native calls the reason was connected with some tests not aimed to be run in the trace debugger mode. Of course, I can spend a week to comment out one test by one and running them all to find the problematic one, but I don't think it's worth it.

build.gradle.kts
```diff
@@ -28,6 +28,22 @@ repositories {
...
+    val testInTraceDebuggerMode: String by project
+    if (testInTraceDebuggerMode.toBoolean().also(::println)) {
+        srcDir("src/jvm/test-trace-debugger")
```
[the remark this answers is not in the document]
zhelenskiy: No because of the reasons stated below.

zhelenskiy:
@ndkoval Why did you disable logging?

build.gradle.kts
ndkoval: We cannot merge a PR that disables ALL the tests.

eupp: But it does not disable all tests. 

When the trace debugger flag is not set, everything runs as before. All our CI configurations run without any changes, all the tests work as before. 

When the flag is set -- we run only tests in `trace_debugger` folder. We can add there new tests which are designed to work in trace debugger mode. On CI, I've already added a bunch of configurations with trace debugger flag enabled, so we will be able to test trace debugger separately. 

The problem with trying to repair all tests at once is that it is complicated task. It is not simple mechanistic task, it requires to review the tests to understand whether they should work in trace debugger mode or not, and perhaps make some adjustments in testing code. 

Moreover, doing this right now also could be wasteful. So imagine we went through all 400 tests and somehow make them work with local object detector and loop detector disabled. Then next week we decide that for trace debugger we also need to disable shared memory tracker. This will invalidate even more tests, and we will need to do this job again, reviewing all 400 tests to see which of them are supposed to work.

PR #461, opened 2025-01-31
First version of loop visualization

lev-serebryakov-jetbrains:
This is first, incomplete, version of loop visualization.

Some constructs are not supported (yet). 

Main unsupported feature is breaking loop and exiting from the middle of the loop. In simple cases such trace will contain last (incomplete) iteration events out of the loop, and in more complex cases loop is not recognized at all and trace will contain "raw" loop events, like labels and jumps.

Some regenerated presentations contain unprocessed "raw" loop events due to problems described above.


ndkoval:
We've decided to proceed on the plugin side

ndkoval:
@lev-serebryakov-jetbrains, please rename the branch to indicate that it should not be deleted

PR #466, opened 2025-02-03
Instantiate `testClass` taking into account its enclosing class

ivandev0:


ndkoval:
@ivandev0, please create a separate PR for the migration to JUnit 5. Also, let's keep using imports with `*` whenever possible.

eupp:
Let's maybe drop JUnit 5 commit from this PR for now?

It looks like migration to JUnit 5 might take some time due to the problem mentioned in #467:
https://github.com/JetBrains/lincheck/pull/467#issuecomment-2635044645

ivandev0:
Unfortunately, the migration is required because I am using `Nested` annotation to be able to run nested test. I didn't find an option to run nested tests with JUnit4.

eupp:
Lincheck does not force you to use JUnit test classes. You can rewrite the test to achieve the same goal:

```
class OuterClass {
    inner class InnerClass {
        @Operation
        fun zero() = 0
    }
}

class InnerClassTest {
    @Test
    fun someTest() {
        StressOptions()
            .invocationsPerIteration(1)
            .iterations(1)
            .check(OuterClass.InnerClass::class)
    }
}
```

BTW, I renamed the test to `InnerClassTest` to better reflect its semantics.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -329,3 +329,20 @@ internal inline fun <R> runOutsideIgnoredSection(descriptor: ThreadDescriptor?,
...
 internal const val LINCHECK_RUNNER_PACKAGE_NAME = "org.jetbrains.kotlinx.lincheck.runner."
+
+internal fun <T> Class<T>.newInstanceRecursive(): T {
+    @Suppress("UNCHECKED_CAST")
+    val constructor = this.declaredConstructors.singleOrNull { it.parameterCount == 0 } as? Constructor<T>
+    if (constructor != null) {
+        return constructor.newInstance()
+    }
+
+    if (this.enclosingClass != null) {
```
eupp, line 340: Shouldn't we check if `enclosingClass != null` as a first step, and only then look for default constructor with zero arguments?

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
```diff
@@ -329,3 +329,20 @@ internal inline fun <R> runOutsideIgnoredSection(descriptor: ThreadDescriptor?,
 
 internal const val LINCHECK_PACKAGE_NAME = "org.jetbrains.kotlinx.lincheck."
 internal const val LINCHECK_RUNNER_PACKAGE_NAME = "org.jetbrains.kotlinx.lincheck.runner."
+
+internal fun <T> Class<T>.newInstanceRecursive(): T {
```
eupp, line 333: The name `newInstanceRecursive` is a bit misleading IMO. 
We just want to create some "default" instance of the class. We do not really care if it is inner or not, it is mostly an implementation detail.  

So maybe rename to something like `newDefaultInstance()` ?

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Utils.kt
ivandev0: Not really. Class can have a default constructor even if `enclosingClass != null`. For example
```
package org.example

class A {
    inner class B
}

class C {
    class D
}

fun main() {
    println(A.B::class.java.enclosingClass) // class org.example.A
    println(C.D::class.java.enclosingClass) // class org.example.C
    // But
    A().B() // requires A object
    C.D()   // doesn't require C object
}
```

I can probably add such a test.

eupp: Ok, I see, thanks. 

Yes, it would be nice to add another test case for nested non-inner class (static nested in Java terminology).

eupp (approved):

PR #467, opened 2025-02-03
Switch to JUnit 5

ivandev0:


eupp:
It seems that the JUnit 5 uses different approach to discover tests, which interferes in a non-obvious way with our class-loading hacks.

When I try to run tests locally from terminal, I got the following error:
```
org.gradle.api.internal.tasks.testing.TestSuiteExecutionException: Could not complete execution for Gradle Test Executor 3.
...
Caused by: java.lang.NoClassDefFoundError: sun/nio/ch/lincheck/EventTracker
```

So the problem seems to be because `sun/nio/ch/lincheck/EventTracker` class cannot be found.
This class belongs to the Lincheck's `bootstrap` package, and should be loaded by the bootstrap class loader.
In Lincheck, the `bootstrap.jar` is added to bootstrap class loader dynamically upon first installation of Lincheck jvm agent:

https://github.com/JetBrains/lincheck/blob/32052dd9b2c41f46c71bde12f489dfab90eb6ebf/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt#L107

Normally, this should occur on the first Lincheck's test run.
And there are no `sun/nio/ch/lincheck/EventTracker` class (or any other class from `bootstrap`) usages outside the Lincheck's tests scope,
so at the point when `sun/nio/ch/lincheck/EventTracker` is first loaded the `bootstrap.jar` should already be attached.

I suspect that JUnit 5 somehow uses reflection to discover tests in a way, that triggers earlier attempt to load `sun/nio/ch/lincheck/EventTracker`, before the Lincheck agent is installed and the `bootstrap.jar` is added to the class path.

Adding the `bootstrap.jar` manually to the tests task seems to fix the problem locally for me.
That is, I've tried adding the following line to `Test` task:
```
jvmArgs("-Xbootclasspath/a:${file("${project(":bootstrap").buildDir}/libs/bootstrap.jar")}")
```

[Full snippet]
```kotlin
tasks.withType<Test> {
    useJUnitPlatform()
    javaLauncher.set(
        javaToolchains.launcherFor {
            val jdkToolchainVersion: String by project
        }
    )
    jvmArgs("-Xbootclasspath/a:${file("${project(":bootstrap").buildDir}/libs/bootstrap.jar")}") // added line
}
```

But I suspect we don't actually want to do this, and we need another solution for the problem.

ivandev0:
We no longer require the switch to JUnit 5. I am going to close this PR but will keep the branch for a while.

PR #471, opened 2025-02-05
Move `ideaPluginEnabled` into the file scope

ivandev0:


eupp:
Adding @zuevmaxim just to re-check if caching this value is ok for the Lincheck/Idea communication protocol to work properly on plugin side.

eupp (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -46,10 +46,12 @@ fun testFailed(
     exceptions: Array<String>
 ) {}
 
+internal val ideaPluginEnabled = ideaPluginEnabled()
```
zuevmaxim, line 49: Please write a comment that by having this property, we limit the number of `ideaPluginEnabled` calls to 1. So, at the plugin side, the callback is called only once.

zuevmaxim (approved):

PR #497, opened 2025-02-06
Update the `IdeaPlugin.kt` API on the Lincheck side for general-purpose model checking

### LINKED ISSUE #474: Update the `IdeaPlugin.kt` API on the Lincheck side for general-purpose model checking

eupp:
Closes #474

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -22,7 +22,7 @@ import org.jetbrains.kotlinx.lincheck.util.ThreadMap
 import org.jetbrains.kotlinx.lincheck.verifier.Verifier
 import java.lang.reflect.Method
 
-const val MINIMAL_PLUGIN_VERSION = "0.2"
+const val MINIMAL_PLUGIN_VERSION = "0.11"
 
 // ============== This methods are used by debugger from IDEA plugin to communicate with Lincheck ============== //
 
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -43,7 +43,7 @@ fun testFailed(
     trace: Array<String>,
     version: String?,
     minimalPluginVersion: String,
-    exceptions: Array<String>
+    exceptions: Array<String>,
 ) {}
 
 /**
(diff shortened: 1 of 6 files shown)


eupp:
@ndkoval please have a look.
Otherwise it is ready to be merged.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -36,14 +39,16 @@ const val MINIMAL_PLUGIN_VERSION = "0.2"
...
  */
 @Suppress("UNUSED_PARAMETER")
 fun testFailed(
     failureType: String,
     trace: Array<String>,
     version: String?,
     minimalPluginVersion: String,
-    exceptions: Array<String>
+    exceptions: Array<String>,
+    isGeneralPurposeModelChecking: Boolean,
```
ndkoval, line 51: NOTE: we will have to replace it with `enum` when adding the Trace Debugger mode

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -112,10 +119,11 @@ fun beforeEvent(eventId: Int, type: String) {
...
 @Suppress("UNUSED_PARAMETER")
 fun visualizeInstance(
-    testInstance: Any,
-    numbersArrayMap: Array<Any>,
+    testInstance: Any?,
+    objectToNumberMap: Array<Any>,
     continuationToLincheckThreadIdMap: Array<Any>,
-    threadToLincheckThreadIdMap: Array<Any>
+    threadToLincheckThreadIdMap: Array<Any>,
+    // TODO: add thread names array parameter
```
ndkoval, line 126: Please create an issue for this

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -128,6 +136,24 @@ fun onThreadSwitchesOrActorFinishes() {}
...
+internal fun runPluginReplay(
+    testCfg: ModelCheckingCTestConfiguration,
+    testClass: Class<*>,
+    scenario: ExecutionScenario,
+    validationFunction: Actor?,
+    stateRepresentationMethod: Method?,
+    invocations: Int,
+    verifier: Verifier
+) {
+    testCfg.createStrategy(testClass, scenario, validationFunction, stateRepresentationMethod).use { replayStrategy ->
```
ndkoval, line 148: Why do we need to re-create the strategy? I'm not sure this approach is good for Trace Debugger.

ndkoval (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
eupp: Because `eventId`-s are only collected in the trace when `ideaPlugin` flag is enabled. 
I think we should just collect them always, it just an integer counter, there should be no overhead if we enable it always. However, `beforeEvent(eventId)` should only be called when the `ideaPlugin` flag is enabled (so no changes here required).

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
eupp: @avpotapov00 should we pass thread names at all? WDYT?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
ndkoval: I don't see why we cannot collect them when collecting the trace without re-creating the strategy.

PR #509, opened 2025-02-10
Ignore flaky coroutines multiple suspensions test

eupp:
Disabled the `CoroutinesMultipleSuspensionsTest.stressTest` test for now, as it is currently flaky on CI. 

Possible reasons could either be:
* The test also requires support of multiple suspension points in the linearizability checker.
* There is some data race in the `ParallelThreadsRunner` class in the code that manages thread suspension statuses.


ndkoval (approved):
Approved, but please raise an issue to investigate the problem later

eupp:
Done, here is the issue: #513

PR #510, opened 2025-02-10
Implement infrastructure to check recorded trace

ivandev0:
This PR contains changes only about infrastructure. There are no new tests, only the example of how to write one. For now with each rerun we get a new trace (if I did everything right). We need to investigate it.


buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.gradle.api.tasks.Copy
+import org.gradle.kotlin.dsl.register
+
+class ProjectToTest(val name: String, val hash: String, val organization: String = "Kotlin")
+
+private val projectsToTest = listOf(
+    ProjectToTest("kotlinx.collections.immutable", "592f05fce02a1ad9e26cc6f3fdb55cdd97910599")
+)
+
+fun Project.registerIntegrationTestsTasks() {
```
eupp, line 23: Consider renaming to `registerIntegrationTestsPrerequisites` so that the name matches with the registered gradle task name.

buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
```diff
@@ -0,0 +1,50 @@
...
+
+            eachFile {
+                val correctPath = listOf("kotlinx.collections.immutable") + relativePath.segments.drop(1)
+                relativePath = RelativePath(file.isFile, *correctPath.toTypedArray())
+            }
+            includeEmptyDirs = false
+        }
+    }
+
+    tasks.register("integrationTestsPrerequisites") {
```
eupp, line 47: Should we rename to `traceDebuggerIntegrationTestsPrerequisites`?

buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
```diff
@@ -0,0 +1,50 @@
...
+        prerequisite.forEach { dependsOn(it) }
+    }
+}
```
eupp, line 50: Also, consider moving this file into `TraceDebuggerTasks.kt` since this functionality is related to the trace debugger specifically.

buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
```diff
@@ -0,0 +1,50 @@
...
+            overwrite(false)
+        }
+
+        tasks.register<Copy>("${projectName}_unzip") {
+            dependsOn(downloadIntegrationTestsDependency)
+            from(zipTree(downloadIntegrationTestsDependency.get().dest))
+            into(layout.buildDirectory.dir("integrationTestProjects"))
+
+            eachFile {
+                val correctPath = listOf("kotlinx.collections.immutable") + relativePath.segments.drop(1)
```
eupp, line 40: Should `listOf(kotlinx.collections.immutable)` be extracted into some variable?
Currently there are two places where the list of libraries is declared, so someone can accidentally forget to update all the necessary places when adding new libraries.

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
```diff
@@ -15,24 +15,51 @@ import org.jetbrains.kotlinx.lincheck.execution.ExecutionScenario
...
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.ensureClassHierarchyIsTransformed
 import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import java.io.File
 import java.lang.reflect.Method
 import java.lang.reflect.Modifier
 import kotlin.reflect.jvm.kotlinFunction
+import kotlin.system.exitProcess
 
-val isInTraceDebuggerMode = System.getProperty("lincheck.traceDebuggerMode", "false").toBoolean()
+private const val traceDebuggerModeProperty = "lincheck.traceDebuggerMode"
```
eupp, line 28: By convention in the Lincheck repo we usually use `CAPITAL_CASE` for constants.

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
```diff
@@ -15,24 +15,51 @@ import org.jetbrains.kotlinx.lincheck.execution.ExecutionScenario
...
 
 internal object TraceDebuggerInjections {
+    @JvmStatic
+    lateinit var classUnderTraceDebugging: String
+
+    @JvmStatic
+    lateinit var methodUnderTraceDebugging: String
+
+    @JvmStatic
+    var dumpTraceIntoFile: String? = null
```
eupp, line 39: maybe just `traceDumpFilename` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/TraceDebuggerAgent.kt
```diff
@@ -25,20 +25,14 @@ import java.security.ProtectionDomain
  * in order to enable trace debugging plugin functionality.
  */
 internal object TraceDebuggerAgent {
-    val classUnderTraceDebugging: String = System.getProperty("traceDebugger.className", "")
-    val methodUnderTraceDebugging: String = System.getProperty("traceDebugger.methodName", "")
```
eupp, line 29: If class and method names are now passed as jvm agent arguments instead of system properties, should it also be addressed on the plugin side?

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
```diff
@@ -63,11 +90,19 @@ internal object TraceDebuggerInjections {
...
         val result = failure!!.results.threadsResults[0][0]
         if (result is ExceptionResult) throw result.throwable
+
+        if (!dumpTraceIntoFile.isNullOrEmpty() && failure.trace != null) {
+            val trace = constructTraceForPlugin(failure, failure.trace)
+            val dumpFile = File(dumpTraceIntoFile!!)
+            dumpFile.parentFile.mkdirs()
+            dumpFile.createNewFile()
+            dumpFile.writeText(trace.joinToString("\n"))
+            exitProcess(0)
```
eupp, line 104: Why `exitProcess(0)` is required here? 
And why it is only called in one branch of the `if` ?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    private fun buildGradleInitScriptToDumpTrace(
+        testClassName: String,
+        testMethodName: String,
+        fileToDump: File
+    ): String {
+        val pathToFatJar = File("build/libs/fat-trace-debugger-SNAPSHOT.jar")
+        return """
+            gradle.taskGraph.whenReady { TaskExecutionGraph taskGraph ->
+                def jvmTasks = taskGraph.allTasks.findAll { task -> task instanceof JavaForkOptions }
```
eupp, line 36: Is it groovy syntax? Why not kotlin syntax?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+            if (it != 0) error("Gradle tests failed with exit code $it")
+        }
+
+        // TODO decide how to test: with gold data or run twice?
+        val goldDataFile = File(getGolderDataPathFor(testClassName, testMethodName))
+        if (goldDataFile.exists()) {
+            Assert.assertEquals(goldDataFile.readText(), tmpFile.readText())
+        } else {
+            copy(tmpFile, goldDataFile)
+            Assert.fail("The gold data file was created. Please rerun the test.")
```
eupp, line 92: Let's fail test in this case. To overwrite the expected test output, let's use a gradle property. 
We already have one used for similar purposes in Lincheck representation tests, so we can re-use it here (it is called `overwriteRepresentationTestsOutput`). 
See the links below:

- https://github.com/JetBrains/lincheck/blob/a758c54e13058899488cb369207d6d24237cb19c/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt#L54
- https://github.com/JetBrains/lincheck/blob/a758c54e13058899488cb369207d6d24237cb19c/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt#L118
- https://github.com/JetBrains/lincheck/blob/a758c54e13058899488cb369207d6d24237cb19c/gradle.properties#L22

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/KotlinImmutableCollectionsIntegrationTest.kt
```diff
@@ -0,0 +1,27 @@
...
+package org.jetbrains.kotlinx.lincheck_test.integration
+
+import org.junit.Test
+
+class KotlinImmutableCollectionsIntegrationTest: AbstractIntegrationTest() {
+    override val projectPath: String = "build/integrationTestProjects/kotlinx.collections.immutable"
+
+    @Test
+    fun `tests contract list GuavaImmutableListTest_list`() {
+//        runGradleTest(
```
eupp, line 20: Please use `@Ignore` instead and TODO comment explaining why the test is disabled for now.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+package org.jetbrains.kotlinx.lincheck_test.integration
+
+import org.junit.Assert
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.util.*
+
+abstract class AbstractIntegrationTest {
```
eupp, line 19:
```suggestion
abstract class AbstractTraceDebuggerIntegrationTest {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/KotlinImmutableCollectionsIntegrationTest.kt
```diff
@@ -0,0 +1,27 @@
...
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.integration
+
+import org.junit.Test
+
+class KotlinImmutableCollectionsIntegrationTest: AbstractIntegrationTest() {
```
eupp, line 15:
```suggestion
class KotlinImmutableCollectionsTraceDebuggerIntegrationTest: AbstractTraceDebuggerIntegrationTest() {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.integration
```
eupp, line 11: Also please move the package to `trace_debugger.integration`.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/TraceDebuggerAgent.kt
ivandev0: Yes. I did this before `develop` branch become default for the trace debugger plugin. I will prepare an MR for the plugin.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
ivandev0: What do you mean? The test will be failed with `Assert.fail`. This branch means that golden test data wasn't found and it will be created automatically. It is not exactly `overwrite`

ivandev0:
Thanks for the review! I will fix everything on Monday.

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
eupp: This is strange. 

> the agent holds the thread for the debugger

Yes, the agent will execute the method under trace debugging in a loop to enable replays, but it should only happen if the special flag is set (which is set by the IDEA debugger). 

Moreover, this loop is run inside this line:
```
val failure = lincheckOptions.checkImpl(instanceClass)
```  

so if the code reached this point, it should have already exited the loop and should proceed normally.

eupp: Could we maybe introduce a separate option for the agent --- the mode that we run the trace debugger in?

```
enum class TraceDebuggerMode { TRACE_COLLECTION, TRACE_DEBUGGING } 
``` 

And then
- use `TRACE_DEBUGGING` by default;
- pass `TRACE_COLLECTION` when running integration tests;
- call `exit(0)` only in `TRACE_COLLECTION` mode.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
eupp: I am also not a Groovy expert, but I guess in this small script replacing `def` with `val` and doing few other things should do the job:

```kotlin
gradle.taskGraph.whenReady { TaskExecutionGraph taskGraph ->
                val jvmTasks = taskGraph.allTasks.filter { task -> task is JavaForkOptions }
                jvmTasks.forEach { task ->
                    task.doFirst {
                        val options = task as JavaForkOptions
                        val jvmArgs = ArrayList<>()
                        jvmArgs.addAll(options.jvmArgs)
                        jvmArgs.addAll([
                            "-Dlincheck.traceDebuggerMode=true", 
                            "-javaagent:${pathToFatJar.absolutePath}=$testClassName,$testMethodName,${fileToDump.absolutePath}"
                        ])
                        options.jvmArgs = jvmArgs
                    }
                }
            }
```

If that does not work, you could also ask LLM to try converting it : )

In general, I think gradle is already complicated on itself, so mixing both Kotlin and Groovy in one project is only makes things worse.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/integration/AbstractIntegrationTest.kt
eupp: Oh, sorry, I missed the `Assert.fail`. 

Yet still, not sure that the golden data should be automatically created in this case. 
IMO, better have one single mechanism to overwrite golden data (e.g., when special flag is set), and in case when there is no file with golden data just output that file was not found.

buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
ivandev0: I renamed `IntegrationTestsTasks` file into `TraceDebuggerIntegrationTestsTasks`. I don't really want to push everything into one big file and it is better to separate test and release stuff.

buildSrc/src/main/kotlin/IntegrationTestsTasks.kt
ivandev0: Yep, just forget to change this list to `projectName`

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
ivandev0: Checked it again and everything works even without `exitProcess`. I wonder why and that's changed...

ivandev0:
All changes are in `fixup` commits.

eupp (approved):

eupp:
Also added @ndkoval to have a final look before we merge.

ndkoval (dismissed):

buildSrc/src/main/kotlin/TraceDebuggerTasks.kt
```diff
@@ -30,6 +30,7 @@ fun Project.registerTraceDebuggerTasks() {
 
     val traceDebuggerFatJar = tasks.register<Jar>("traceDebuggerFatJar") {
         archiveBaseName.set("fat-trace-debugger")
+        archiveVersion.set("SNAPSHOT")
```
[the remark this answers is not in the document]
ivandev0: What do you mean by "default version"? If we leave it as is, then the version will be different for each release. It is hard to work with that because we should manually update this version in the plugin.

src/jvm/test/org/jetbrains/kotlinx/trace_debugger/integration/AbstractTraceDebuggerIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    private fun createInitScriptAsTempFile(content: String): File {
+        val tempFile = File.createTempFile("initScript", ".gradle.kts")
+        tempFile.writeText(content)
+        return tempFile
+    }
+
+    protected fun runGradleTest(
+        testClassName: String,
+        testMethodName: String,
+        vararg gradleCommand: String,
```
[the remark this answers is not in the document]
ivandev0: > Can gradleCommand vary or should it always be jvmTest/test?

Yes, it can change and it can depend on the project's infra that we are testing

> Let's also do not use vararg here, as it makes the test code less intuitive

Do you propose to use an array here? Or do you mean to pass it as comma-separated string and parse later? I would like to keep it flexible to allow running multiple commands

src/jvm/test/org/jetbrains/kotlinx/trace_debugger/integration/KotlinImmutableCollectionsTraceDebuggerIntegrationTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import org.junit.Ignore
+import org.junit.Test
+
+class KotlinImmutableCollectionsTraceDebuggerIntegrationTest: AbstractTraceDebuggerIntegrationTest() {
+    override val projectPath: String = "build/integrationTestProjects/kotlinx.collections.immutable"
+
+    @Ignore
+    @Test
+    fun `tests contract list GuavaImmutableListTest_list`() {
```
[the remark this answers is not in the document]
ivandev0: Will add an example

build.gradle.kts
```diff
@@ -181,7 +181,10 @@ tasks {
         jvmArgs(extraArgs)
     }
 
+    registerIntegrationTestsPrerequisites()
+
     val jvmTest = named<Test>("jvmTest") {
+        dependsOn(traceDebuggerIntegrationTestsPrerequisites)
```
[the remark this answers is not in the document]
ivandev0: Let's do it separately for every trace debugger test. I will create a ticket for that

src/jvm/test/org/jetbrains/kotlinx/trace_debugger/integration/AbstractTraceDebuggerIntegrationTest.kt
```diff
@@ -0,0 +1,104 @@
...
+package org.jetbrains.kotlinx.trace_debugger.integration
+
+import org.jetbrains.kotlinx.lincheck_test.util.OVERWRITE_REPRESENTATION_TESTS_OUTPUT
+import org.junit.Assert
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.util.*
+
+abstract class AbstractTraceDebuggerIntegrationTest {
```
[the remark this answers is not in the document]
ivandev0: We are running these tests in this mode, please check `buildGradleInitScriptToDumpTrace`

buildSrc/src/main/kotlin/TraceDebuggerTasks.kt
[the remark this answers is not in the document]
ndkoval: You can obtain the current version of Lincheck from the code, see https://github.com/JetBrains/lincheck/blob/f2011d4d0c6915cfe6a3a5baf9455e4469e1c012/src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt#L339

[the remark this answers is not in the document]
ivandev0: Yes, but it will not work for the plugin and the plugin has exactly the same logic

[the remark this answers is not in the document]
ndkoval: It's fine to change the `jar` name but please do not change the version.

src/jvm/test/org/jetbrains/kotlinx/trace_debugger/integration/AbstractTraceDebuggerIntegrationTest.kt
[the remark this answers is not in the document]
ndkoval: list?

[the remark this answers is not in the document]
ivandev0: Yeah, if you think this is better, no problem.

buildSrc/src/main/kotlin/TraceDebuggerTasks.kt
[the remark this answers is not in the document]
ivandev0: This one is our internal fat jar. It is not published anywhere (yet). We can push it as is and I will create a task to think about how to extract lincheck version for the plugin.

ndkoval:
@ivandev0 can we please use the same human-readable output format as we use in other trace representation tests?

buildSrc/src/main/kotlin/TraceDebuggerTasks.kt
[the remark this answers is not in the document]
ndkoval: You don't need to extract anything, you can just change the JAR name to "fat-trace-debugger.jar"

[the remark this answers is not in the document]
ndkoval: Please note that it should not be `fat-trace-debugger`, this is just Lincheck

ivandev0:
> @ivandev0 can we please use the same human-readable output format as we use in other trace representation tests?

Can you please point me where to look?

ndkoval:
I guess you need the [`appendTrace` function](https://github.com/JetBrains/lincheck/blob/f2011d4d0c6915cfe6a3a5baf9455e4469e1c012/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TraceReporter.kt#L23), but please take a look at any representation test as well, to check whether something else is necessary.

Issue #505, opened 2025-02-06
Fix bugs in basic thread pool and coroutine tests


diffitask:
Implement simple tests that utilize Kotlin coroutines

PR #525, opened 2025-02-12
Fix instrumentation of newly loaded classes 

eupp:
Before this commit, the instrumentation would only be triggered for re-transformed or re-defined classes. 
With this change, instrumentation is also applied to new classes loaded during Lincheck test execution.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
@@ -80,9 +80,8 @@ object ObjectLabelFactory {
         // There is a Kotlin compiler bug that leads to exception
         // `java.lang.InternalError: Malformed class name`
         // when trying to query for a class name of an anonymous class on JDK 8:
-        // https://youtrack.jetbrains.com/issue/KT-16727/
-        //
-        // in such a case we fall back to returning `<unknown>` class name
+        // - https://youtrack.jetbrains.com/issue/KT-16727/
+        // in such a case we fall back to returning `<unknown>` class name.
         .getOrElse {
             "<unknown>"
         }
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
@@ -69,12 +69,13 @@ object ObjectLabelFactory {
         if (obj is Thread) {
             return "Thread#${getObjectNumber(Thread::class.java, obj)}"
         }
-        val objectName = runCatching {
+        runCatching {
             if (obj.javaClass.isAnonymousClass) {
-                obj.javaClass.simpleNameForAnonymous
-            } else {
-                objectName(obj)
+                return obj.javaClass.simpleNameForAnonymous
             }
+        }
+        val objectName = runCatching {
+            objectName(obj)
         }
         // There is a Kotlin compiler bug that leads to exception
         // `java.lang.InternalError: Malformed class name`
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
@@ -69,10 +69,23 @@ object ObjectLabelFactory {
         if (obj is Thread) {
             return "Thread#${getObjectNumber(Thread::class.java, obj)}"
         }
-        if (obj.javaClass.isAnonymousClass) {
-            return obj.javaClass.simpleNameForAnonymous
+        val objectName = runCatching {
+            if (obj.javaClass.isAnonymousClass) {
+                obj.javaClass.simpleNameForAnonymous
+            } else {
+                objectName(obj)
+            }
         }
-        return objectName(obj) + "#" + getObjectNumber(obj.javaClass, obj)
+        // There is a Kotlin compiler bug that leads to exception
+        // `java.lang.InternalError: Malformed class name`
+        // when trying to query for a class name of an anonymous class on JDK 8:
+        // https://youtrack.jetbrains.com/issue/KT-16727/
+        //
+        // in such a case we fall back to returning `<unknown>` class name
+        .getOrElse {
+            "<unknown>"
+        }
+        return objectName + "#" + getObjectNumber(obj.javaClass, obj)
     }
 
     private fun objectName(obj: Any): String {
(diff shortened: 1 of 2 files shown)


eupp:
This is a part of #505

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -133,8 +133,10 @@ internal object LincheckJavaAgent {
                     // old classes that were already loaded before and have coroutine method calls inside
                     canonicalClassName in coroutineCallingClasses
                 }
-                instrumentation.retransformClasses(*classes.toTypedArray())
-                instrumentedClasses.addAll(classes.map { it.name })
+                if (classes.isNotEmpty()) {
+                    instrumentation.retransformClasses(*classes.toTypedArray())
```
dmitrii-artuhov, line 137: nit: `retransformClasses` on empty array is [guaranteed](https://docs.oracle.com/javase/8/docs/api/java/lang/instrument/Instrumentation.html#:~:text=retransformClasses,-void%20retransformClasses(Class&text=classes)%20throws%20UnmodifiableClassException-,Retransform%20the%20supplied%20set%20of%20classes.,be%20transformed%20with%20the%20ClassFileTransformer%20.) no-op, so this `if` is exhaustive

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -193,7 +195,9 @@ internal object LincheckJavaAgent {
             }
         // `retransformClasses` uses initial (loaded in VM from disk) class bytecode and reapplies
         // transformations of all agents that did not remove their transformers to this moment
-        instrumentation.retransformClasses(*classes.toTypedArray())
+        if (classes.isNotEmpty()) {
+            instrumentation.retransformClasses(*classes.toTypedArray())
```
dmitrii-artuhov, line 199: Same here

dmitrii-artuhov (approved):
Two `nit`s: I would remove the mentioned `if`-statements, but leaving them is also ok

eupp:
> nit: `retransformClasses` on empty array is guaranteed no-op, so this if is exhaustive

I also thought it should be no-op, but unfortunately without the surrounding `if` statement I can observe `NullPointerException` thrown from `install`/`uninstall` on Java 8, leading to failed tests. 

I don't know what exactly the reason, could be some bug in ByteBuddy or Kotlin compiler (e.g., maybe it decides to convert vararg argument with 0 provided arguments to `null` instead of empty array). I did not dig further to be honest, as simply adding the `if` statement seem to solve the problem. 

I've added a comment to the code about this problem.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
```diff
@@ -69,10 +69,23 @@ object ObjectLabelFactory {
...
+        val objectName = runCatching {
+            objectName(obj)
+        }
+        // There is a Kotlin compiler bug that leads to exception
+        // `java.lang.InternalError: Malformed class name`
+        // when trying to query for a class name of an anonymous class on JDK 8:
+        // - https://youtrack.jetbrains.com/issue/KT-16727/
+        // in such a case we fall back to returning `<unknown>` class name.
+        .getOrElse {
+            "<unknown>"
```
ndkoval, line 86: Should we add a representation test for this case?

ndkoval:
@eupp, should a test that verifies the fix be added to this PR?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
eupp: Added a new test for this.

eupp:
> should a test that verifies the fix be added to this PR?

This issue manifests on the coroutines related test that I am going to add under #505.

I've tried to minimize the bug but so far with no success. 
It seems that the problem can only manifest if several non-trivial conditions are satisfied, including some interplay of lazy class transformation logic and ignored sections.

Normally in most cases, the lazy instrumentation would be applied to a class, for example, before the object is created:

```
override fun beforeNewObjectCreation(className: String) = runInIgnoredSection {
    LincheckJavaAgent.ensureClassHierarchyIsTransformed(className)
}

// `ensureClassHierarchyIsTransformed` first calls `Class.forName(canonicalClassName)` to ensure class is loaded
```

So at the point when `LincheckClassFileTransformer::transform` is called, the class is already loaded. 

But if some object is created inside an ignored section, or its creation is untracked by some other reason, the `beforeNewObjectCreation` injection will not be called, and the class will not be instrumented. In such cases, instrumenting the class when it is loaded could be the only opportunity to instrument it.

src/jvm/test/resources/expected_logs/run_concurrent_test/anonymous_object_jdk8.txt
```diff
@@ -0,0 +1,28 @@
...
+| AnonymousObjectRunConcurrentRepresentationTest#1.block(): threw IllegalStateException at BaseRunConcurrentRepresentationTest$testRunWithModelChecker$result$1$1.invoke(RunConcurrentRepresentationTests.kt:40) |
+|   block(): threw IllegalStateException at AnonymousObjectRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:183)                                                                        |
+|     AnonymousObjectRunConcurrentRepresentationTest.runnable = <unknown>#1 at AnonymousObjectRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:204)                                     |
+|     AnonymousObjectRunConcurrentRepresentationTest.runnable ➜ <unknown>#1 at AnonymousObjectRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:205)                                     |
+|     <unknown>#1.run() at AnonymousObjectRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:205)                                                                                         |
```
ndkoval, line 24: If we have two anonymous class instances for different classes, then both representations are likely `<unknown>#1`, with `#1` in both cases. Is it better to fix the counter or simply not show the caller name?

ndkoval: Or let's do nothing :)

ndkoval: Agreed to remove the counter in this case.

ndkoval (approved):

src/jvm/test/resources/expected_logs/run_concurrent_test/anonymous_object_jdk8.txt
eupp: Done

ndkoval (approved):

PR #526, opened 2025-02-13
Supress internal `ThreadAbortedError` in user threads

eupp:


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ThreadTransformer.kt
```diff
@@ -57,6 +66,40 @@ internal class ThreadTransformer(
         visitInsn(opcode)
     }
 
+    override fun visitMaxs(maxStack: Int, maxLocals: Int) = adapter.run {
+        if (isThreadRunMethod(methodName, desc)) {
```
ndkoval, line 70: You can avoid the additional padding by changing the condition to `if (!isThreadRunMethod(..)) { ... }` and calling `return` inside it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ThreadTransformer.kt
```diff
@@ -57,6 +66,40 @@ internal class ThreadTransformer(
...
+            visitLabel(runMethodTryBlockEnd)
+            visitLabel(runMethodCatchBlock)
+
+            // STACK: exception
+            dup()
+            invokeStatic(Injections::onThreadException)
+            // STACK: exception, isSuppressed
+
+            // Notify that the thread has finished.
+            // TODO: currently does not work, because `ManagedStrategy::onThreadFinish`
```
ndkoval, line 80: Is there any plan to fix this? What the absence of the `invokeStatic(Injections::afterThreadFinish)` call might lead to?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -582,46 +601,51 @@ abstract class ManagedStrategy(
...
-     * @param exception the exception that was thrown
+     * This method is executed if an internal exception has been thrown (see [isInternalException]).
+     *
+     * @param threadId the thread id of the thread where exception was thrown.
+     * @param exception the exception that was thrown.
+     * @return true if the exception should be suppressed, false otherwise.
      */
-    open fun onThreadFailure(iThread: Int, exception: Throwable) {
-        // This method is called only if exception can't be treated as a normal operation result,
+    open fun onInternalException(threadId: Int, exception: Throwable): Boolean {
```
ndkoval, line 634: Should the function name be changed to `onThreadRunException`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -582,46 +601,51 @@ abstract class ManagedStrategy(
...
+        // suppress `ThreadAbortedError`
+        if (exception is ThreadAbortedError) return true
+        // Though the corresponding failure will be detected by the runner,
+        // the managed strategy can construct a trace to reproduce this failure.
+        // Let's then store the corresponding failing result and construct the trace.
+        suddenInvocationResult = UnexpectedExceptionInvocationResult(
+            exception, 
+            runner.collectExecutionResults()
+        )
+        threadScheduler.abortAllThreads()
```
ndkoval, line 647: Why should we abort all threads in this case?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -512,6 +511,26 @@ abstract class ManagedStrategy(
...
 
+    /**
+     * Handles exceptions that occur in a specific thread.
+     * This method is called when a thread finishes with an exception.
+     *
+     * @param exception The exception that was thrown within the thread.
+     * @return true if the exception should be suppressed, false otherwise.
+     */
+    override fun onThreadException(exception: Throwable): Boolean = runInIgnoredSection {
+        val currentThreadId = threadScheduler.getCurrentThreadId()
```
ndkoval, line 522: Please add `check(isInternalException(..)) { .. }` to highlight the contract

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ThreadTransformer.kt
eupp: > Is there any plan to fix this?

Yes, I am going to fix it later in another PR, as it is not critical. I've created an issue for now: https://github.com/JetBrains/lincheck/issues/530

> What the absence of the invokeStatic(Injections::afterThreadFinish) call might lead to?

Apparently, there are no serious problems because of that. The only issue is that after termination the thread will end up in `Aborted` state instead of `Finished` state, which leads to few inconveniences in the code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: There are currently two exception-handling injections, because of the slight difference in the data structure test case (i.e., execution scenario) and GPMC case:

- `onThreadException` which is injected into `Thread::run` in case when thread finishes with an exception; 
- `onInternalException` is injected by the `TestThreadExecutionGenerator` into scenario runner in case  when some actor finishes with an exception. 

These two cases have a bit different semantics, because in case of actors we want to:
- re-throw any internal exception (except suppressed `ThreadAbortedError`); 
- treat any other exception as actor result and store it into `ExecutionResults` object. 

In case of `Thread::run` we want only to suppress `ThreadAbortedError` and re-throw any other exception --- this exception will be the final result of the whole `Thread::run` method.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Because we set `suddenInvocationResult` on the previous lines, after this there is no point to continue execution of threads, because `suddenInvocationResult` will be anyway taken as the final result of the whole invocation.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Here I use `if (isInternalException(...))` to delegate to function `onInternalException` (which is also used in actors case). Otherwise the exception is just normal exception of `Thread::run`, and should not be suppressed, so I `return false` in this case.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -582,46 +601,51 @@ abstract class ManagedStrategy(
...
-            // Despite the fact that the corresponding failure will be detected by the runner,
-            // the managed strategy can construct a trace to reproduce this failure.
-            // Let's then store the corresponding failing result and construct the trace.
-            suddenInvocationResult = UnexpectedExceptionInvocationResult(exception, runner.collectExecutionResults())
-            threadScheduler.abortAllThreads()
-        }
-        // notify the scheduler that the thread is going to be finished
-        threadScheduler.finishThread(iThread)
+        // suppress `ThreadAbortedError`
+        if (exception is ThreadAbortedError) return true
```
ndkoval, line 639: Let's rename `ThreadAbortedError` to `LincheckAnalysisAbortedError` or smth similar, to highlight how we use it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ndkoval: Let's then rename `onInternalException` to `onActorFailedWithInternalException` and `onThreadException` to `onThreadRunException`. Please also update the documentation of these methods according to the comment above.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/ThreadTransformer.kt
```diff
@@ -57,6 +66,40 @@ internal class ThreadTransformer(
...
+            // invokeStatic(Injections::afterThreadFinish)
+
+            // Suppress exception if necessary.
+            val suppressLabel = newLabel()
+            // STACK: exception, isSuppressed
+            ifZCmp(NE, suppressLabel)
+
+            // Re-throw exception
+            // STACK: exception
+            visitInsn(ATHROW)
```
ndkoval, line 92: Let's throw the exception in `onThreadException`

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -176,6 +176,22 @@ public static void afterThreadFinish() {
...
+    /**
+     * Called from instrumented code when the thread finishes with an exception.
+     *
+     * @param exception the exception that was thrown in the current thread.
+     * @return true if the exception was handled by the tracker and should be suppressed, false otherwise.
+     */
+    public static boolean onThreadException(Throwable exception) {
+        Thread thread = Thread.currentThread();
+        // TestThread is handled separately
+        if (thread instanceof TestThread) return false;
```
ndkoval, line 188: Should we handle this case separately?

eupp: There are already several places where we handle `TestThread` differently. 
I've created an issue to fix this: #540.

ndkoval (approved):

Issue #452, opened 2025-01-28
Treat first actor (operation) as regular method call


eupp:
On the following test

```
class RecursiveSpinLockRepeatsOperationTest {
    private val value = AtomicBoolean(false)

    fun foo() {
        if (value.get()) return
        foo()
    }

    @Test
    fun test() = ModelCheckingOptions()
        .iterations(0)
        .addCustomScenario { parallel { thread { actor(::foo) } } }
        .checkImpl(this::class.java)
        .checkLincheckOutput("spin_lock/recursive_spin_lock_repeats_operation.txt")
}
```

Lincheck produces an incorrect spin-lock

```
The following interleaving leads to the error:
| ------------------------------------------------------------------------------------------------- |
|                                             Thread 1                                              |
| ------------------------------------------------------------------------------------------------- |
|   foo(): <hung>                                                                                   |
|     /* The following events repeat infinitely: */                                                 |
| ┌╶> value.get(): false at RecursiveSpinLockRepeatsOperationTest.foo(RecursiveSpinLockTest.kt:301) |
| └╶╶ switch (reason: active lock detected)                                                         |
| ------------------------------------------------------------------------------------------------- |
All unfinished threads are in deadlock
```

eupp:
Related to #453

eupp:
The problem is because in the `ManagedStrategy` class we treat the top-most function (the actor) not as a regular method call. We should refactor this to prevent this bug and similar bugs.

eupp:
Migrated to YT

PR #529, opened 2025-02-17
Refactor runner API to avoid creating `testInstance` (and other scenario-specific logic) in case of GPMC

eupp:
This PR attempts to refactor the interaction of the `Strategy` class and the `Runner` class to make the `Strategy` runner-agnostic. The goal is to encapsulate the execution scenario specific logic related to data structures testing. 

In particular, in this PR the `testInstance` object is not used (and thus not re-created) on each invocation when GPMC is used. 

In more details, this PR:
- introduces `ExecutionScenarioRunner` specific to execution scenario running logic;
- introduces `LambdaRunner` specific to GPMC case when only lambda is used to run the test --- this runner avoids most complexities related to `ExecutionScenarioRunner` providing much simpler implementation;
- simplifies common `Runner` interface.

Unfortunately, this PR does not completely decouples `ManagedStrategy` from the execution scenario logic, there are some leftovers that should be handled in the future in separate PRs (as this one is already quite large):
- `currentActorIsBlocking` and `concurrentActorCausesBlocking` properties should probably be moved to `ExecutionScenarioRunner`;
- `scenario` object is used in several places in `ManagedStrategy` (related to https://github.com/JetBrains/lincheck/issues/452#issuecomment-2640621696);
- number of threads in the scenario is used to distinguish between scenario threads and user-created threads --- in the future we should remove any distinction between the two in ` ManagedStrategy`.


src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ActiveThreadPoolExecutor.kt
```diff
@@ -45,83 +46,89 @@ internal class FixedActiveThreadsExecutor(private val testName: String, private
...
-        require(tasks.all { it.iThread in 0 until nThreads}) {
+    fun submitAndAwait(tasks: ThreadMap<Runnable>, timeoutNano: Long): Long {
+        require(tasks.keys.all { it in 0 until nThreads}) {
             "Submitted tasks contain thread index outside of current executor bounds."
         }
-        require(tasks.distinctBy { it.iThread }.size == tasks.size) {
-            "Submitted tasks have duplicate thread indices."
+        check(!isStuck) {
+            "This executor is stuck and cannot accept any new tasks."
         }
```
dmitrii-artuhov, line 100: Why is removing the distinction check valid?

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
```diff
@@ -0,0 +1,98 @@
...
+    private fun collectExecutionResults(wrapper: LambdaWrapper) = ExecutionResult(
+        parallelResultsWithClock = listOf(listOf(
+            ResultWithClock(wrapper.result?.toLinCheckResult() ?: NoResult, emptyClock(1))
+        )),
+        initResults = listOf(),
+        postResults = listOf(),
+        afterInitStateRepresentation = null,
+        afterParallelStateRepresentation = null,
+        afterPostStateRepresentation = null,
+    )
```
dmitrii-artuhov: Maybe add a shortcut function, like it was done for the `emptyExecutionScenario`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -117,11 +115,12 @@ abstract class Strategy protected constructor(
  * @return the failure, if detected, null otherwise.
  */
 fun Strategy.runIteration(invocations: Int, verifier: Verifier): LincheckFailure? {
-    for (invocation in 0 until invocations) {
+    val scenario = (runner as? ExecutionScenarioRunner)?.scenario ?: emptyScenario()
```
dmitrii-artuhov, line 118: I think you don't need to recreate an empty execution scenario on each iteration. If strategy does not use `ExecutionScenarioRunner`, then it does not require a scenario at all, and there is no reason to reinstantiate its object

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -48,29 +47,38 @@ import kotlin.Result as KResult
...
+        get() = (runner as? ExecutionScenarioRunner)?.scenario
 
     // The number of parallel threads.
-    protected val nThreads: Int = scenario.nThreads
+    protected val nScenarioThreads: Int =
+        scenario?.nThreads ?: 0
+
+    // The number of runner threads
+    protected val nRunnerThreads: Int =
+        if (runner is ExecutionScenarioRunner) runner.scenarioThreads.size else 1
```
dmitrii-artuhov, line 72: What is the semantic difference between these two values? The only difference seems to be in the default value, when strategy is not `ExecutionScenarioRunner`, right?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -820,7 +826,7 @@ internal abstract class ManagedStrategy(
      * Checks if [threadId] is a `TestThread` instance (threads created and managed by lincheck itself).
      */
     private fun isTestThread(threadId: Int): Boolean {
-        return threadId in (0 ..< nThreads)
+        return threadId in (0 ..< nRunnerThreads)
```
dmitrii-artuhov, line 838: Why did `isTestThread` chang its semantics? In all places its usage was substituted with comparison `threadId < nScenarioThreads`. What does its implementation return now then?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2297,7 +2322,8 @@ internal abstract class ManagedStrategy(
     }
 
     private fun TraceCollector.addStateRepresentation() {
-        val stateRepresentation = runner.constructStateRepresentation() ?: return
+        val scenarioRunner = (runner as? ExecutionScenarioRunner) ?: return
+            val stateRepresentation = scenarioRunner.constructStateRepresentation() ?: return
```
dmitrii-artuhov: extra indentation here

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -367,7 +367,7 @@ private class Lambda
  * and managing objects and their entries in the registry.
  */
 open class BaseObjectTracker(
-    val executionMode: ExecutionMode
+    val isFirstThreadMain: Boolean,
```
dmitrii-artuhov: Please, add a description to this field. It is non-obvious

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/ActiveThreadPoolExecutor.kt
eupp: Before, tasks were submitted as `Array<out TestThreadExecution>` where each `TestThreadExecution` had `iThread` property telling on what thread it should run. 

After refactoring, I made `ActiveThreadPoolExecutor` more generic, and now it just takes tasks as `Runnable`. 
The tasks are passed as `ThreadMap<Runnable>`, that is mapping `threadId -> runnable`. 
Because it is a map, naturally, there are no `threadId` duplicates.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
eupp: Done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
eupp: The fact that we need `scenario` here is more of a consequence that we still need to use `Verifier.verify` here, that expects `scenario` as an argument. It would be good to refactor it too, but this PR is already quite large. 

In the meantime, I don't think that re-creating empty scenario on each iteration would induce any significant overhead here.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: * `nScenarioThreads` --- is the number of parallel threads in a scenario, it is only relevant in case of data structures testing. 
* `nRunnerThreads` --- is the number of threads managed by the runner (= 1 in case of GPMC, = `nScenarioThreads` in case of data structures testing).

eupp: I've rewritten the comments to better explain the difference.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: I removed `isTestThread(...)` and instead introduced `isRunnerThread(...)` and `isScenarioThread(...)`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
eupp: I refactored thread numeration logic to avoid `isFirstThreadMain` field at all.

PR #533, opened 2025-02-18
Pass location information of each trace point to IDEA debugger.

lev-serebryakov-jetbrains:
Add date needed for loop visualization in IDEA plugin:

1. Stack trace element data (class, file, line)
2. Internal code locations, which means nothing outside `lincheck`, but have very important property: they are strictly incrementing. It allows to use them to distinguish trace points (and what goes first!) which reside on same line.

This PR change all trace points constructors to hold location Id and not stack trace element, as all of them need location Id now and stack trace element can be derived from location Id.

Please note, that this review contains change of version to make it custom to be able to reference it in plugin. This change should not be merged, of course.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -173,7 +173,12 @@ internal fun ManagedStrategy.runReplayIfPluginEnabled(failure: LincheckFailure) 
  * (due to difficulties with passing objects like List and TracePoint, as class versions may vary)
  *
  * Each trace point is transformed into the line of the following form:
- * `type,iThread,callDepth,shouldBeExpanded,eventId,representation`.
+ * `type;iThread;callDepth;shouldBeExpanded;eventId;representation;stackTraceElement;codeLocationId`.
+ *
+ *   stackTraceElement is "className:methodName:fileName:lineNumber" or "null" string if it is not applicable
+ *   codeLocationId is strictly growing abstract id of location, and it must grow in syntactic order to
+ *                  be able to order events occurred at same line in the same file. It is `-1` if it is not
+ *                  applicable and stackTranceElement is "null".
  *
  * Later, when [testFailed] breakpoint is triggered debugger parses these lines back to trace points.
  *
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -203,6 +203,13 @@ private fun constructTraceForPlugin(failure: LincheckFailure, trace: Trace): Arr
                 val event = node.event
                 val eventId = event.eventId
                 val representation = event.toStringImpl(withLocation = false)
+                val (location, locationId) = if (event is CodeLocationTracePoint) {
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -10,7 +10,7 @@
 
 group=org.jetbrains.kotlinx
 name=lincheck
-version=2.36-SNAPSHOT
+version=2.36-SNAPSHOT-LEVSER
```
ndkoval, line 13: Please revert

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -210,30 +217,33 @@ private fun constructTraceForPlugin(failure: LincheckFailure, trace: Trace): Arr
                 }
 
                 if (representation.isNotEmpty()) {
-                    representations.add("$type;${node.iThread};${node.callDepth};${node.shouldBeExpanded(false)};${eventId};${representation}")
+                    representations.add("$type;${node.iThread};${node.callDepth};${node.shouldBeExpanded(false)};${eventId};${representation};${location};${locationId}")
```
ndkoval, line 220: Please update the documentation indicating the ådded parameters

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -195,6 +195,13 @@ private fun constructTraceForPlugin(failure: LincheckFailure, trace: Trace): Arr
                 val event = node.event
                 val eventId = event.eventId
                 val representation = event.toStringImpl(withLocation = false)
+                val (location, locationId) = if (event is CodeLocationTracePoint) {
+                    val ste = event.stackTraceElement
+                    Pair("${ste.className}:${ste.methodName}:${ste.fileName}:${ste.lineNumber}", event.codeLocation)
+                }
+                else {
+                    Pair("null", -1)
```
ndkoval, line 203: `null to -1`?

eupp (changes requested):
Please revert the version name, repair the tests, and fix other comments. 
Otherwise looks good to me.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
lev-serebryakov-jetbrains: I don't understand this comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
lev-serebryakov-jetbrains: Done

gradle.properties
lev-serebryakov-jetbrains: Fixed

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
ndkoval: You can use `"null" to -1` instead

lev-serebryakov-jetbrains: Oh, I see :) 
You about syntax sugar!

eupp (approved):

Issue #512, opened 2025-02-10
Improve representation of Kotlin `thread` creation function


eupp:
Currently, the call to Kotlin `thread { ... }` function looks as follows in the trace:
```
ThreadsKt.thread$default(false, false, null, null, 0, t$1, 31, null): Thread#2 at KotlinThreadRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:241)
```

We need to shorten this.

PR #534, opened 2025-02-19
Improve thread creation representation #512

### LINKED ISSUE #521: Improve representation of Kotlin's `suspend` function calls
* When calling a `suspend` function, optionally indicate that this is a `suspend` function and hide the `Continuation` parameter.
* When a coroutine (`Continuation`) is resumed, hide the machinery to recover the execution stack.

bbrockbernd:
Closes #521 

Thread creation will be represented as:
```
thread(name = "threadname") { ... } : Thread#N at Class.method(File.kt:line)
```


src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
```diff
@@ -0,0 +1,65 @@
...
+|     Thread#4.start() at ThreadsKt.thread(Thread.kt:42)                                                                                            |                                                                                                                                                                                            |                                                                                                                                                                                            |                                                                                                                                                                                            |
```
ndkoval, line 35: Let's remove `Thread.start()` calls when Kotlin's `thread` is called -- we do not need to show the standard library implementation details

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
```diff
@@ -0,0 +1,65 @@
...
+|                                                                                                                                                   | run()                                                                                                                                                                                      |                                                                                                                                                                                            |                                                                                                                                                                                            |
```
ndkoval, line 45: Can we raise an issue (and solve it) to get rid of these `run()` calls and the lambda invocation calls below?

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
```diff
@@ -0,0 +1,65 @@
...
+|                                                                                                                                                   |                                                                                                                                                                                            |                                                                                                                                                                                            | run()                                                                                                                                                                                      |
```
ndkoval, line 53: Let's raise an issue to show `result: ...` only for actor calls (and solve it) -- we do not want this garbage in the trace :)

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
```diff
@@ -0,0 +1,65 @@
...
+|                                                                                                                                                   |       ThreadCreationRepresentationTest.access$callMe(ThreadCreationRepresentationTest#1) at ThreadCreationRepresentationTest$operation$t1$1.invoke(ThreadCreationRepresentationTest.kt:21) |                                                                                                                                                                                            |                                                                                                                                                                                            |
```
ndkoval, line 48: this one we should get rid of as well

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -237,6 +265,16 @@ internal class MethodCallTracePoint(
     }
 }
 
+private fun List<String>.checkDefault(paramNames: List<String>, defaultValues: List<String>): List<String> {
```
eupp, line 268: Let rename to `getNamedArguments` or something similar. 
Also please add kdoc explaining the function semantics, its parameters, and return value.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -182,28 +193,45 @@ internal class MethodCallTracePoint(
...
-        append(")")
-        val returnedValue = returnedValue
-        if (returnedValue is ReturnedValueResult.ValueResult) {
-            append(": ${returnedValue.valueRepresentation}")
-        } else if (returnedValue is ReturnedValueResult.CoroutineSuspended) {
-            append(": COROUTINE_SUSPENDED")
-        } else if (thrownException != null && thrownException != ThreadAbortedError) {
-            append(": threw ${thrownException!!.javaClass.simpleName}")
+        // If thread creation site
+        if (descriptor == THREAD_FUN_DESCRIPTOR && className == THREAD_CLASS_NAME) {
```
eupp, line 204: Please extract this check into separate function: `isKotlinThreadMethod(className, methodName)`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -182,28 +193,45 @@ internal class MethodCallTracePoint(
...
+            append("thread")
+            val params = parameters?.checkDefault(paramNames, threadDefaultValues) ?: emptyList()
+            if (!params.isEmpty()) {
+                append("(")
+                append(params.joinToString(", "))
+                append(")")
+            }
+            append(" { ... } ")
```
eupp, line 212: Please extract `thread` function printing logic into separate function. 
Then also replace `if` with `when` (as we might need to add more special cases in the future).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -237,6 +265,16 @@ internal class MethodCallTracePoint(
     }
 }
 
+private fun List<String>.checkDefault(paramNames: List<String>, defaultValues: List<String>): List<String> {
+    check(this.size == paramNames.size)
+    check(paramNames.size == defaultValues.size)
+    val result = mutableListOf<String>()
+    dropLast(1).forEachIndexed { index, currentValue ->
```
eupp, line 272: why `dropLast` ?

src/jvm/test/resources/expected_logs/run_concurrent_test/kotlin_thread_jdk8.txt
```diff
@@ -25,7 +25,7 @@ Detailed trace:
...
 |   block(): threw IllegalStateException at KotlinThreadRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:231)                                                                        |                                                                                                                                                                      |
-|     ThreadsKt.thread(true, false, null, null, -1, block$t$1#1): Thread#2 at KotlinThreadRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:241)                                      |                                                                                                                                                                      |
+|     thread { ... } : Thread#2 at KotlinThreadRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:241)                                                                                 |                                                                                                                                                                      |
```
eupp, line 28: For now, let's display it as `thread(block$t$1#1)`. 
We will decide how to improve lambdas representation in general later under #519 .

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -73,7 +83,7 @@ internal class SwitchEventTracePoint(
 internal abstract class CodeLocationTracePoint(
     iThread: Int, actorId: Int,
     callStackTrace: CallStackTrace,
-    val stackTraceElement: StackTraceElement
+    var stackTraceElement: StackTraceElement
```
eupp, line 86: Why do you need to make it mutable?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/ThreadCreationRepresentationTest.kt
```diff
@@ -0,0 +1,36 @@
...
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.representation
+
+import kotlin.concurrent.thread
+
+class ThreadCreationRepresentationTest: BaseTraceRepresentationTest("thread_creation_representation_test.txt") {
```
eupp, line 15: Please use `BaseRunConcurrentRepresentationTest` as a base for new tests. 
It runs the code in GPMC mode and does not involve Lincheck's data structures testing functionality.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -24,6 +29,11 @@ data class Trace(
     }
 }
 
+private val threadDefaultValues = listOf("true", "false", "null", "null", "-1", "")
+private const val THREAD_CLASS_NAME = "kotlin/concurrent/ThreadsKt"
```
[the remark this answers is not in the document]
bbrockbernd: This is the format provided by `MethodCallTracePoint.className`

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -66,7 +66,7 @@ boolean beforeWriteField(Object obj, String className, String fieldName, Object
     boolean beforeWriteArrayElement(Object array, int index, Object value, int codeLocation);
     void afterWrite();
 
-    void beforeMethodCall(Object owner, String className, String methodName, int codeLocation, int methodId, Object[] params);
+    void beforeMethodCall(Object owner, String className, String methodName, int codeLocation, int methodId, String desc, Object[] params);
```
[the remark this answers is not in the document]
bbrockbernd: I used this to find the parameter names, through reflection. Since `MethodCallTracePoint` doesn't have enough information to uniquely identify a method. (parameters is just the values in string repr). We could ofcourse also extract the parameter types and construct a signature from there. But I would argue that is out of scope for this issue. 

So will hardcode the parameter names, as we have also already hardcoded the default values.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -24,6 +29,11 @@ data class Trace(
...
+private const val THREAD_FUN_DESCRIPTOR = "(ZZLjava/lang/ClassLoader;Ljava/lang/String;ILkotlin/jvm/functions/Function0;)Ljava/lang/Thread;"
```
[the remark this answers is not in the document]
bbrockbernd: This might work for this function, but if we are gonna need to identify more functions in the future I think it would be good to think of a more robust method than just matching on "name"

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
bbrockbernd: #536

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
bbrockbernd: #536

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
bbrockbernd: is part of #535

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -184,14 +191,35 @@ internal class MethodCallTracePoint(
...
+    }.toString()
+    
+    private fun StringBuilder.appendThreadCreation() {
+        append("thread")
+        val params = parameters?.let { getNonDefaultParametersWithName(threadParamNames, threadDefaultValues, it) } 
+            ?: emptyList()
+        if (!params.isEmpty()) {
+            append("(")
+            append(params.joinToString(", "))
+            append(")")
```
eupp, line 208: `joinToString` already can handle such situations:

```kotlin
append(params.joinToString(", ", prefix = "(", postfix = ")"))
```

https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/join-to-string.html

Alternatively, you can use string template:
```kotlin
append("(${params.joinToString(", ")})")
```

Please also fix this in `appendDefaultMethodCall`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -24,6 +25,12 @@ data class Trace(
     }
 }
 
+private val threadDefaultValues = listOf("true", "false", "null", "null", "-1", "")
+private val threadParamNames = listOf("start", "isDaemon", "contextClassLoader", "name", "priority", "block")
+private const val THREAD_CLASS_NAME = "kotlin.concurrent.ThreadsKt"
+internal const val THREAD_FUN_NAME = "thread"
```
eupp, line 31: Let's group all this information into one class, something like this:

```kotlin
private class FunctionInfo(
    val className: String, 
    val methodName: String, 
    val parameterNames: List<String>, 
    val defaultValues: List<String>,
)
```
 
and then add: 
```kotlin
val threadFunctionInfo = FunctionInfo(...)
```

The name `FunctionInfo` is debatable, if you have better name in mind please use it.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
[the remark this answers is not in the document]
eupp: > We could ofcourse also extract the parameter types and construct a signature from there. But I would argue that is out of scope for this issue.

Agree its out of scope. Yet we still can do this without passing descriptor argument, just by looking at parameters array somewhere in `ManagedStrategy` class (at that point the parameters are not yet converted to string representation).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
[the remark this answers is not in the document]
eupp: Yes, in future we can also additionally check by signature if we will need it. Yet we still shoudn't use ASM descriptors here, and we should instead derive argument types from `param` array passed to `ManagedStrategy` listener.

src/jvm/test/resources/expected_logs/thread_creation_representation_test.txt
bbrockbernd: And #537

eupp (approved):

Issue #434, opened 2024-12-06
Set-up logging infrastructure


eupp:
As the Lincheck codebase grows in size and complexity, debugging becomes more and more challenging process. 
To improve the development process and speed-up debugging process, we need to add a logging facilities to Lincheck. 

- [ ] Decide whether to use existing library, or implement simple logging facilities from scratch.

- [ ] Select a set of subsystems that can benefit from logging, and add logging there gradually in several steps. 
  - [ ] `Runner` and `FixedActiveThreadsExecutor` classes: log tasks execution process, thrown exceptions, etc. 
  - [ ] `ManagedStrategy`: log calls to injected methods, such as `beforeReadField`, `beforeWriteField`, etc. 
  - [ ] `LincheckJavaAgent`: log what classes are transformed.
      - [ ] add possibility to dump transformed bytecode into files for inspection. 
  - [ ] TODO: what else?

eupp:
Logging infra was set-up in #546. Tickets for adding logging to specific components of the framework can be created later on-demand, or just resolved without dedicated tickets.

PR #546, opened 2025-02-24
Add minimalistic logger to Lincheck

dmitrii-artuhov:
Partially addresses: #434 
* Simple logger added
* Transformed classes are printed to log file
* How it works:
   * One can specify (dev: in `gradle.properties`, lib users in VM arguments) flag `lincheck.log.file` to print log to file (default is `System.err`) 
   * Specify flag `lincheck.log.level` to setup logging level (default is a logging level defined in lincheck source code)

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
@@ -21,7 +21,7 @@ import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransforme
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.INSTRUMENT_ALL_CLASSES
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentationMode
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentedClasses
-import org.jetbrains.kotlinx.lincheck.util.LincheckLogger
+import org.jetbrains.kotlinx.lincheck.util.Logger
 import org.jetbrains.kotlinx.lincheck.util.readFieldViaUnsafe
 import org.objectweb.asm.ClassReader
 import org.objectweb.asm.ClassWriter
@@ -389,7 +389,7 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
         internalClassName: String,
         classBytes: ByteArray
     ): ByteArray = transformedClassesCache.computeIfAbsent(internalClassName.canonicalClassName) {
-        LincheckLogger.log("Transform class '${internalClassName.canonicalClassName}'")
+        Logger.debug { "Transforming $internalClassName" }
 
         val reader = ClassReader(classBytes)
         val writer = SafeClassWriter(reader, loader, ClassWriter.COMPUTE_FRAMES)
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
@@ -21,6 +21,7 @@ import org.jetbrains.kotlinx.lincheck.transformation.LincheckClassFileTransforme
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.INSTRUMENT_ALL_CLASSES
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentationMode
 import org.jetbrains.kotlinx.lincheck.transformation.LincheckJavaAgent.instrumentedClasses
+import org.jetbrains.kotlinx.lincheck.util.LincheckLogger
 import org.jetbrains.kotlinx.lincheck.util.readFieldViaUnsafe
 import org.objectweb.asm.ClassReader
 import org.objectweb.asm.ClassWriter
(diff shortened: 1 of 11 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,91 @@
...
+            initFile(it)
+            logWriter = BufferedWriter(FileWriter(it))
+            println("Lincheck log file could be found at: ${it.absolutePath}")
+        }
+    }
+
+    /**
+     * Logs will be printed to log file if it exists.
+     */
+    fun log(s: String) {
```
eupp, line 41: Please implement the logging as follows:

```suggestion
    inline fun log(logLevel: LogLevel, lazyMessage: () -> String) { 
        if (logLevel <= this.logLevel) {
            write(message)
        }
    }
```

By making the function `inline` and accepting the lambda to produce the message, you avoid actually allocating memory and creating `String` object in case when the log message will not be printed because of the insufficient logging level:

https://kotlinlang.org/docs/inline-functions.html

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,91 @@
...
+        logFile?.let {
+            write(s, logWriter!!)
+        }
+    }
+
+    /**
+     * Errors will be printed to log file if it is set.
+     * However, they will always be printed to the `System.err`.
+     */
+    fun error(m: String, e: Throwable) {
```
eupp, line 51: Let's create logging methods for each logging level:

```kotlin
fun error(lazyMessage: () -> String)
fun warning(lazyMessage: () -> String)
fun info(lazyMessage: () -> String)
fun debug(lazyMessage: () -> String)
```

Then for `error` and `warning` you can also versions that take exception as an argument:
```kotlin
fun error(throwable: Threowable) {
    error { throwable.toString() }
}
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,91 @@
...
+    private val formatter: ThreadLocal<SimpleDateFormat> = object : ThreadLocal<SimpleDateFormat>() {
+        override fun initialValue(): SimpleDateFormat {
+            return SimpleDateFormat("dd.MM.yyyy HH:mm:ss")
+        }
+    }
+
+    private val logFilename = System.getProperty("lincheck.log.file") ?: null
+    private val logFile = if (logFilename != null) File(logFilename) else null
+    private var logWriter: Writer? = null
+
```
eupp, line 28: Let's also re-use `LogLevel` enum from `Reporter.kt`  (and let's move it into this file), and use it as a parameter to set logging level.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -388,6 +389,8 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
         internalClassName: String,
         classBytes: ByteArray
     ): ByteArray = transformedClassesCache.computeIfAbsent(internalClassName.canonicalClassName) {
+        LincheckLogger.log("Transform class '${internalClassName.canonicalClassName}'")
```
eupp, line 392: Let's use `debug` level for instrumentation logging.

eupp: Also no need to convert to `canonicalClassName` here, lets print class name as is.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
+
+import java.io.*
+import java.text.SimpleDateFormat
+import java.util.*
+
+object LincheckLogger {
+    private val LINE_SEPARATOR = System.lineSeparator()
```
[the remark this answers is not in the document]
dmitrii-artuhov: `System.lineSeparator` property is determined at runtime, I cannot make it `const`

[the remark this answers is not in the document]
eupp: Ok, at least let's put it at the end of file, in the file scope (not `Logger` scope).

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    private val logFilename = System.getProperty("lincheck.log.file") ?: null
+    private val logFile = if (logFilename != null) File(logFilename) else null
+    private var logWriter: Writer? = null
+
+    init {
+        println("property: lincheck.log.file = " + System.getProperty("lincheck.log.file"))
+        logFile?.let {
```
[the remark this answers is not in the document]
dmitrii-artuhov: removed

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Logger.kt
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
+
+import java.io.*
+import java.text.SimpleDateFormat
```
eupp, line 14: unused import?

eupp (approved):

PR #550, opened 2025-02-26
Automate representation test overwrite

bbrockbernd:
This PR fixes two problems when working on trace representation.

- It provides a script that runs all tests on all relevant jdks on all modes (trace debugger and non trace debugger)
- It auto generates txt output files for new tests, only for jdks and modes where the outputs differ (to prevent duplication)
- Since it generates the txt files necessary, we only provide a `expectedOutputFilePrefix`. The result of this is that a lot of test files changed.
- Since most test files changed (in line numbers) the output txts changed as well.
- Additionally, since it chooses when a seperate txt file for a test is needed, some txts are gone and others have been added.

How it works:
The test outputs of the run with jdk 8 and non trace mode, are considered as the base output and will therefore always (over)write the txt. If a test is run with a higher jdk and/or in trace mode. We will look if a txt for that combo exists and if not we will look at lower jdks (and at the non-trace side). We will then compare the output to the found file. If it differs we (over)write the file with current jdk and mode.


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -87,6 +138,13 @@ internal fun getExpectedLogFileFromResources(fileName: String): File =
     ClassLoader.getSystemResource("expected_logs/$fileName")?.file?.let { File(it) }
         ?: error("Expected log resource $fileName does not exist")
 
+// Returns true if file exists.
+internal fun logFileFromResourcesExists(fileName: String): Boolean =
+    try { 
+        getExpectedLogFileFromResources(fileName) 
+        true
+    } catch (_: IllegalStateException) { false }
```
eupp, line 146: Better use `runCatching` here instead:
https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/run-catching.html

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -22,43 +22,94 @@ import java.io.File
...
+        if (logFileFromResourcesExists(fileName)) return fileName
+    }
+    error("No file exists yet for this test and current jdk = $testJdkVersion, please run on jdk 8 with overwrite enabled")
+}
+
+// Generates file name in the form of 
+// prefix.txt, prefix_jdk_15.txt, prefix_trace_debugger.txt, prefix_trace_debugger_jdk_15.txt, etc..
+private fun getFileNameFor(expectedOutputFilePrefix: String, jdkVersion: TestJdkVersion, traceMode: Boolean): String {
+    return "${expectedOutputFilePrefix}${if (traceMode) "_trace_debugger" else ""}${
+        if (jdkVersion == TestJdkVersion.JDK_8) "" else "_${jdkVersion.name.lowercase()}"
```
eupp, line 109: Instead of `${jdkVersion.name.lowercase()}` could you please override `toString()` of `TestJdkVersion`?
Also, please keep `_jdkN` format (without underscore).

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -22,43 +22,94 @@ import java.io.File
  * This happens when the configuration of the test is incorrect.
  * @param expectedOutputFile name of file stored in resources/expected_logs, storing the expected lincheck output.
  */
-internal inline fun <reified E: Exception> Options<*, *>.checkFailsWithException(testClass: Class<*>, expectedOutputFile: String) {
+internal inline fun <reified E: Exception> Options<*, *>.checkFailsWithException(testClass: Class<*>, expectedOutputFilePrefix: String) {
     try {
         LinChecker(testClass, this).check()
     } catch (e: Exception) {
+        check(!expectedOutputFilePrefix.contains(".txt")) { "should only be prefix of outputfile" }
```
eupp, line 29:
```suggestion
        check(!expectedOutputFilePrefix.contains(".txt")) { 
             "Filename $expectedOutputFilePrefix$ should not contain file extension .txt" 
        }
```

representationtests.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# Copyright (C) 2019 - 2025 JetBrains s.r.o.
+#
+# This Source Code Form is subject to the terms of the
+# Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+# with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+# jdk 11 is SLOW! So remove it if you don't expect files to change
+jdks=("8" "11" "13" "15" "17")
```
eupp, line 12: Why there are no 19, 20 and 21 versions? 
We also test on these versions on CI.

representationtests.sh
```diff
@@ -0,0 +1,27 @@
+#
+# Lincheck
+#
+# Copyright (C) 2019 - 2025 JetBrains s.r.o.
+#
+# This Source Code Form is subject to the terms of the
+# Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+# with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
```
eupp, line 10: Please document what this script does.

representationtests.sh
```diff
@@ -0,0 +1,27 @@
...
+  
+  
+  #https://github.com/JetBrains/lincheck/issues/500
+  if [ "$jdk" = "8" ]; then
+     continue
+  fi 
+  
+  echo "[Representation Tests Overwrite] Running tests for jdk: $jdk in trace mode  --------------------------"
+  ./gradlew jvmTest --tests "$testFilter" -PjdkToolchainVersion="$jdk" -PoverwriteRepresentationTestsOutput=true -PtestInTraceDebuggerMode=true
+done
```
eupp, line 27: Can we create a top-level folder `scripts` and put this script there?

Also, please make the name of the script more obvious and use snake_case: `overwrite_representation_tests.sh`.

eupp (changes requested):

eupp (approved):

PR #559, opened 2025-03-06
Set JDK 17 as a default JDK for representation tests

eupp:


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -55,53 +54,88 @@ private fun compareAndOverwrite(expectedOutputFilePrefix: String, actualOutput:
...
-    }
-    // Check lower sdks in non-trace mode
-    for (i in testJdkVersion.ordinal downTo 0) {
-        val jdkVersion = TestJdkVersion.entries[i]
-        val fileName = getFileNameFor(expectedOutputFilePrefix, jdkVersion, false)
-        if (logFileFromResourcesExists(fileName)) return fileName
+        val traceDebuggerFileName = generateExpectedLogFileName(
+            fileNamePrefix = expectedOutputFilePrefix,
+            jdkVersion = DEFAULT_TEST_JDK_VERSION,
+            inTraceDebuggerMode = true,
```
bbrockbernd, line 98: Wouldn't it make more sense to compare to `jdkVersion = testJdkVersion` and `inTraceDebuggerMode = false` aka compare trace tot non-trace of same jdk.

Often the output is the same for trace and non-trace modes, but when there are a lot of different outputs on the jdk axis (like the `thread_pool_jdkN.txt` files) we would have a lot of duplicates the for the trace mode.

scripts/overwrite_representation_tests.sh
```diff
@@ -22,14 +22,13 @@ testFilter="org.jetbrains.kotlinx.lincheck_test.representation.*"
 for jdk in "${jdks[@]}"
```
bbrockbernd, line 22: If jdk 17 is default we should put it first in the `jdks` list. Otherwise we might end up with unnecessary duplicates.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -55,53 +54,88 @@ private fun compareAndOverwrite(expectedOutputFilePrefix: String, actualOutput:
     check(!expectedOutputFilePrefix.contains(".txt")) {
         "Filename $expectedOutputFilePrefix should not contain a file extension (.txt)"
     }
-    // Always overwrite jdk8 non-trace
-    if (testJdkVersion == TestJdkVersion.JDK_8 && !isInTraceDebuggerMode && OVERWRITE_REPRESENTATION_TESTS_OUTPUT) {
-        getExpectedLogFileFromSources(getFileNameFor(expectedOutputFilePrefix, TestJdkVersion.JDK_8, false)).writeText(actualOutput)
```
bbrockbernd, line 60: When running the test on the default jdk, it would be nice if it created a file if it was not found. Might be cleaner to do that in `getExpectedLogFile`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -55,53 +54,88 @@ private fun compareAndOverwrite(expectedOutputFilePrefix: String, actualOutput:
...
 
-// To prevent file duplication this function finds the file to compare the results to.
-// With preference for jdk: 17, 15, 13, 11, 8 (trace mode) and 17, 15, 13, 11, 8 (non trace mode)
-// Search starts at current jdk level
-// For instance we are running tests for jdk 11 (trace mode), we will check file existence in the following order:
-// 11 (trace), 8 (trace), 11 (non-trace) and 8 (non-trace).
-private fun getFileToCompareTo(expectedOutputFilePrefix: String): String {
-    // If in trace mode first check if a trace debugger file can be found
+/* To prevent file duplication, this function finds the file to compare the results to.
+ * With preference for jdk: 17, 15, 13, 11, 8 (trace mode) and 17, 15, 13, 11, 8 (non-trace mode).
```
bbrockbernd, line 77: This order is not valid anymore

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -55,53 +54,88 @@ private fun compareAndOverwrite(expectedOutputFilePrefix: String, actualOutput:
...
+ * Search starts at the current jdk level.
+ * For instance, we are running tests for jdk 11 (trace mode),
+ * we will check file existence in the following order:
+ * 11 (trace), 8 (trace), 11 (non-trace) and 8 (non-trace).
```
bbrockbernd, line 81: idem

bbrockbernd: Also in the docs of the .sh file btw

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
eupp: Oh, sorry, I did not realize it was for this purpose. 
I restored this behavior.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
eupp: I actually implemented both approaches, please have a look : )

eupp:
As a part of this PR, I also disabled spin-loop representation tests in trace debugger mode. 
Spin-loop detection is at the moment unsupported in the trace debugger mode. 
Because of this, the log files for these tests are just too long and not human-readable anyway.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -55,53 +54,114 @@ private fun compareAndOverwrite(expectedOutputFilePrefix: String, actualOutput:
...
+
+    val testConfigurations = listOfNotNull(
+        // first try to pick the most specific file if it exists
+        TestFileConfiguration(
+            jdkVersion = testJdkVersion,
+            inTraceDebuggerMode = isInTraceDebuggerMode
+        ),
+
+        // next, try to pick the jdk-specific file
+        TestFileConfiguration(
```
bbrockbernd, line 116: Suggestion: this can also be wrapped in `if (isInTraceDebuggerMode)` and the bottom two in `if (testJdkVersion != DEFAULT_TEST_JDK_VERSION)`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
bbrockbernd: Maybe also throw when there exists no file AND we are not in JDK17. To encourage first running and creating the default.

bbrockbernd (changes requested):

scripts/overwrite_representation_tests.sh
```diff
@@ -6,30 +6,41 @@
...
-#
+
+
 # This script runs all representation tests in overwrite mode.
-# It runs the tests on all jdks and both in trace and non-trace modes,
-# in the order jdk8 non-trace -> jdk8 trace -> jdk 11 non-trace -> jdk 11 trace, etc..
+# It runs the tests on all jdks and both in trace and non-trace modes, in the order
+# jdk-default non-trace -> jdk-default trace -> jdk-8 non-trace -> jdk 8 trace -> jdk-11 non-trace, jdk-11 trace, ...
+#
 # Note that due to current issue (https://github.com/JetBrains/lincheck/issues/500) we skip jdk8 trace.
```
ndkoval, line 15: We skip different tests for different reasons, let's remove this comment

ndkoval (approved):

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
eupp: Let's allow creating the file automatically if the `OVERWRITE_REPRESENTATION_TESTS_OUTPUT` flag is passed.
It is easier this way to create the file on the first run.

A developer can then look at generated file or diff to verify it.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
eupp: I'll keep it as is. The logic is already complicated enough, I would prefer to keep the number of `if`-s minimal.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
eupp: Oh, sorry, I see your point now. 
If there is no "default" file, it will always create the jdk-specific file. Ok, I'll fix that.

PR #563, opened 2025-03-10
Additional info in trace sent to plugin for code coloring

bbrockbernd:
Added two entries to the trace communicated from lincheck to the plugin.

- The first is an array of all related types. So for read and write points this is a single type (for example: `[java.lang.Integer]`). For calls the first entry in the array is the return type, the rest are the argument types `[returnType,arg1,arg2]`. (When a call does not have a return type the first entry is empty `[,arg1type,arg2type]`). Additionally when a type is an enum it is prepended with `Enum:`.
- The second added entry is a boolean that is true if a call is static.

Example of a trace element with a function `joe` that takes an `int` and returns an `int`, and is not `static`: `0;0;3;false;-1;joe(15): 16;org/jetbrains/kotlinx/lincheck_test/representation/BobTest:block:BobTest.kt:27;56;[java.lang.Integer,java.lang.Integer];false`

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -174,12 +174,15 @@ internal fun ManagedStrategy.runReplayIfPluginEnabled(failure: LincheckFailure) 
  * (due to difficulties with passing objects like List and TracePoint, as class versions may vary)
  *
  * Each trace point is transformed into the line of the following form:
- * `type;iThread;callDepth;shouldBeExpanded;eventId;representation;stackTraceElement;codeLocationId`.
+ * `type;iThread;callDepth;shouldBeExpanded;eventId;representation;stackTraceElement;codeLocationId;relatedTypes;isStatic`.
+ * 
  *
  *   stackTraceElement is "className:methodName:fileName:lineNumber" or "null" string if it is not applicable
  *   codeLocationId is strictly growing abstract id of location, and it must grow in syntactic order to
  *                  be able to order events occurred at same line in the same file. It is `-1` if it is not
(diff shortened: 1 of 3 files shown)


lev-serebryakov-jetbrains (approved):
Looks good as first prototype.
Kinds of objects can be added in next change, when this information will be integrated into plugin.

ndkoval:
@lev-serebryakov-jetbrains what do you mean by kind of objects?

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -229,9 +229,15 @@ private fun constructTraceForPlugin(failure: LincheckFailure, trace: Trace): Arr
...
+                
+                val readOrWriteType = when (event) {
+                    is ReadTracePoint -> event.valueType
+                    is WriteTracePoint -> event.valueType
+                    else -> ""
+                }
 
                 if (representation.isNotEmpty()) {
-                    representations.add("$type;${node.iThread};${node.callDepth};${node.shouldBeExpanded(false)};${eventId};${representation};${location};${locationId}")
+                    representations.add("$type;${node.iThread};${node.callDepth};${node.shouldBeExpanded(false)};${eventId};${representation};${location};${locationId};[${readOrWriteType}];false")
```
eupp, line 240: Please update the documentation of the `constructTraceForPlugin` with the description of the new parameters of trace points passed to the plugin.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
bbrockbernd: done!

eupp (approved):

PR #565, opened 2025-03-11
Pass read/write field trace point types to plugin

eupp:


eupp:
Was delivered as a part of https://github.com/JetBrains/lincheck/pull/579

PR #570, opened 2025-03-12
Fix missing type info for ij plugin

bbrockbernd:
Typeinfo wasn't always added (for instance atomicfu calls). Now it should be

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
@@ -283,12 +283,13 @@ internal class MethodCallTracePoint(
         this.thrownException = exception
     }
 
-    fun initializeParameters(parameters: List<String>) {
-        this.parameters = parameters
-    }
-    
-    fun initializeParameterTypes(parameters: List<String>) {
-        this.parameterTypes = parameters
+    fun initializeParametersRepresentationAndType(
+        parameters: List<Any?>, 
+        getValueRepresentation: (Any?) -> String, 
+        getTypeRepresentation: (Any?) -> String, 
+        ) {
+        this.parameters = parameters.map { getValueRepresentation(it) }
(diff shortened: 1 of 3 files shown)


lev-serebryakov-jetbrains (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -283,12 +283,13 @@ internal class MethodCallTracePoint(
...
-        this.parameters = parameters
-    }
-    
-    fun initializeParameterTypes(parameters: List<String>) {
-        this.parameterTypes = parameters
+    fun initializeParametersRepresentationAndType(
+        parameters: List<Any?>, 
+        getValueRepresentation: (Any?) -> String, 
+        getTypeRepresentation: (Any?) -> String, 
+        ) {
```
eupp, line 290: Indentation is broken here

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/TracePoint.kt
```diff
@@ -283,12 +283,13 @@ internal class MethodCallTracePoint(
         this.thrownException = exception
     }
 
-    fun initializeParameters(parameters: List<String>) {
-        this.parameters = parameters
-    }
-    
-    fun initializeParameterTypes(parameters: List<String>) {
-        this.parameterTypes = parameters
+    fun initializeParametersRepresentationAndType(
```
eupp, line 286: Let's rename to `initializeParameters` and pass `parameters: List<String>` and `parametersTypes: List<String>`. 
Then, in `ManagedStrategy` class you can declare extension function `MethodCallTracePoint.initializeParameters(parameters: List<Any?>)` that will apply necessary string conversions.

eupp (changes requested):

eupp (approved):

Issue #576, opened 2025-03-14
Bug in local object tracker leads to invalid thread switch attempt


dmitrii-artuhov:
The following test causes a check violation in lincheck source code on old `jdkToolchainVersion`'s (reproduced on 8 and 11)

```kotlin
@OptIn(ExperimentalModelCheckingAPI::class)
class ThreadListManipulationTest {

    @Test
    fun testModelCheckingCustom() {
        runConcurrentTest(50000) {
            val threads = mutableListOf<Thread>()
            val l = CountDownLatch(1)

            threads += thread {
                // threads; // making `threads` non-local allows the test to pass
                l.await()
            }

            threads += thread {
                l.await()
            }

            l.countDown()
            for (t in threads) {
                t.join()
            }
        }
    }
}
```

Gives the following output:
```txt
Wow! You've caught a bug in Lincheck.
We kindly ask to provide an issue here: https://github.com/JetBrains/lincheck/issues,
attaching a stack trace printed below and the code that causes the error.

Exception stacktrace:
java.lang.IllegalStateException: Trying to switch the execution to thread 2,
but only the following threads are eligible to switch: [1]
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.chooseThreadSwitch(ManagedStrategy.kt:773)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread(ManagedStrategy.kt:747)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread$default(ManagedStrategy.kt:732)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPoint(ManagedStrategy.kt:511)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.beforeWriteArrayElement(ManagedStrategy.kt:1127)
	at sun.nio.ch.lincheck.Injections.beforeWriteArray(Injections.java:348)
	at java.base/java.util.ArrayList.add(ArrayList.java:487)
	at java.base/java.util.ArrayList.add(ArrayList.java:499)
	at kotlinx.coroutines.gpmc.internal.LockFreeLinkedListLongModelCheckingTest$testSome$1.invoke(LockFreeLinkedListLongModelCheckingTest.kt:58)
[...]

PR #578, opened 2025-03-18
Add support for intrinsic candidate methods handling

dmitrii-artuhov:
Resolves issue #576


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/IntrinsicCandidateMethodFilter.kt
```diff
@@ -0,0 +1,69 @@
...
+        return super.visitAnnotation(desc, visible)
+    }
+
+    /**
+     * Changes the delegate [MethodVisitor] to redirect to transformer "further" in the chain.
+     * Essentially it allows to skip all transformers between `nextAdapter` and `initialAdapter`.
+     */
+    private fun delegate() {
+        this.mv = initialAdapter
+    }
```
eupp, line 48: This solution is fine for now. But I am concerned it can potentially be error-prone.
What if delegation happens in the middle of `visitX` chains? 
Some of the `visitX` methods will be delegated to one transformer, then others to another transformer. 

In the future, I think we can improve this by using two passes over bytecode. The first pass can collect all necessary information (including what methods are intrinsic), so the second pass already has all necessary information and can make the decision at the beginning. 

Anyway, we witness more and more similar problems (need to compute some information beforehand), so we will need to implement two-pass approach in the future. 

Or, alternatively, instead of 2-pass, we probably can utilize `ClassNode` API, which could allow us to pre-compute all required information before doing instrumentation:
https://asm.ow2.io/javadoc/org/objectweb/asm/tree/ClassNode.html

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1353,6 +1355,22 @@ abstract class ManagedStrategy(
         }
     }
 
+    private fun processIntrinsicMethodEffectsOnObjectTracker(
+        methodId: Int,
+        result: Any?
+    ) {
+        if (!IntrinsicMethods.isIntrinsicMethod(methodId)) return
```
eupp, line 1362: Maybe use `check(...)` here instead?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/IntrinsicCallNames.kt
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        return IntrinsicCallType.UnknownIntrinsicCall
+    }
+}
+
+internal sealed interface IntrinsicCallType {
+    data object UnknownIntrinsicCall : IntrinsicCallType
+    data object ArraysCopyOfIntrinsicCall : IntrinsicCallType
+    data object ArraysCopyOfRangeIntrinsicCall : IntrinsicCallType
```
eupp, line 44: Let's just implement these as extension methods of `MethodSignature`,  e.g., `isArraysCopyOfIntrinsicCall()`, `isArraysCopyOfRangeIntrinsicCall()`, etc.

eupp: Then it also would make sense to move them to `MethodSignature` file, or at least to `utils`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1353,6 +1355,22 @@ abstract class ManagedStrategy(
         }
     }
 
+    private fun processIntrinsicMethodEffectsOnObjectTracker(
```
eupp, line 1358: Maybe just `processIntrinsicMethodEffects` ?
In general, an intrinsic method can have other effects, affecting not only `objectTracker`.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/IntrinsicCandidateMethodTest.kt
```diff
@@ -0,0 +1,53 @@
...
+ * See [Invalid thread switch attempt](https://github.com/JetBrains/lincheck/issues/576).
+ */
+@OptIn(ExperimentalModelCheckingAPI::class)
+class IntrinsicCandidateMethodTest {
+
+    @Test
+    fun testArraysCopyOf() {
+        runConcurrentTest(10000) {
+            val threads = mutableListOf<Thread>()
+            val l = CountDownLatch(1)
```
eupp, line 33: Is `CountDownLatch` necessary for the test to fail (without intrinsic methods support)?
If so, please add a comment about it.

eupp (changes requested):

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/IntrinsicCandidateMethodTest.kt
dmitrii-artuhov: I have rewritten test using atomic integer to simplify it. Added a comment which explains that the operation inside each thread does not matter, but with some it is easier to trigger jit

dmitrii-artuhov: On the second thought: the same test with `AtomicInteger` sometimes passes, and it is harder to reproduce the issue with it. So I will stick to `CountDownLatch` and add a comment that with primitves issue exists but sometimes does not reproduce

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Types.kt
```diff
@@ -16,4 +16,43 @@ import sun.nio.ch.lincheck.Types.*
...
+    val className: String,
+    val methodSignature: MethodSignature
+) {
+    constructor(className: String, methodName: String, desc: String) :
+        this(className, MethodSignature(methodName, convertAsmMethodType(desc)))
+
+    val methodName: String get() = methodSignature.name
+    val returnType: Type get() = methodSignature.methodType.returnType
+    val argumentTypes: List<Type> get() = methodSignature.methodType.argumentTypes
+}
```
eupp, line 31: Please add `toString()` override for `MethodDescriptor` (after rebase you'll see other classes also have this override #588).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1448,8 +1470,11 @@ abstract class ManagedStrategy(
         return deterministicMethodDescriptor
     }
 
-    override fun onMethodCallReturn(descriptorId: Long, descriptor: Any?, receiver: Any?, params: Array<Any?>, result: Any?) {
+    override fun onMethodCallReturn(descriptorId: Long, descriptor: Any?, methodId: Int, receiver: Any?, params: Array<Any?>, result: Any?) {
         runInIgnoredSection {
+            // process intrinsic candidate methods which instantiate objects
+            processIntrinsicMethodEffects(methodId, result)
```
eupp: Actually, it makes more sense to move the `if` check here, so it is obvious for the code reader that this processing is only applied to the intrinsic methods:
```kotlin
if (MethodIds.isIntrinsicMethod(methodId)) {
    processIntrinsicMethodEffects(methodId, result)
}
```

eupp (changes requested):
Ok, I think we almost done here.

Please address few remaining minor comments (also note there is at least one unresolved comment from the previous review).

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Types.kt
dmitrii-artuhov: added

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: also added a `check(...)` inside the `processIntrinsicMethodEffects`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: added

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/IntrinsicCandidateMethodFilter.kt
dmitrii-artuhov: On the first attempt, I actually added a 2-pass solution. By it seemed to work pretty slow.
When this problem appears again we should check for efficiency of 2-pass solution

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Types.kt
```diff
@@ -16,4 +16,9 @@ import sun.nio.ch.lincheck.Types.*
 import sun.nio.ch.lincheck.Types.convertAsmMethodType
 
 internal fun Method.toMethodSignature() = MethodSignature(this.name, convertAsmMethodType(this.descriptor))
-internal fun java.lang.reflect.Method.toMethodSignature() = Method.getMethod(this).toMethodSignature()
\ No newline at end of file
+internal fun java.lang.reflect.Method.toMethodSignature() = Method.getMethod(this).toMethodSignature()
+
+object PredefinedTypes {
+    internal val ARRAY_OF_OBJECTS_TYPE = ArrayType(ObjectType("java.lang.Object"))
+    internal val CLASS_TYPE = ObjectType("java.lang.Class")
```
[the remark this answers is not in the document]
dmitrii-artuhov: habit from university java homeworks)

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/CodeLocations.kt
```diff
@@ -69,6 +74,23 @@ internal object MethodIds {
     }
 }
 
+internal object IntrinsicMethods {
+    private val methodDescriptors: ConcurrentHashMap<Int, Pair<String, MethodSignature>> = ConcurrentHashMap()
+    private val methodIds: MutableSet<Int> = Collections.newSetFromMap(ConcurrentHashMap())
```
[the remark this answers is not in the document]
dmitrii-artuhov: `MethodDescriptor` is okey I think. Cannot think of anything else either 🤷‍♂️

eupp (approved):

Issue #409, opened 2024-10-01
Halting problem when using I/O


I-Gleb:
Lincheck version 2.34

After adding debug output to the data structure testing is not finishing.

DS under test:
```
package day1

import kotlinx.atomicfu.*

interface MyQueue<E> {
    fun enqueue(element: E)
    fun dequeue(): E?
    fun validate() {}
}

class MSQueue<E> : MyQueue<E> {
    private val head: AtomicRef<Node<E>>
    private val tail: AtomicRef<Node<E>>

    init {
        val dummy = Node<E>(null)
        head = atomic(dummy)
        tail = atomic(dummy)
    }

    override fun enqueue(element: E) {
        val newTail = Node(element)
        while (true) {
            val currentTail  = tail.value
            if (currentTail.next.compareAndSet(null, newTail)){
                tail.compareAndSet(currentTail, newTail)
                println("Enqueue $element" + toString())
                return
            } else {
                tail.compareAndSet(currentTail, currentTail.next.value!!)
            }
        }
    }

    override fun dequeue(): E? {
        while (true) {
            val currentHead = head.value
            val currentNext = currentHead.next.value ?: return null

            if (head.compareAndSet(currentHead, currentNext)){
                val result = currentNext.element
                println("Dequeued $result" + toString())
                return result
            }
        }
    }

    private class Node<E>(
        var element: E?
    ) {
        val next = atomic<Node<E>?>(null)
    }

    override fun toString(): String {
        val builder = StringBuilder(" | ")
        var current = head.value
        while (current != tail.value) {
            builder.append(current.element)
            builder.append(" ")
            current = current.next.value!!
        }
        builder.append(" | ${hashCode()}")
        return builder.toString()
    }
}
````

Test class:
```
package day1

import org.jetbrains.kotlinx.lincheck.annotations.*
import org.jetbrains.kotlinx.lincheck.paramgen.*
import org.jetbrains.kotlinx.lincheck.*
import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.*
import org.junit.*
import org.junit.runners.*
import kotlin.reflect.*

@Param(name = "element", gen = IntGen::class, conf = "0:3")
abstract class AbstractQueueTest(
    private val queue: MyQueue<Int>,
    checkObstructionFreedom: Boolean = true,
    threads: Int = 3,
    actorsBefore: Int = 1
) : TestBase(
    sequentialSpecification = IntQueueSequential::class,
[...]

PR #580, opened 2025-03-19
Fix hanging problems with standard streams

dmitrii-artuhov:
This PR fixes issue #409

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -148,6 +148,9 @@ abstract class ManagedStrategy(
     // User-specified guarantees on specific function, which can be considered as atomic or ignored.
     private val userDefinedGuarantees: List<ManagedStrategyGuarantee>? = testCfg.guarantees.ifEmpty { null }
 
+    // Predefined guarantees on specific `PrintStream` instances on which we do not want to track method calls.
+    private val predefinedGuarantees: List<ManagedStrategyGuarantee> = listOf(forInstances(System.`in`, System.out, System.err).allMethods().ignore())
+
     // Utility class for the plugin integration to provide ids for each trace point
     private var eventIdProvider = EventIdProvider()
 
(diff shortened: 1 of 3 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1356,12 +1359,18 @@ abstract class ManagedStrategy(
     }
 
     private fun methodGuaranteeType(owner: Any?, className: String, methodName: String): ManagedGuaranteeType? = runInIgnoredSection {
-        userDefinedGuarantees?.forEach { guarantee ->
-            val ownerName = owner?.javaClass?.canonicalName ?: className
-            if (guarantee.classPredicate(ownerName) && guarantee.methodPredicate(methodName)) {
-                return guarantee.type
+        predefinedGuarantees
```
eupp, line 1362: Let's not re-use `ManagedStrategyGuarantee` mechanism for this purpose. 

1. I remember there were some performance problem, because the lookup through the list of guarantees was working slowly. 
2. We might revisit their API at some point.

So could you just hard-code here handling of I/O streams?

In the long run, we need to revisit this place in the code anyway, because e.g., I think we also need to return `Ignored` guarantee for classes for which `shouldTransform` returns false. 
So we need some centralized mechanism to specify guarantees of different classes/methods.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/PrintingTest.kt
```diff
@@ -21,26 +20,30 @@ import kotlin.concurrent.thread
...
+     *
      * See [CyclicBarrierTest] for identical test but with no printing.
      */
-    @Ignore("Println's hang the execution")
     @Test(timeout = TIMEOUT)
     fun testPrints() {
         runConcurrentTest(1) {
             val barrier = CyclicBarrier(2) {
-                println("All threads have reached the barrier. Proceeding...")
+                System.err.println("All threads have reached the barrier. Proceeding...")
```
eupp, line 32: It makes sense to add tests both for regular Kotlin's `println` as well as `System.our` and `System.err`.

eupp (changes requested):

eupp (approved):

PR #581, opened 2025-03-20
Add new integration tests that show problems with identity hash code

ivandev0:
These tests show that with every execution of an agent we get
different identity hash code for objects without predefined
hash code. This should not be a problem during normal execution, but
it is a problem for integration testing. We need to find a way to get
more stable results.


eupp:
> These tests show that with every execution of an agent we get different identity hash code for objects without predefined hash code. 

I just realized, that we will have similar problem not only for hash-codes, but also for other sources of non-determinism. 

For example, suppose the program uses time or random APIs. On each re-run trace will be different, even though replays withing the same run should be deterministic (assuming that used non-deterministic API is supported in trace debugger).

ivandev0:
> I just realized, that we will have similar problem not only for hash-codes, but also for other sources of non-determinism.

Yep, this is the great point. I have a couple of ideas how we can avoid that.
1. Hack everything in Lincheck. For hash codes we can use [-XX:hashCode](https://shipilev.net/jvm/anatomy-quarks/26-identity-hash-code/) flag, for random we can hard code some const seed.
2. Don't check a trace that is non-deterministic. Most of the time such complicated logic comes from libraries, for example, if we speak about hash codes then the problem comes from LinkedHashMap and not from user's code. We can hide such trace, but with this approach we are limiting our testing ground.

eupp:
> Hack everything in Lincheck. For hash codes we can use [-XX:hashCode](https://shipilev.net/jvm/anatomy-quarks/26-identity-hash-code/) flag, for random we can hard code some const seed.

Yeap, this can work, but we would need to invent different devices for each non-determinism feature used in the test. 
- For hash codes, using `-XX:hashCode` is good idea.
- For random, yes, we can fix the seeds. The question is though, is there some way to somehow fix the seeds globally for the whole test (without modifying its source code)?
- What to do about time APIs ?
-  Filesystem IO should probably work fine, if the test uses it deterministicall (e.g., always creates file with the same name, writes the same content, etc).

Note that there is a significant intersection here with the native calls record/replay feature we are implementing in the trace debugger itself. We already have a mechanism to replay the trace by substituting results of some method calls.
So, perhaps, we can also just re-use that?

It would probably require some work and some refactoring in the code, but could be quite useful in the long run. 

So the workflow can be as follows:
- We record the trace and save the recorded trace in the file. 
- In the test, we ask the trace debugger to replay the trace stored in the file. If the trace can be replayed successfully => test is passed.  

So in terms of coding, we will need to introduce a "replay" mode in trace debugger, where instead of collecting the trace on the first run, it takes the trace from the file and tries to replay it.

ivandev0:
Sounds interesting, I will investigate this option.

ivandev0:
Unfortunately, the hash code test produce different results when running on CI and locally

build.gradle.kts
```diff
@@ -76,10 +77,13 @@ kotlin {
             val junitVersion: String by project
             val jctoolsVersion: String by project
             val mockkVersion: String by project
+            val gradleToolingApiVersion: String by project
             dependencies {
                 implementation("junit:junit:$junitVersion")
                 implementation("org.jctools:jctools-core:$jctoolsVersion")
                 implementation("io.mockk:mockk:${mockkVersion}")
+                implementation("org.gradle:gradle-tooling-api:${gradleToolingApiVersion}")
+                runtimeOnly("org.slf4j:slf4j-simple:1.7.10")
```
ndkoval, line 86: Why do you need it?

ndkoval (approved):

build.gradle.kts
ivandev0: Do you mean why we need `slf4j`? We need it for Gradle API. Otherwise there is the following warning in the console `SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder"...`. I can drop it if you are OK with that.

ndkoval: BTW, do you need to add these dependencies for jvmTest, or is adding them only for your new task sufficient?

I'm fine with adding the dependency, but please comment in the code on why it is necessary.

ivandev0: > BTW, do you need to add these dependencies for jvmTest, or is adding them only for your new task sufficient?

It is enough to add them only for my tests. Should I create a new source set for trace debugger tests?

ivandev0:
I moved the tests into new source set. Please take a look

ndkoval (approved):

PR #582, opened 2025-03-20
Add permanent run configuration to simplify trace-debugger's agent debug

ivandev0:


.idea/runConfigurations/Trace_debugger_Remote.xml
```diff
@@ -0,0 +1,15 @@
+<component name="ProjectRunConfigurationManager">
```
eupp, line 1: Should we add some documentation for this file?
Something like: "This file set-ups a remote debug configuration to add a possibility to debug Lincheck code when its JVM agent is attached to another application."

eupp (approved):

Issue #328, opened 2024-05-30
Enforce consistent rules of internal to canonical class name conversion


eupp:
Currently, there are several places in the code base where conversions between internal and canonical class names happen.

Currently these conversions happen chaotically, which leads to bugs and performance penalties.
For example, sometimes the internal class name is passed into `ManagedStrategy` methods only to be immediately converted into canonical class name:
https://github.com/JetBrains/lincheck/blob/ecede824e9b3399a499c9efb5aba9d69dae91aa9/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt#L707
Because of this, additional redundant memory allocations happen at runtime. 

We need to carefully review all the places when the conversions happen and enforce unified rules and invariants across all codebase. 

One possible approach would be to enforce the invariant that internal class names are only used in the `transformation` module, and they are always converted to canonical class names during byte-code transformation. 

Using two values classes: `CanonicalClassName` and `InternalClassName`, can help to enforce this invariant and make code easier to read and maintain.

PR #583, opened 2025-03-20
Ensure internal class names do not leak from `transformation` module

eupp:
Currently in the code base two different formats of JVM class names are used: internal and canonical class names. 

Internal class names are primarily used in the `transformation` module in conjunctions with ASM library and other bytecode instrumentation facilities. Other modules of code actively use Java Reflection API and other APIs that expect class names to be given in canonical class names format. 

However, currently in the code base internal class names are leaked from `transformation` module, which sometimes leads to hard-to-spot bugs. 

The solution to the problem is to accept the convention to never leak class names in internal format beyond `transformation` module, and always perform the conversions on the boundaries of `transformation` module. This PR implements this policy. 

This PR addresses #328.

**Alternatives considered.**

There was an idea to use `typealias` or `inline value class` features of Kotlin to distinguish between internal and canonical class name types. After some considerations, it was decided to not implement these approaches. 

- `typealias`  can give more meaningful names to types and help to distinguish two class name formats for code reader. But it would not prevent bugs, e.g. passing `InternalClassName` into a function expecting `CanonicalClassName`. However it would still require substantial refactoring effort to change the type names in a lot of places. 

- `inline value class` would help to prevent accidental bugs. However, we need to interop with Java libraries such as ASM and ByteBuddy, which would not handle Kotlin's inline value classes out-of-the-box. Writing an integration-layer APIs to interop our value classes with these libraries also seems like a substantial effort.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
@@ -78,7 +78,7 @@ internal class MethodCallTransformer(
             null -> pushNull()
             else -> loadLocal(receiver)
         }
-        push(owner.canonicalClassName)
+        push(owner.toCanonicalClassName())
         push(name)
         loadNewCodeLocationId()
         // STACK [INVOKEVIRTUAL]: owner, owner, className, methodName, codeLocation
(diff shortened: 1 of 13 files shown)


zhelenskiy (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/MethodCallTransformer.kt
```diff
@@ -95,16 +94,16 @@ internal class MethodCallTransformer(
...
         storeLocal(deterministicMethodDescriptor)
-        
+
         val deterministicCallIdLocal = newLocal(LONG_TYPE)
         pushDeterministicCallId(deterministicMethodDescriptor)
         storeLocal(deterministicCallIdLocal)
-        
+
         invokeBeforeEventIfPluginEnabled("method call $methodName", setMethodEventId = true)
-
```
dmitrii-artuhov, line 107: We should add a linter to our development cycle leter. So that it at least deals with the left tabulations automatically

dmitrii-artuhov (approved):

PR #587, opened 2025-03-26
Include spin-cycle replays in invocations count 

dmitrii-artuhov:
This PR adds a functionality which counts each spin-cycle replay as part of the total invocations count.

Examples of tests, where there is a significant difference between the specified invocations and actual run invocations for ModelChecking regime (among all iterations the biggest difference in invocations is specified):
1. CausesBlockingOperationTest: `15236 / 10000`
1. CyclicBarrierTest: `24162 / 10000`
1. ReentrantLockTest: `2314 / 500`
1. SpinLockTest: `1796 / 500`
1. CoroutineCancelTest::testCancel: `1482 / 100`
1. CoroutineJoinTest: `868 / 100`
1. ReentrantLockSetTest: `3868 / 1000`
... and a couple more tests with a smaller difference

All tests run locally with the same configuration (naive benchmarking, difference could be just an error):
- included spin-cycles replays: `9m 25s`
- excluded spin-cycle replays (old behaviour): `10m 33s`

---------
An issue, that describes a problem, introduces by this PR: https://github.com/JetBrains/lincheck/issues/590

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -180,6 +180,7 @@ tasks {
         extraArgs.add("-Dlincheck.version=$version")
         findProperty("lincheck.logFile")?.let { extraArgs.add("-Dlincheck.logFile=${it as String}") }
         findProperty("lincheck.logLevel")?.let { extraArgs.add("-Dlincheck.logLevel=${it as String}") }
+        findProperty("lincheck.strictInvocationsBound")?.let { extraArgs.add("-Dlincheck.strictInvocationsBound=${it as String}") }
         jvmArgs(extraArgs)
     }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -180,7 +180,6 @@ tasks {
         extraArgs.add("-Dlincheck.version=$version")
         findProperty("lincheck.logFile")?.let { extraArgs.add("-Dlincheck.logFile=${it as String}") }
         findProperty("lincheck.logLevel")?.let { extraArgs.add("-Dlincheck.logLevel=${it as String}") }
-        findProperty("lincheck.strictInvocationsBound")?.let { extraArgs.add("-Dlincheck.strictInvocationsBound=${it as String}") }
         jvmArgs(extraArgs)
     }
 
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
@@ -29,6 +29,14 @@ abstract class Strategy protected constructor(
 ) : Closeable {
 
     /**
+     * Strict invocations bound will treat spin cycle replays in [ManagedStrategy] as actual invocations.
+     *
+     * Setting this flag to `true` could lead to existing errors being undetected by the strategy
+     * when small number of invocations is set (like, 1) and spin cycle replay is required.
+     */
+    internal val strictInvocationsBound: Boolean = System.getProperty("lincheck.strictInvocationsBound").toBoolean()
+
+    /**
      * Runner used for executing the test scenario.
      */
     internal abstract val runner: Runner
@@ -116,9 +124,23 @@ abstract class Strategy protected constructor(
  * @return the failure, if detected, null otherwise.
  */
 fun Strategy.runIteration(invocations: Int, verifier: Verifier): LincheckFailure? {
+    var isReplayingSameInvocation = false
     for (invocation in 0 until invocations) {
-        if (!nextInvocation()) return null
-        val result = runInvocation()
+        if (!isReplayingSameInvocation && !nextInvocation()) return null
+        isReplayingSameInvocation = false
+        var result: InvocationResult = runInvocation()
+
+        if (strictInvocationsBound) {
+            if (result is SpinCycleFoundAndReplayRequired) {
+                isReplayingSameInvocation = true
+                continue
+            }
+        }
+        else {
+            while (result is SpinCycleFoundAndReplayRequired) {
+                result = runInvocation()
+            }
+        }
 
         val failure = try {
             verify(result, verifier)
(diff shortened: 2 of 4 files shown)


build.gradle.kts
```diff
@@ -180,6 +180,7 @@ tasks {
         extraArgs.add("-Dlincheck.version=$version")
         findProperty("lincheck.logFile")?.let { extraArgs.add("-Dlincheck.logFile=${it as String}") }
         findProperty("lincheck.logLevel")?.let { extraArgs.add("-Dlincheck.logLevel=${it as String}") }
+        findProperty("lincheck.strictInvocationsBound")?.let { extraArgs.add("-Dlincheck.strictInvocationsBound=${it as String}") }
```
eupp, line 183: Do we want a flag for this? Why shouldn't it be just regular expected behavior?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -116,9 +124,23 @@ abstract class Strategy protected constructor(
  * @return the failure, if detected, null otherwise.
  */
 fun Strategy.runIteration(invocations: Int, verifier: Verifier): LincheckFailure? {
+    var isReplayingSameInvocation = false
     for (invocation in 0 until invocations) {
-        if (!nextInvocation()) return null
-        val result = runInvocation()
+        if (!isReplayingSameInvocation && !nextInvocation()) return null
```
eupp, line 129: Alternatively, we can tweak `nextInvocation()` implementation so it would behave differently after `enableSpinCycleReplay()` was called at the end of previous invocation. 

This way we can keep `runIteration()` implementation simpler, hiding the implementation details of managed strategy.

eupp: You'll probably only need to modify `verify` method, so it does not actually calls verifier in case of `SpinCycleFoundAndReplayRequired` invocation result.

eupp (changes requested):

build.gradle.kts
dmitrii-artuhov: Assume I have such test (which requires many replays of the spin cycles, to minimize them):
```kotlin
runConcurrentTest(1) {
  val pool = Executors.newFixedThreadPool(1).asCoroutineDispatcher()
  runBlocking(pool) { /* do nothing */ }
  pool.close()
  check(false)
}
```

Having here `strictInvocationBounds=true` and only 1 invocation will allow this test to pass. Which looks completely misleading in my opinion. I would prefer this test to always fail, even if a single invocation is specified. Thus, I would have the `strictInvocationBounds` as a flag for benchmarks. 

Such a test could be a very rare case, but still, I, as a user, have no idea about spin-cycle replays, and I would have no clue why such a test passes.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
dmitrii-artuhov: I have hidden the details inside the [ModelChecking/Managed]Startegy

build.gradle.kts
eupp: Hmm, I see your point. 

I still think the problem when number of actual invocations is much more greater than requested number is worse. 

- It breaks any execution time estimations which a user might have when it sets the invocations number. 
- Because of this, some of our tests on CI take more time than expected. 
- It is simply a confusing behavior.

I agree that a behavior when:
```
runConcurrentTest(invocations = 1) {
    repeat(SOME_BIG_NUMBER) { ... }
    check(false)
}
```

does not fail can also be confusing, but IMO less-so, because of the following. 

1. Once the user discovers the spin-loop detection feature and how it works, the behavior becomes more logical. Yes, long-running loop with a lot of iterations may confuse the tool into thinking it found live-lock. It is false positive love-lock, but okay, the world is imperfect, halting problem is undecidable in general, I can just increase the spin-loop detection threshold. 

2. Chances that user deliberately writes a test with `invocations = 1` are pretty low. For large enough invocations count (say even 100), the test is probably going to fail, because the number of times spin-loop replay is requested is proportional to the number of "long" loops in the test (which should not be large). 

3. On the other hand, the problem when actual number of invocations differs from requested affects almost any test.  


I would propose to change the default behavior, remove the flag, and create an issue about the `invocations = 1` problem, so we can think how to improve the spin-loop detection algorithm and make the behavior more predictable for such cases.

dmitrii-artuhov: As an alternative, I could suggest that we allow `ManagedStrategy` to finish its last invocation, completing all spin-cycles. What I mean is that we allow the last invocation to complete fully instead of aborting it mid-way through spin-cycle replays, even if the total invocations is less than required by the strategy (`i` means invocation, and the following `spin`s mean spin-cycle replay for previous invocation):
```txt
i1 spin spin spin i2 spin spin spin spin spin stop
                           ^
                           |
                   total invocations          but we allow i2 to complete gracefully
```

Such behaviour would solve the problem with test, that I have written earlier. And the real invocations number will not exceed total too much. And for strict invocations count we can leave the flag.

What do you think? Because I assume that hardly any user of lincheck will dig into its sources and even less will take a look at spin-cycle replayer))

eupp: > Because I assume that hardly any user of lincheck will dig into its sources and even less will take a look at spin-cycle replayer))

Yes, I agree. I think that in general, the behavior of spin-loop detection algorithm could be confusing for the user. 
We probably can partly solve this by providing more feedback and output something to the console (or the log), whenever an invocation is "aborted and skipped" because of the requested spin-loop replay.

I also think that in general, whenever test fails because of the detected livelock, we should add a message about possibility of false positive livelock report, and explain how to tune the spin-loop detection parameters. 

> As an alternative, I could suggest that we allow ManagedStrategy to finish its last invocation, completing all spin-cycles.

Yes, this looks reasonable. But let's do it separately in another PR. Could you also please create an issue ticket with the description of the problem, an example, and list of possible solutions (I think for now the list is just one solution you just proposed in the previous comment).

dmitrii-artuhov: what are your suggestions on this PR?
Leave how it is or remove flag or something else?

dmitrii-artuhov: added an issue: https://github.com/JetBrains/lincheck/issues/590

eupp (approved):

PR #594, opened 2025-03-29
Fix the illegal `lookupClass` issue when using `MethodHandles.Lookup` on JDK 8

zhelenskiy:
The issue was caused by the fact that `MethodHandles.lookup`, needed for custom handling of invoke-dynamic cannot be called from certain locations. In the case of Java > 8, the only limitation is `java.lang.invoke`, which is fine. While in Java 8 it also includes all Java stdlib packages. So the workaround was to call a private constructor of `MethodHandles` privately with reflection.

The issue was reproducible on one of the tests from Kotlin compiler repository.

### DIFF
--- a/bootstrap/src/sun/nio/ch/lincheck/Injections.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/Injections.java
@@ -11,6 +11,8 @@
 package sun.nio.ch.lincheck;
 
 import java.lang.invoke.CallSite;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
 
 import static sun.nio.ch.lincheck.Types.convertAsmMethodType;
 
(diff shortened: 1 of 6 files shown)


ndkoval:
@zhelenskiy, please update the PR title and description to indicate what this PR fixes.

ndkoval:
@zhelenskiy, please also add an integration test for the issue.

zhelenskiy:
> @zhelenskiy, please also add an integration test for the issue.

The integration test for this issue is considered too complex, so we decided not to include it, as it would require too much effort. It would be possible to include the whole Kotlin project once it would start working completely fine.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1466,6 +1466,9 @@ abstract class ManagedStrategy(
             processIntrinsicMethodEffects(methodId, result)
         }
 
+        if (deterministicMethodDescriptor != null) {
+            Logger.debug { "On method return with descriptor $deterministicMethodDescriptor: $result" }
+        }
```
eupp, line 1471: Should this be the first statement of the method?

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -69,7 +70,10 @@ internal class LincheckClassVisitor(
         exceptions: Array<String>?
     ): MethodVisitor {
         var mv = super.visitMethod(access, methodName, desc, signature, exceptions)
-        if (access and ACC_NATIVE != 0) return mv
+        if (access and ACC_NATIVE != 0) {
+            Logger.debug { "Skipping native method $className.$methodName" }
```
eupp, line 74:
```suggestion
            Logger.debug { "Skipping transformation of the native method $className.$methodName" }
```

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -632,4 +634,24 @@ public static int getNextEventId(String type) {
...
+    
+    public static Class<?> getClassForNameOrNull(String name) {
+        try {
+            return Class.forName(name);
+        } catch (ClassNotFoundException e) {
+            return null;
+        }
+    } 
+    
+    public static MethodHandles.Lookup trustedLookup(Class<?> callerClass) throws Exception {
```
eupp, line 649: Please add a documentation explaining what this method does, and why it is needed.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -632,4 +634,24 @@ public static int getNextEventId(String type) {
     public static void setLastMethodCallEventId() {
         getEventTracker().setLastMethodCallEventId();
     }
+    
+    
+    private static Constructor<MethodHandles.Lookup> lookUpPrivateConstructor = null;
```
eupp, line 639: Please move near `trustedLookup`.

eupp (changes requested):

eupp (approved):

PR #605, opened 2025-04-07
Support thread interruptions

eupp:


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -500,23 +497,137 @@ abstract class ManagedStrategy(
...
+        threadScheduler.scheduleThread(nextThread)
+    }
+
+    private fun throwIfInterrupted() {
+        if (Thread.interrupted()) {
+            throw InterruptedException()
+        }
+    }
+
+    private fun unblockInterruptedThreads() {
```
ndkoval, line 593: Let's add documentation here

ndkoval (approved):

PR #531, opened 2025-02-18
Add coroutine tests from deadlock test repo

bbrockbernd:
Closes #492

PR #606, opened 2025-04-08
Check that only `TestThread`s perform special suspend/resume logic

dmitrii-artuhov:
PR adds some checks that make sure that only test threads (created by the runner) perform manual suspend/resume logic.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -505,7 +505,7 @@ abstract class ManagedStrategy(
      */
     private fun isTestThreadCoroutineSuspended(iThread: Int): Boolean =
         (
-           isTestThread(iThread) &&
+            isTestThread(iThread) &&
             // TODO: coroutine suspensions are currently handled separately from `ThreadScheduler`
             isSuspended[iThread]!! &&
             !runner.isCoroutineResumed(iThread, currentActorId[iThread]!!)
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -500,6 +500,9 @@ abstract class ManagedStrategy(
         }
     }
 
+    /**
+     * Returns whether in test thread [iThread] coroutine has been suspended but not yet resumed.
+     */
     private fun isTestThreadCoroutineSuspended(iThread: Int): Boolean =
         (
             isTestThread(iThread) &&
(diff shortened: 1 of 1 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -507,7 +507,10 @@ abstract class ManagedStrategy(
     private fun isActive(iThread: Int): Boolean =
         threadScheduler.isSchedulable(iThread) &&
         // TODO: coroutine suspensions are currently handled separately from `ThreadScheduler`
-        !(isSuspended[iThread]!! && !runner.isCoroutineResumed(iThread, currentActorId[iThread]!!))
+        (
+            !isTestThread(iThread) ||
+            !(isSuspended[iThread]!! && !runner.isCoroutineResumed(iThread, currentActorId[iThread]!!))
+        )
```
eupp, line 513: Let's extract this check into a method.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1593,6 +1596,7 @@ abstract class ManagedStrategy(
         methodParams: Array<Any?>,
         atomicMethodDescriptor: AtomicMethodDescriptor?,
     ): Boolean {
+        check(isTestThread(threadId)) { "Special coroutines handling is only required for test threads" }
```
eupp, line 1599:
```suggestion
        check(isTestThread(threadId)) { 
            "Special coroutines handling methods should only be called from test threads" 
        }
```

eupp: Same in other places.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2029,6 +2036,10 @@ abstract class ManagedStrategy(
 
     // == UTILITY METHODS ==
 
+    private fun isTestThread(threadId: Int): Boolean {
+        return threadId in (0..<nThreads)
```
eupp, line 2040:
```suggestion
        return threadId in (0 ..< nThreads)
```

eupp (changes requested):

eupp:
Does this fix resolves the problem with the test from https://github.com/JetBrains/lincheck/pull/531 ?

dmitrii-artuhov:
> Does this fix resolves the problem with the test from #531 ?

This was the bug I discovered when debugging test `channel05`, but fixing it does not allow that test to pass. I am still working on that one

eupp (approved):

Issue #596, opened 2025-03-31
Tracking exceptions in "user" threads with GPMC


dmitrii-artuhov:
## Problem
```kotlin
    @Test
    fun testInfiniteBackgroundThreadError() {
        runConcurrentTest(1) {
            val t = thread {
                check(false)
            }
            t.join()
        }
    }
```

This test hangs when run using GPMC.

## Expected
- Test does not hang
- Test ? _fails/succeeds_ ? with trace where child thread is shown with the result of exception => **the behavior should be discussed separately. Right now, the implementation does not stop interleaving exploration when an exception is thrown in user thread**

PR #609, opened 2025-04-08
Finish user thread when exception is throw from it

dmitrii-artuhov:
Fixes: #596

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -691,7 +691,7 @@ abstract class ManagedStrategy(
             // re-throw any non-internal exception,
             // so it will be treated as the final result of `Thread::run`.
             disableAnalysis()
-            Logger.info { "Exception was thrown in user thread:" }
+            Logger.info { "Exception was thrown in user thread \"Thread-$currentThreadId\":" }
             Logger.info(exception)
             // current thread will not be put in the ABORTED state, like in `onInternalException` case,
             // thus, it is still executed in isolation, and we can finish it properly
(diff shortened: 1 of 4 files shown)


eupp:
Could you also please check that after this changes this test works:
https://github.com/JetBrains/lincheck/blob/dacf5d9bdaf76c670fd57bc1ef8a225675313f40/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/InterruptionTest.kt#L137

eupp:
Does this fix resolves the problem with the test from https://github.com/JetBrains/lincheck/pull/531

dmitrii-artuhov:
> Could you also please check that after this changes this test works:
> 
> https://github.com/JetBrains/lincheck/blob/dacf5d9bdaf76c670fd57bc1ef8a225675313f40/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/InterruptionTest.kt#L137

That test works.

Right now I am only finishing the thread when non-internal exception was thrown from it. This guarantees, that the the thread will not be aborted (like in `onInternalException`) and it is still executed in isolation. Thus, I call `onThreadFinish(...)` and then rethrow exception for user to handle. However, the actual `throw userException` will be executed in parallel with some other thread from the strategy, but it doesn't seem to cause problems

eupp:
> However, the actual throw userException will be executed in parallel with some other thread from the strategy, but it doesn't seem to cause problems

Yes, I think `onThreadRunException` is injected as a last instruction in `Thread::run`, so there should be no problems.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -690,6 +690,12 @@ abstract class ManagedStrategy(
         } else {
             // re-throw any non-internal exception,
             // so it will be treated as the final result of `Thread::run`.
+            disableAnalysis()
+            Logger.info { "Exception was thrown in user thread:" }
```
eupp, line 694: Let's also add thread name into the message.

eupp (changes requested):

eupp (approved):

Issue #593, opened 2025-03-29
Object numeration is non-deterministic


zhelenskiy:
```kotlin
class DoubleAccumulatorRandomTest : RandomTests() {
    @Operation
    fun operation() = DoubleAccumulator(operator(), 0.0)
        .apply { repeat(1) { accumulate(1.0) } }
        .get()

    private fun operator(): DoubleBinaryOperator = listOf(DoubleBinaryOperator { x, y -> x + y }, DoubleBinaryOperator { x, y -> x + y })[0]
}
```

The following fails with a non-determinism error:
```cs
java.lang.IllegalStateException: Non-determinism found. Probably caused by non-deterministic code (WeakHashMap, Object.hashCode, etc).
== Reporting the first execution without execution trace ==
= Invalid execution results =
| ----------------------------------------- |
|        Thread 1        |     Thread 2     |
| ----------------------------------------- |
| operation(): 1.0 [0,1] | operation(): 1.0 |
| ----------------------------------------- |
```
The actual problem happens because of shifted object indices:
```java
| operation(): 1.0                                                                                                                                                |
|   operator(): DoubleAccumulatorRandomTest$$Lambda$108/0x000000b001212958#1 at DoubleAccumulatorRandomTest.operation(RandomTests.kt:280)                         |
|     Array<DoubleBinaryOperator>#2[0] = DoubleAccumulatorRandomTest$$Lambda$108/0x000000b001212958#1 at DoubleAccumulatorRandomTest.operator(RandomTests.kt:284) |
|     Array<DoubleBinaryOperator>#2[1] = DoubleAccumulatorRandomTest$$Lambda$109/0x000000b001213bb8#1 at DoubleAccumulatorRandomTest.operator(RandomTests.kt:284) |
|     ArrayList#1.get(0): DoubleAccumulatorRandomTest$$Lambda$108/0x000000b001212958#1 at DoubleAccumulatorRandomTest.operator(RandomTests.kt:284)                |
|       Array<DoubleBinaryOperator>#2[0] ➜ DoubleAccumulatorRandomTest$$Lambda$108/0x000000b001212958#1 at Arrays$ArrayList.get(Arrays.java:4165)                 |
|   $this$operation_u24lambda_u241.accumulate(1.0) at DoubleAccumulatorRandomTest.operation(RandomTests.kt:281)                                                   |
|     cells ➜ null at DoubleAccumulator.accumulate(DoubleAccumulator.java:110)                                                                                    |
[...]

Issue #575, opened 2025-03-14
Refactor `TraceCollector`, moving all the filtering logic to `TraceReporter`


ndkoval:

eupp:
Migrated to YT

Issue #572, opened 2025-03-14
Refactor the trace collection logic


ndkoval:
- [ ] Actor tracepoints in the trace should be method call tracepoints  
- [ ] Refactor `TraceReporter`
- [ ] Refactor `TraceCollector`, moving all the filtering logic to `TraceReporter`

eupp:
Migrated to YT

PR #612, opened 2025-04-10
Improve the `ATOMIC` code section semantics

eupp:


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -386,7 +386,7 @@ class IncorrectHashmapRunConcurrentRepresentationTest : BaseRunConcurrentReprese
 class ThreadPoolRunConcurrentRepresentationTest : BaseRunConcurrentRepresentationTest<Unit>("run_concurrent_test/thread_pool/thread_pool") {
     // TODO: check why flakiness appears in this test
     override val isFlakyTest: Boolean
-        get() = true
+        get() = false
```
ndkoval, line 389: Should we remove the override?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ParkingTracker.kt
```diff
@@ -36,7 +36,7 @@ interface ParkingTracker {
      * @return true if the thread should continue waiting,
      *   false if it was already unparked.
      */
-    fun waitUnpark(threadId: Int): Boolean
+    fun waitUnpark(threadId: Int, allowSpuriousWakeUp: Boolean): Boolean
```
ndkoval, line 39: This is an elegant solution 🔥

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategyGuarantee.kt
```diff
@@ -77,7 +77,35 @@ class ManagedStrategyGuarantee private constructor(
...
+    className.startsWith("java.util.concurrent.ForkJoinPool")
+
 internal enum class ManagedGuaranteeType {
     IGNORE,
-    ATOMIC
-}
\ No newline at end of file
+    ATOMIC,
+    SILENT,
+    SILENT_NESTED,
```
ndkoval, line 106: What is the difference between "silent" and "silent nested"? Could you please document the modes?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
```diff
@@ -67,6 +67,10 @@ sealed class BlockingReason {
     data class  ThreadJoin(val joinedThreadId: ThreadId) : BlockingReason()
 }
 
+fun BlockingReason.allowsSpuriousUnblocking(): Boolean =
+    this is BlockingReason.Parked       ||
```
ndkoval, line 71: Is that always the case?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -162,7 +158,14 @@ abstract class ManagedStrategy(
...
     // TODO: extract into separate class
-    private var shadowStack = mutableThreadMapOf<ArrayList<ShadowStackFrame>>()
+    private val shadowStack = mutableThreadMapOf<ArrayList<ShadowStackFrame>>()
+
+    /**
+     * For each thread, stores a stack of entered method guarantees sections.
+     */
+    // TODO: unify with `shadowStack`
+    // TODO: handle coroutine resumptions (i.e., unify with `suspendedFunctionsStack`)
+    private val methodGuaranteesStack = mutableThreadMapOf<MutableList<ManagedGuaranteeType?>>()
```
ndkoval, line 168: How does maintaining this additional data structure affect performance?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategyGuarantee.kt
eupp: Added documentation that explains the difference between various sections.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: If there will be no flakiness on CI anymore, then yes : )

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Locally, when comparing with `develop`, there is no difference.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
ndkoval: Got it. Then I have two separate questions:

1. Can we rename it to `allowsSpuriousWakeUps`
2. How can we leave a live lock? (see the line below: `this is BlockingReason.LiveLocked`)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -162,7 +157,14 @@ abstract class ManagedStrategy(
...
     // TODO: handle coroutine resumptions (i.e., unify with `suspendedFunctionsStack`)
     // TODO: extract into separate class
-    private var shadowStack = mutableThreadMapOf<ArrayList<ShadowStackFrame>>()
+    private val shadowStack = mutableThreadMapOf<ArrayList<ShadowStackFrame>>()
+
+    /**
+     * For each thread, stores a stack of entered analysis sections.
+     */
+    // TODO: unify with `shadowStack`
+    // TODO: handle coroutine resumptions (i.e., unify with `suspendedFunctionsStack`)
```
ndkoval, line 166: Please fix the todo.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -571,9 +576,18 @@ abstract class ManagedStrategy(
...
             return nextThread
         }
-        // otherwise exit if the thread switch is optional, or all threads are finished
-        if (!mustSwitch || threadScheduler.areAllThreadsFinished()) {
-           return iThread
+        // exit if all threads are finished
+        if (threadScheduler.areAllThreadsFinished()) {
+            return iThread
+        }
+        // try to unpark some parked thread
```
ndkoval, line 583: Why should Lincheck unpark a parked thread?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -582,6 +596,10 @@ abstract class ManagedStrategy(
         if (suspendedThread != null) {
            return suspendedThread
         }
+        // if the thread switch is optional, continue with the current thread
+        if (!mustSwitch) {
```
ndkoval, line 600: Why is this check necessary?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -920,9 +941,15 @@ abstract class ManagedStrategy(
...
         newSwitchPoint(iThread, codeLocation, tracePoint)
         parkingTracker.park(iThread)
-        while (parkingTracker.waitUnpark(iThread)) {
+        // forbid spurious wake-ups if inside silent sections
+        var allowSpuriousWakeUp = shouldAllowSpuriousUnpark(iThread, codeLocation)
+        while (parkingTracker.waitUnpark(iThread, allowSpuriousWakeUp)) {
             // switch to another thread and wait till an unpark event happens
-            switchCurrentThread(iThread, BlockingReason.Parked)
+            val switchHappened = switchCurrentThread(iThread, BlockingReason.Parked)
+            // if the thread cannot switch, then allow spurious wake-up
```
ndkoval, line 949: Why should Lincheck allow spurious wake-ups here? Should the execution finish with a deadlock instead?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -942,6 +969,22 @@ abstract class ManagedStrategy(
         }
     }
 
+    private fun shouldAllowSpuriousUnpark(threadId: ThreadId, codeLocation: Int): Boolean {
+        val stackTraceElement = CodeLocations.stackTrace(codeLocation)
+        val analysisSectionStack = this.analysisSectionStack[threadId]!!
+        // TODO: refactor, track LockSupport.park directly instead
```
ndkoval, line 975: Please move it to the `park` function in the strategy

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1471,14 +1487,11 @@ abstract class ManagedStrategy(
...
+        if (section == AnalysisSectionType.NORMAL) {
             loopDetector.beforeMethodCall(codeLocation, params)
         }
-        // if the method is atomic or should be ignored, then we enter an ignored section
-        if (guarantee == ManagedGuaranteeType.IGNORE ||
-            guarantee == ManagedGuaranteeType.ATOMIC) {
-            enterIgnoredSection()
-        }
+        // if the method has certain guarantees, enter the corresponding section
+        enterAnalysisSection(threadId, section)
```
ndkoval, line 1494: What is `section` here? Please explain in the comments.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1608,6 +1615,71 @@ abstract class ManagedStrategy(
...
+            // TODO: atomic should have different semantics compared to ignored
+            section == AnalysisSectionType.ATOMIC
+        ) {
+            enterIgnoredSection()
+        }
+    }
+
+    private fun leaveAnalysisSection(threadId: ThreadId, section: AnalysisSectionType) {
+        if (section == AnalysisSectionType.IGNORE ||
+            // TODO: atomic should have different semantics compared to ignored
```
ndkoval, line 1668: What do you mean?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1608,6 +1615,71 @@ abstract class ManagedStrategy(
         }
     }
 
+    private fun methodAnalysisSectionType(
+        owner: Any?,
+        className: String,
+        methodName: String,
+        atomicMethodDescriptor: AtomicMethodDescriptor?,
+        deterministicMethodDescriptor: DeterministicMethodDescriptor<*, *>?,
```
ndkoval, line 1623: What is `DeterministicMethodDescriptor`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -11,8 +11,79 @@
...
+ * Local sections are required to support methods like `ConcurrentHashMap.computeIfAbsent(key, lambda)`.
+ * The method `computeIfAbsent` itself can be trusted and
+ * not analyzed by the framework in the user code by default.
+ * However, the user-provided lambda is not trusted and needs to be analyzed.
+ *
+ * Thus, `computeIfAbsent` can be put into a (local, non-propagating) silent section.
+ * As such, code inside `computeIfAbsent` will be muted,
+ * but if this method calls the provided lambda, it still will be analyzed fully.
+ *
+ * Local sections:
```
ndkoval, line 65: Please move these comments to the listing of the enum items below

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -93,4 +164,26 @@ internal inline fun <R> runOutsideIgnoredSection(block: () -> R): R {
     } finally {
         descriptor.restoreIgnoredSectionDepth(depth)
     }
-}
\ No newline at end of file
+}
+
+@Suppress("UNUSED_PARAMETER")
+internal fun getDefaultSilentSectionType(className: String, methodName: String): AnalysisSectionType? {
```
ndkoval, line 170: How do we decide when to enter and leave a silent section? 

Besides, I would place atomic and silent section declarations somewhere in a single place in a separate file, so one can easily observe them.

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked.txt
```diff
@@ -9,18 +9,11 @@ The following interleaving leads to the error:
...
-|             blocker.isReleasable(): false at ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3465)                                                   |
-|             blocker.block() at ForkJoinPool.unmanagedBlock(ForkJoinPool.java:3465)                                                                 |
-|               /* The following events repeat infinitely: */                                                                                        |
-|           ┌╶> isReleasable(): false at AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:506)                         |
-|           |   LockSupport.park() at AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:506)                            |
-|           └╶╶ switch (reason: active lock detected)                                                                                                |
+|           /* The following events repeat infinitely: */                                                                                            |
```
ndkoval, line 14: There should be no spin loop- the code should be suspended on the first attempt.

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked.txt
```diff
@@ -37,24 +30,12 @@ Detailed trace:
...
-|           |     status ➜ 3 at AbstractQueuedSynchronizer$ConditionNode.isReleasable(AbstractQueuedSynchronizer.java:502)                                     |
-|           |   LockSupport.park() at AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:506)                                      |
-|           |     PARK at LockSupport.park(LockSupport.java:341)                                                                                               |
-|           └╶╶ switch (reason: active lock detected)                                                                                                          |
+|           /* The following events repeat infinitely: */                                                                                                      |
+|       ┌╶> LockSupport.park() at AbstractQueuedSynchronizer$ConditionNode.block(AbstractQueuedSynchronizer.java:506)                                          |
+|       |     PARK at LockSupport.park(LockSupport.java:341)                                                                                                   |
```
ndkoval, line 38: PARK at LockSupport.park(LockSupport.java:341) -- this event should not be listed in the execution trace

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_trace_debugger_jdk13.txt
```diff
@@ -1 +1,66 @@
...
+| $this_runCatching.block() at BaseRunConcurrentRepresentationTest.testRunWithModelChecker$lambda$1$lambda$0(RunConcurrentRepresentationTests.kt:47) |
+|   block() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:96)                                        |
+|     q.take() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:101)                                    |
+|       lock.lockInterruptibly() at ArrayBlockingQueue.take(ArrayBlockingQueue.java:414)                                                             |
+|       count ➜ 0 at ArrayBlockingQueue.take(ArrayBlockingQueue.java:416)                                                                            |
+|       ConditionObject#1.await() at ArrayBlockingQueue.take(ArrayBlockingQueue.java:417)                                                            |
+|         addConditionWaiter(): Node#1 at AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2077)                     |
```
ndkoval, line 13: Should AQS detail be in a silent section?

src/jvm/test/resources/expected_logs/run_concurrent_test/thread_pool/thread_pool_jdk8.txt
ndkoval, line 1: There are still a lot of events in the trace, should we minimize them?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: It is related to native method calls support by `@zhelenskiy`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Unfortunately, this is not that simple : (
I've already tried this refactoring before in another branch, and it is more complicated:
https://github.com/JetBrains/lincheck/tree/eupp/shadow-stack

This is not a new problem introduced in this PR though. It was existing before.

The problem is that we have several stack-like data structures, that manage various information about program stack trace:
- `callStackTrace` stores `CallStackTraceElement` (essentially just `MethodCallTracePoint`) of the methods.
- `suspendedFunctionsStack` stores `CallStackTraceElement` of suspended methods.
- `shadowStack` stores local variables info

Now this PR just adds another structure:
-  `analysisSectionStack` to store current analysis section.

The problem is that because the info is stored in different properties, it leads to code duplication (and bugs!), because in essence we need to maintain all of them consistently with respect to actual program stack. 

But as I said, unifying all of this is a big change on itself, and I propose to do this in a separate PR.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -463,7 +466,9 @@ abstract class ManagedStrategy(
         if (!loopDetector.replayModeEnabled) {
             loopDetector.onNextExecutionPoint(codeLocation)
         }
-        traceCollector?.passCodeLocation(tracePoint)
+        if (!inSilentSection(threadId)) {
+            traceCollector?.passCodeLocation(tracePoint)
```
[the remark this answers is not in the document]
eupp: After https://github.com/JetBrains/lincheck/issues/575 we can. Now, if we collect them, they would appear in the trace, and the trace will not be shorten.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Suppose a thread is parked inside a silent section, and all other threads are blocked. 
Shouldn't the strategy then attempt to unpark it spuriously?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: It was in the code before, I only moved it below, after handling other cases first.

In general, this case occurs when strategy attempts a to make a thread switch, all other threads are blocked or finished, but the thread we are switching from is not blocked. Thus it is fine to continue executing the current thread.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: There is a case I mentioned above.

> Suppose a thread is parked inside a silent section, and all other threads are blocked.
> Shouldn't the strategy then attempt to unpark it spuriously?

If we don't want to unpark in this case, then perhaps it is safe to revert this.

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_trace_debugger_jdk13.txt
eupp: No I think it shouldn't. I haven't updated this file yet, on it.

src/jvm/test/resources/expected_logs/run_concurrent_test/thread_pool/thread_pool_jdk8.txt
eupp: There are two reasons that lead to more events appearing in the trace:

1. `switch` is inserted inside `LockSupport.park()`. Just adding `LockSupport.park()` to the silent section does not work though. I'll investigate it. 

2. All other switch points are forced by the `park` inside silent sections. Their whole call stack is displayed in this case. We can shorten the call stack by printing `...` instead, but it is easier to do on trace post-processing stage, after we implement: https://github.com/JetBrains/lincheck/issues/572 .

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Before this PR, `ATOMIC` and `IGNORE` sections were handled similarly, by just entering ingored section and disabling the analysis completely. 

However, I think `ATOMIC` should be handled different. It should be similar to `SILENT` (so analysis is still run inside it, but with no switch or trace points created). But unlike `SILENT` section, in `ATOMIC` section there should be no thread switches at all. 

If some thread switch is enforced inside `ATOMIC` section because of blocking, it means that the method is not truly atomic, and shouldn't be marked as `ATOMIC`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
eupp: 1. Yes, I'll rename. 
2. No, live lock does not block currently, you right. I'll remove it.

eupp: Oh, no, regarding (2), currently the code is:

https://github.com/JetBrains/lincheck/blob/7f4d8d4a418c8b6c6d8039ee7b79afa49ab08778/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt#L536

So in essence I just extended it to `BlockingReason.Parked`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: There is already a comment:

https://github.com/JetBrains/lincheck/blob/dba37f265dc8d18c03d44221ae040ea4a4354f19/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt#L1455

And I renamed `section` to `methodSection` for clarity

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
eupp: > Besides, I would place atomic and silent section declarations somewhere in a single place in a separate file, so one can easily observe them.

Ok, I agree we should collect this logic in a single place. 
I created a separate issue for this refactoring: https://github.com/JetBrains/lincheck/issues/616

We can do it separately so not to postpone this PR. 
Also, we already briefly discussed this problem with @bbrockbernd , so perhaps he can do this.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
eupp: I reverted it

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Nevermind, reverted it.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Reverted

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Reverted

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: Unfortunately, the test is still flaky due to accidental changes in the object numeration:
https://github.com/JetBrains/lincheck/issues/593

We should investigate and fix it separately.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -437,7 +439,10 @@ class CoroutinesRunConcurrentRepresentationTest : BaseRunConcurrentRepresentatio
     @Before
     fun setUp() {
         assumeFalse(isInTraceDebuggerMode) // unstable hash-code
-        assumeFalse(isJdk8) // TODO: investigate why test is unstable on JDK8
+        // TODO: investigate why test is unstable on these JDKs
```
ndkoval, line 442: Note: do not forget to create an issue if we decide not to fix this in this PR

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_jdk11.txt
```diff
@@ -30,15 +29,9 @@ Detailed trace:
...
-|     |     PARK at LockSupport.park(LockSupport.java:194)                                                                                           |
-|     |     LockSupport.setBlocker(Thread#1, null) at LockSupport.park(LockSupport.java:195)                                                         |
-|     |       Unsafe#1.putObject(Thread#1, 76, null) at LockSupport.setBlocker(LockSupport.java:144)                                                 |
-|     |   NonfairSync#1.isOnSyncQueue(Node#1): false at AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2080)       |
-|     |     node.waitStatus ➜ -2 at AbstractQueuedSynchronizer.isOnSyncQueue(AbstractQueuedSynchronizer.java:1673)                                   |
-|     └╶╶ switch (reason: active lock detected)                                                                                                      |
+|         LockSupport.park(ConditionObject#1) at AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2081)              |
```
ndkoval, line 32: Consider creating an issue to not show `LockSupport` internals

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2172,12 +2249,16 @@ abstract class ManagedStrategy(
...
-        fun passCodeLocation(tracePoint: TracePoint?) {
-            if (tracePoint !is SectionDelimiterTracePoint && tracePoint?.isActorMethodCallTracePoint() == false) 
+        fun passCodeLocation(tracePoint: TracePoint?, isSilentTracePoint: Boolean = false) {
+            if (tracePoint !is SectionDelimiterTracePoint && tracePoint?.isActorMethodCallTracePoint() == false) {
                 checkActiveLockDetected()
+            }
             // tracePoint can be null here if trace is not available, e.g. in case of suspension
             if (tracePoint != null) {
-                _trace += tracePoint
+                // do not add the trace point to the trace if inside the silent section
```
ndkoval, line 2258: Let's track everything in silent sections -- we can filter out unnecessary events at a later stage

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -11,8 +11,91 @@
...
+     */
+    SILENT_PROPAGATING,
+
+    /**
+     * Code inside an atomic section behaves like an atomic indivisible operation.
+     * Technically, the atomic section is a stronger version of the nested silent section,
+     * with an additional guarantee that thread switch points cannot occur inside an atomic section at all.
+     * TODO: "no-switch-points inside atomic section" guarantee is not checked currently.
+     */
+    ATOMIC,
```
ndkoval, line 84: Is there any difference between atomic and silent? Do we need both?

ndkoval: The only (and maybe important) difference I see is that for atomic sections, it's fine not to transform the corresponding byte-code -- we just switch off the analysis

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -537,33 +537,59 @@ internal class ModelCheckingMonitorTracker : MonitorTracker {
 
 class ModelCheckingParkingTracker(val allowSpuriousWakeUps: Boolean = false) : ParkingTracker {
 
-    private val parked = mutableThreadMapOf<Boolean>()
+    private object PERMIT
+
+    private enum class State {
+        UNPARKED, PARKED, INTERRUPTED,
```
ndkoval, line 543: let's add an additional state instead of adding permit and draw a state diagram for clarity

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: There is an issue for this problem: https://github.com/JetBrains/lincheck/issues/593

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_jdk11.txt
eupp: https://github.com/JetBrains/lincheck/issues/627

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
eupp: > it's fine not to transform the corresponding byte-code -- we just switch off the analysis

I think we still might want to track some events inside atomic sections --- that is potentially blocking events (like `lock`, `park`, etc). This would be necessary to check that these events do not force blocking and thus a thread switch inside an atomic section (to verify that this section is indeed atomic and does not have any thread switches at all).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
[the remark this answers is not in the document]
eupp: Roll-backed these changes. All filtering will be done on trace post processing phase later.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: https://github.com/JetBrains/lincheck/issues/628

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
ndkoval: Let @bbrockbernd do this, sure

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -537,33 +537,82 @@ internal class ModelCheckingMonitorTracker : MonitorTracker {
...
+        PARKED,
+        PERMITTED,
+        INTERRUPTED,
+    }
+
+    private class Handle {
+        var state: State = State.UNPARKED
+    }
+
+    private val handle = mutableMapOf<ThreadId, Handle>()
```
ndkoval, line 573: Consider renaming to `threadStates`, storing `State` instead of `Handle`, and getting rid of the `Handle` abstraction.

src/jvm/test/resources/expected_logs/method_handles/find_special_trace_debugger_jdk15.txt
```diff
@@ -26,11 +26,13 @@ Detailed trace:
 |         value ➜ 0 at CounterDerived.increment(MethodHandlesLookupTests.kt:178)                                                                                                  |
 |         value = 2 at CounterDerived.increment(MethodHandlesLookupTests.kt:178)                                                                                                  |
 |       methodHandle.invoke(CounterDerived#1) at MethodHandlesFindSpecialRepresentationTest.doTest(MethodHandlesLookupTests.kt:30)                                                |
-|         CounterDerived#1.value ➜ 2 at Counter.increment(MethodHandlesLookupTests.kt:158)                                                                                        |
-|         CounterDerived#1.value = 3 at Counter.increment(MethodHandlesLookupTests.kt:158)                                                                                        |
+|         MethodHandle.linkToSpecial(CounterDerived#1, MemberName#1) at DirectMethodHandle$Holder.invokeSpecial(DirectMethodHandle$Holder:0)                                      |
```
ndkoval, line 29: Why was this event added?

ndkoval (approved):
Please fix the remaining minor comments and merge the PR 🚀

Issue #613, opened 2025-04-10
`ModelCheckingStrategy::chooseThread` interleaves threads not on the switch positions


dmitrii-artuhov:
```kotlin
// ModelCheckingStrategy.kt

fun chooseThread(iThread: Int): Int =
    if (nextThreadToSwitch.hasNext()) {
        check(this.executionPosition == -1 || this.executionPosition in this.switchPositions) // this check must always pass
        // Use the predefined choice.
        nextThreadToSwitch.next()
    } else {
        // There is no predefined choice.
        // This can happen if there were forced thread switches after the last predefined one
        // (e.g., thread end, coroutine suspension, acquiring an already acquired lock or monitor.wait).
        // We use a deterministic random here to choose the next thread.
        lastNotInitializedNodeChoices =
            null // end of execution position choosing initialization because of new switch
        switchableThreads(iThread).random(interleavingFinishingRandom)
    }
```

The `check` added to the `ModelCheckingStrategy::chooseThread` method should always pass, however, there are tests for which this check fails on some invocation:
- `JavaAPITest` (gpmc)
- `KotlinAPITest` (gpmc)
- `IncorrectConcurrentLinkedDequeRunConcurrentRepresentationTest` (gpmc)
- `MultipleSuspensionPointsTraceRepresentationTest` (data structures)
- `SingleSuspensionPointTraceRepresentationTest` (data structures)

I have seen this check failing when jit was intrinsifying (substituting with predefined asm code) methods like `java.util.Arrays::copyOf(...)`, like in these PRs: https://github.com/JetBrains/lincheck/pull/578, https://github.com/JetBrains/lincheck/pull/531 ([this test](https://github.com/JetBrains/lincheck/pull/531/files#diff-ec19f7cefc20a19593ec83c10928c34b4aacd3c3da028bb2ef80c42ae403828dR62) was failing) which caused non-determinism between invocations. This non-determinism was causing `invalid thread switch attempt` error: the reason is that when replaying a prefix of some interleaving the switch attempt was performed earlier than some saved `switchPosition` of this prefix (basically when `executionPosition` was less than the corresponding `switchPositions[i]`), and on earlier switch there could be a case, when thread to switch to, saved in the interleaving prefix, has not been started yet. Such a situation leads to an invalid switch attempt. 

In the tests list there is also some non-determinism, but it does not cause any errors right now.

---
Might be related to https://github.com/JetBrains/lincheck/issues/470.

[...]

PR #619, opened 2025-04-15
Debugging utils for `ModelCheckerStrategy`

dmitrii-artuhov:
This PR attempts to accumulate additional `check`'s for `ModelCheckingStartegy` and mark tests for which these checks fail.

Associated issues:
- [x] https://github.com/JetBrains/lincheck/issues/613
- [x] [Resolved, see comment below] lincheck internal errors are ignored and do not appear in the resulting trace for tests:
  - `KotlinAPITest`
  - `JavaAPITest`
  - `SingleSuspensionPointTraceRepresentationTest`
  - `IncorrectConcurrentLinkedDequeRunConcurrentRepresentationTest`

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
@@ -311,20 +311,21 @@ internal class ModelCheckingStrategy(
 
         fun chooseThread(iThread: Int): Int =
             if (currentInterleavingPosition < threadSwitchChoices.size) {
-                check(
-                    // no thread switch happened yet, initial thread id will be returned
-                    executionPosition == -1 ||
-                    // loop detector fully controls 'switchPositions' by itself, thus, 'executionPosition' is always 0, but 'threadSwitchChoices' are still valid
-                    (executionPosition == 0 && loopDetector.replayModeEnabled) ||
-                    // 'threadSwitchChoices.size == switchPositions.size + 1', thus, we subtract 1 from 'currentInterleavingPosition'
-                    // (indexing is correct, because if 'currentInterleavingPosition' is 0, then 'executionPosition == -1' would hold, and we would exit disjunction earlier)
-                    executionPosition == switchPositions[currentInterleavingPosition - 1]
-                ) {
-                    """
-                        Attempt to switch thread on execution position which does not correspond to any saved switch position.
-                        Execution position: $executionPosition, switch positions: $switchPositions.
-                    """.trimIndent()
-                }
+                // TODO: causes TL on CI
+                // check(
+                //     // no thread switch happened yet, initial thread id will be returned
+                //     executionPosition == -1 ||
+                //     // loop detector fully controls 'switchPositions' by itself, thus, 'executionPosition' is always 0, but 'threadSwitchChoices' are still valid
+                //     (executionPosition == 0 && loopDetector.replayModeEnabled) ||
+                //     // 'threadSwitchChoices.size == switchPositions.size + 1', thus, we subtract 1 from 'currentInterleavingPosition'
+                //     // (indexing is correct, because if 'currentInterleavingPosition' is 0, then 'executionPosition == -1' would hold, and we would exit disjunction earlier)
+                //     executionPosition == switchPositions[currentInterleavingPosition - 1]
+                // ) {
+                //     """
+                //         Attempt to switch thread on execution position which does not correspond to any saved switch position.
+                //         Execution position: $executionPosition, switch positions: $switchPositions.
+                //     """.trimIndent()
+                // }
                 // Use the predefined choice.
                 threadSwitchChoices[currentInterleavingPosition++]
             } else {
@@ -350,6 +351,7 @@ internal class ModelCheckingStrategy(
                 // Add a new thread choosing node corresponding to the switch at the current execution position.
                 if (lastNotInitializedNodeChoices == null) return
                 val availableThreads = switchableThreads(iThread)
+                // TODO: causes TL on CI
                 // val lastThreadSwitchChoice = threadSwitchChoices.lastOrNull()
                 // check(lastThreadSwitchChoice !in availableThreads) {
                 //     """
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
@@ -311,20 +311,20 @@ internal class ModelCheckingStrategy(
 
         fun chooseThread(iThread: Int): Int =
             if (currentInterleavingPosition < threadSwitchChoices.size) {
-//                check(
-//                    // no thread switch happened yet, initial thread id will be returned
-//                    executionPosition == -1 ||
-//                    // loop detector fully controls 'switchPositions' by itself, thus, 'executionPosition' is always 0, but 'threadSwitchChoices' are still valid
-//                    (executionPosition == 0 && loopDetector.replayModeEnabled) ||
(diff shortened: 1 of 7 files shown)


ndkoval:
Hi @dmitrii-artuhov! Could you please clarify why the listed tests fail?

dmitrii-artuhov:
> Hi @dmitrii-artuhov! Could you please clarify why the listed tests fail?

I investigated the problem a little more. These tests turned out to be working properly, however one of the checks should be rewritten as:
```kotlin
check(executionPosition == -1 || executionPosition in switchPositions || (loopDetector.replayModeEnabled && executionPosition == 0)) {
  """
    Attempt to switch thread on execution position which does not correspond to any saved switch position.
    Execution position: $executionPosition, switch positions: $switchPositions.
  """.trimIndent()
}
```
Because `executionPosition` is always 0, when loop detector controls thread switches.

---
These checks useful when some non-determinism is happening in the test (like it was with `@IntrinsicCandidate` annotated methods). But these checks slow down the execution quite a lot (x1.5 total execution time locally).

From this PR I would merge the pretty-printing for the interleaving tree. And the checks I would either comment-out and merge like that or only enabled them for local testing (via separate task or flag). What do you think?

eupp:
> But these checks slow down the execution quite a lot (x1.5 total execution time locally).

I suspect the performance degradation is due to this particular check: 
```kotlin
executionPosition in switchPositions
```

This is a linear lookup in a list that is typically quite large. 

I believe the invariant that you really want to check is the following: 
```kotlin
check(executionPosition == -1 || executionPosition == switchPositions[currentInterleavingPosition] || (loopDetector.replayModeEnabled && executionPosition == 0))
```

where `currentInterleavingPosition: Int` is the current position in the interleaving being explored. 

So instead of storing field `nextThreadToSwitch: Iterator<Int>`, initialized as:
```kotlin
nextThreadToSwitch = threadSwitchChoices.iterator()
```

I would propose to store `currentInterleavingPosition: Int`, then you can obtain both `nextThreadToSwitch` and `nextSwitchPosition`:
```kotlin
nextThreadToSwitch == threadSwitchChoices[currentInterleavingPosition`
nextSwitchPosition == switchPositions[currentInterleavingPosition]
```

Using just an index in the list, instead of the iterator, would also simplify the code and make it easier to understand. 

Besides, would be useful to add a check for invariant `switchPositions.size == threadSwitchChoices.size` in the constructor of `Interleaving` class.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -306,9 +338,19 @@ internal class ModelCheckingStrategy(
          */
         fun newExecutionPosition(iThread: Int) {
             executionPosition++
-            if (executionPosition > switchPositions.lastOrNull() ?: -1) {
+            if (executionPosition > (switchPositions.lastOrNull() ?: -1)) {
                 // Add a new thread choosing node corresponding to the switch at the current execution position.
-                lastNotInitializedNodeChoices?.add(Choice(ThreadChoosingNode(switchableThreads(iThread)), executionPosition))
+                lastNotInitializedNodeChoices?.apply {
+                    switchableThreads(iThread).let {
```
eupp, line 344: Using scope functions (`apply` and `let`) here makes the code less clear and harder to understand. 
Please simplify this code by avoiding scope functions.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -260,14 +286,14 @@ internal class ModelCheckingStrategy(
     ) {
         private var lastNotInitializedNode: SwitchChoosingNode? = initialLastNotInitializedNode
         private lateinit var interleavingFinishingRandom: Random
-        private lateinit var nextThreadToSwitch: Iterator<Int>
+        private var currentInterleavingPosition = 0
```
eupp, line 289: Would it make sense to initialize to `-1` instead?
`0` is technically a valid position in a list.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -260,14 +286,14 @@ internal class ModelCheckingStrategy(
...
         private var lastNotInitializedNodeChoices: MutableList<Choice>? = null
         private var executionPosition: Int = 0
 
         fun initialize() {
             executionPosition = -1 // the first execution position will be zero
             interleavingFinishingRandom = Random(2) // random with a constant seed
-            nextThreadToSwitch = threadSwitchChoices.iterator()
+            currentInterleavingPosition = 0
```
eupp, line 296: Could we initialize it as:
```kotlin
// set to -1 if loop detector controls thread switches
currentInterleavingPosition = if (!loopDetector.replayModeEnabled) 0 else -1 
```

Then it will also simplify the checked invariant later in `chooseThread`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -284,9 +310,15 @@ internal class ModelCheckingStrategy(
         }
 
         fun chooseThread(iThread: Int): Int =
-            if (nextThreadToSwitch.hasNext()) {
+            if (currentInterleavingPosition < threadSwitchChoices.size) {
+                check(executionPosition == -1 || (executionPosition == 0 && loopDetector.replayModeEnabled) || executionPosition == switchPositions[currentInterleavingPosition - 1]) {
```
eupp, line 314: Perhaps makes sense to handle case `executionPosition == -1` separately. 
In this case the loop detector controls the scheduling, so what should `chooseThread` return?
I think returning just `threadSwitchChoices[currentInterleavingPosition++]` is incorrect?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -306,9 +338,17 @@ internal class ModelCheckingStrategy(
          */
         fun newExecutionPosition(iThread: Int) {
             executionPosition++
-            if (executionPosition > switchPositions.lastOrNull() ?: -1) {
+            if (executionPosition > (switchPositions.lastOrNull() ?: -1)) {
                 // Add a new thread choosing node corresponding to the switch at the current execution position.
-                lastNotInitializedNodeChoices?.add(Choice(ThreadChoosingNode(switchableThreads(iThread)), executionPosition))
+                if (lastNotInitializedNodeChoices == null) return
+                val availableThreads = switchableThreads(iThread)
+                check(threadSwitchChoices.lastOrNull() !in availableThreads) {
```
eupp, line 345: adding: 
```
val lastThreadSwitchChoice = threadSwitchChoices.lastOrNull()
check(lastThreadSwitchChoice !in availableThreads)
```

would clarify the code a bit.

eupp (changes requested):
Added just a few final comments.

dmitrii-artuhov:
Expected to pass CI after merge of #531 (which contains fixes for non-determinism)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
dmitrii-artuhov: this would only change the way nextThread is returned in chooseThread method: prefix ++ instead of suffix one.
I think zero as initial value is more understandable

eupp (approved):

PR #620, opened 2025-04-15
Fix the `findOwnerName` semantics 

ndkoval:


ivandev0:
I checked this fix with the Kotlin project. It helps to improve instrumentation performance, but not that much. After this fix is applied, most of the time is spent in `ensureClassHierarchyIsTransformed` and lately in `retransformClasses`. Anyway, the fix is useful and we should try to finish it.

eupp:
I made a pass over the PR. 
@ivandev0 @ndkoval should we merge it?

ivandev0 (approved):
Looks fine to me. I checked it again to be sure, and it helps, but again not as much as I hoped. Still, representation improved and that alone is a nice change.

ndkoval:
@eupp could you please apply a similar logic to get rid of using `FieldSearchHelper.findFinalFieldWithOwner`in `AtomicReferenceNames`?

eupp:
> @eupp could you please apply a similar logic to get rid of using FieldSearchHelper.findFinalFieldWithOwnerin AtomicReferenceNames?

Done, `FieldSearchHelper` is completely removed from the code.

Let's merge?

PR #621, opened 2025-04-17
Improve suspend function trace representation

bbrockbernd:
- Method call trace point has `isSuspendMethodCall` property that can be used to pass this info to idea plugin.
- `<cont>` parameters are removed from trace representation.
- `label` reads and writes during suspend fun state changes are not recorded.
- `L$0` L$1` etc.. for stack variable recovery after resumption are als not recorded.

Note: `coroutines.txt` has more changes than what is mentioned above this is due to the test being marked as flaky which prevented the txt from being updated -- on current develop branch the same changes would have happened.


src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -239,6 +240,10 @@ internal class MethodCallTracePoint(
     var thrownException: Throwable? = null
     var parameters: List<String>? = null
     var parameterTypes: List<String>? = null
+    
+    var isSuspendMethodCall = false 
+        private set
```
ndkoval, line 245: Can we compute it in the constructor?

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -323,8 +327,14 @@ internal class MethodCallTracePoint(
     }
 
     fun initializeParameters(parameters: List<String>, parameterTypes: List<String>) {
-        this.parameters = parameters
-        this.parameterTypes = parameterTypes
+        if (parameters.lastOrNull() == ObjectLabelFactory.CONTINUATION_REPRESENTATION) {
```
ndkoval, line 330: 1. Should we always apply this transformation? 
2. Should we do it here, or should it be a part of the post-processing?

src/jvm/test/resources/expected_logs/multiple_suspension_points_channels.txt
```diff
@@ -10,116 +10,116 @@ The number next to an exception name helps you find its stack trace provided aft
...
+|                                                                                                                                                       | operation2(): IllegalStateException #2                                                                                                                |
+|                                                                                                                                                       |   channel1.receive(): 1 at MultipleSuspensionPointsChannelsTraceRepresentationTest.operation2(MultipleSuspensionsTests.kt:166)                        |
+|                                                                                                                                                       |     BufferedChannel.receive$suspendImpl(BufferedChannel#1): 1 at BufferedChannel.receive(BufferedChannel.kt:0)                                        |
```
ndkoval, line 18: Can we get rid of `$suspendImpl` function calls?

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
```diff
@@ -22,10 +22,10 @@ The following interleaving leads to the error:
 |   barStarted ➜ true at SuspendTraceReportingTest.foo(SuspendTraceReportingTest.kt:29)                                                                       |                                                                                                                                          |
 |   canEnterForbiddenBlock = true at SuspendTraceReportingTest.foo(SuspendTraceReportingTest.kt:29)                                                           |                                                                                                                                          |
 |   $i$a$-withLock$default-SuspendTraceReportingTest$foo$2.lock(null): COROUTINE_SUSPENDED at SuspendTraceReportingTest.foo(SuspendTraceReportingTest.kt:184) |                                                                                                                                          |
```
ndkoval, line 24: Please raise an issue for this weird function call

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
```diff
@@ -44,124 +44,124 @@ The following interleaving leads to the error:
...
+|                                                                                                                                                                                 |                       resumeMode ➜ 1 at CancellableContinuationImpl.isReusable(CancellableContinuationImpl.kt:141)                                                                     |
+|                                                                                                                                                                                 |                     detachChild$kotlinx_coroutines_core() at CancellableContinuationImpl.detachChildIfNonResuable(CancellableContinuationImpl.kt:565)                                  |
+|                                                                                                                                                                                 |                       getParentHandle(): ChildContinuation#1 at CancellableContinuationImpl.detachChild$kotlinx_coroutines_core(CancellableContinuationImpl.kt:572)                    |
```
ndkoval, line 134: Should we show `$kotlinx_coroutines_core` here?

bbrockbernd: Isn't this something that should be resolved with silent sections?

ndkoval: We still need to support analyzing coroutine internals

bbrockbernd:
> Can we get rid of $suspendImpl function calls?

Fixing this was easier to directly manipulate the trace graph `TraceNode`s instead of the provided trace object. Also, considering we want to start collecting the trace directly into this graph #575, I changed all compression logic to work directly on the tree structure. This significantly improved compression logic see `TraceCompression.kt`. Maybe @eupp can have a look there as he reviewed last refactor in this area as well.

src/jvm/test/resources/expected_logs/multiple_suspension_points_channels.txt
bbrockbernd: Done

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: Wether or not a function can be marked as suspend currently depends on it's parameters. These parameters are initialized after construction by `ManagedStrategy` calling `MethodCallTracePoint.initializeParameters`.

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
bbrockbernd: #625

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
eupp: I agree we should initialize all properties of all trace points in their constructors (and purge all `lateinit` properties), but it is probably worth doing in a separate PR.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -54,15 +54,14 @@ internal class TraceReporter(
...
-            .compressTrace()
-            .removeNestedThreadStartPoints()
 
         // Turn trace into graph which is List of sections. Where a section is a list of rootNodes (actors).
-        val traceGraph = traceToGraph(optimizedTrace)
-        graph = if (isGeneralPurposeModelCheckingScenario(failure.scenario)) removeGPMCLambda(traceGraph) else traceGraph
+        val traceGraph = traceToGraph(fixedTrace).compressTrace()
+
+        // Optimizes trace by combining trace points for synthetic field accesses etc..
+        val compressedTraceGraph = traceGraph.compressTrace()
```
eupp, line 62: `copressTrace()` is called twice. Is it the expected behavior?

eupp (changes requested):

eupp:
Now that we remove continuation parameter from the suspend function call in the trace, I wonder, are there any cases when it makes the trace less clear?

What I mean, is that when `resumeWith` is called on the continuation in the trace, it might be unclear to what suspension point (or `suspend fun`) it corresponds.

@ndkoval @bbrockbernd WDYT ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: #626

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
```diff
@@ -39,7 +39,7 @@ object ObjectLabelFactory {
         // simplified representation for Continuations
         // (we usually do not really care about details).
         if (any is Continuation<*>)
-            return "<cont>"
+            return CONTINUATION_REPRESENTATION
```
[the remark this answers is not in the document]
bbrockbernd: Or some other ideas
```
channel1.send(1) with(<continuation#2>) at MultipleSuspensionPointsChannelsTraceRepresentationTest.operation1(MultipleSuspensionsTests.kt:156)

```

```
channel1.send(1) [suspendable: Continuation#2] at MultipleSuspensionPointsChannelsTraceRepresentationTest.operation1(MultipleSuspensionsTests.kt:156)
```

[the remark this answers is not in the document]
eupp: I like the third option:

```kotlin
channel1.send(1) [suspendable: Continuation#2] at MultipleSuspensionPointsChannelsTraceRepresentationTest.operation1(MultipleSuspensionsTests.kt:156)
```

@ndkoval WDYT ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: > Should we always apply this transformation?

If you are asking if this check always correctly identifies suspend functions, the answer is no. Technically any function can provide a continuation as last parameter. If we want to check this more thoroughly we might have to use [reflection](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.reflect/-k-function/is-suspend.html). But I am open to suggestions!

> Should we do it here, or should it be a part of the post-processing?

Not sure why not, this is the point when it becomes clear the method call is suspend. Which should be a property of the method call. Other option would be to only set the `isSuspendMethodCall` and based on that boolean remove the last parameter in a later stage which seems cumbersome.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1078,7 +1079,7 @@ abstract class ManagedStrategy(
             return false
         }
         // Do not track accesses to untracked objects
-        if (!shouldTrackObjectAccess(obj)) {
+        if (!shouldTrackObjectAccess(obj) || isStackRecoveryFieldAccess(obj, fieldName)) {
```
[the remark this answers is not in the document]
bbrockbernd: Not sure if my solution is what you meant please have a look

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
bbrockbernd: Done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1120,7 +1121,7 @@ abstract class ManagedStrategy(
     /** Returns <code>true</code> if a switch point is created. */
     override fun beforeReadArrayElement(array: Any, index: Int, codeLocation: Int): Boolean = runInsideIgnoredSection {
         updateSnapshotOnArrayElementAccess(array, index)
-        if (!shouldTrackObjectAccess(array)) {
+        if (!shouldTrackFieldAccess(array, null)) {
```
eupp, line 1124: Could you please also add shouldTrackArrayAccess for uniformity.

eupp (approved):

PR #629, opened 2025-04-28
Fix timeout in deterministic time tests

eupp:
Tests `ActualTimeMillisTest` and `ActualTimeNanosTest` fail with:

```
= The execution has hung, see the thread dump =
```

instead of normal completion. 

The bug was missed by `AbstractDeterministicTest`. 

This PR fixes the tests by increasing the timeout for them.

### DIFF
--- a/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/AbstractDeterministicTest.kt
+++ b/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/AbstractDeterministicTest.kt
@@ -45,7 +45,7 @@ abstract class AbstractDeterministicTest {
             ModelCheckingOptions()
                 .actorsBefore(0)
                 .actorsAfter(0)
-                .iterations(1)
+                .iterations(30)
                 .threads(2)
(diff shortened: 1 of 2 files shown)


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/AbstractDeterministicTest.kt
```diff
@@ -45,7 +45,7 @@ abstract class AbstractDeterministicTest {
             ModelCheckingOptions()
                 .actorsBefore(0)
                 .actorsAfter(0)
-                .iterations(30)
+                .iterations(1)
```
zhelenskiy, line 48: Some bugs are only reproducible since approximately this number of iterations, so I would just override the value for exact tests using `customize`. And I would recommend setting it to at least `2`.

eupp: `iterations` parameter and data structures API shouldn't be used in these tests in the first place.

zhelenskiy (approved):

PR #645, opened 2025-05-13
Refactor loop detector should switch method call

eupp:
Small refactoring to clarify code in `ManagedStrategy::newSwitchPoint()`

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -449,17 +449,17 @@ abstract class ManagedStrategy(
         }
         // check we are in the right thread
         check(threadId == threadScheduler.scheduledThreadId)
-        // check if we need to switch
-        val shouldSwitch = when {
-            // check if a switch is required in replay mode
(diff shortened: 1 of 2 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -449,17 +449,17 @@ abstract class ManagedStrategy(
...
-            // check if a switch is required in replay mode
-            loopDetector.replayModeEnabled -> loopDetector.shouldSwitchInReplayMode()
+            // check if a switch is required by the loop detector replay mode
+            loopDetector.replayModeEnabled -> loopDetector.shouldSwitch()
             // do not make thread switches inside a silent section
             inSilentSection(threadId) -> false
-            // otherwise, as strategy if thread switch is needed
+            // otherwise, follow the strategy decision
             else -> shouldSwitch(threadId)
         }
```
dmitrii-artuhov, line 462: If `loopDetector.shouldSwitch()` is now with no side effects, then why not moving the whole `val shouldSwitch = ...` variable initialization after the `if (...) { processLoopDetectorDesicion(...) }` code (where the variable is used)?

dmitrii-artuhov (changes requested):

dmitrii-artuhov (approved):

Issue #542, opened 2025-02-20
Non-closed threadpool treated as inifinitely running


dmitrii-artuhov:
GPMC counts idle threads in threadpools as actively running and reports deadlock on them.

## Problem
Minimal reproducing test (note that writing explicit pool closing such as `pool.close()` or `pool.use {...}` solves the issue, but the gpmc still explores schedulings of the idle threads, which is probably not what we want)
```kotlin
@OptIn(ExperimentalModelCheckingAPI::class)
class DaemonThreadPoolTest {
    @Test
    fun testUselessThreadPool() {
        runConcurrentTest(1000) {
            val pool = Executors.newFixedThreadPool(1).asCoroutineDispatcher()
            runBlocking(pool) { /* do nothing */ }
            // pool.close() -- this call tells gpmc that threads are not in infinite spinning,
            //                         but interleavings exploration still happens
        }
    }
}
```
Output:
```txt
= Concurrent test has hung =

The following interleaving leads to the error:
| ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
|                                                                                Thread 1                                                                                |                                                                       Thread 2                                                                       |
| ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| testUselessThreadPool$1.invoke() at DaemonThreadPoolTest$testUselessThreadPool$1.invoke(DaemonThreadPoolTest.kt:24)                                                    |                                                                                                                                                      |
|   BuildersKt.runBlocking(ExecutorCoroutineDispatcherImpl#1,<cont>): Unit#1 at DaemonThreadPoolTest$testUselessThreadPool$1.invoke(DaemonThreadPoolTest.kt:26)          |                                                                                                                                                      |
|     BuildersKt__BuildersKt.runBlocking(ExecutorCoroutineDispatcherImpl#1,<cont>): Unit#1 at BuildersKt.runBlocking(:1)                                                 |                                                                                                                                                      |
|       CoroutineContextKt.newCoroutineContext(GlobalScope#1,ExecutorCoroutineDispatcherImpl#1): CombinedContext#1 at BuildersKt__BuildersKt.runBlocking(Builders.kt:55) |                                                                                                                                                      |
[...]

PR #647, opened 2025-05-15
Support for aborting parked/livelocked user threads when test thread are finished

dmitrii-artuhov:
Resolves issue https://github.com/JetBrains/lincheck/issues/542 (instead of this PR https://github.com/JetBrains/lincheck/pull/585)

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
@@ -206,10 +206,19 @@ open class ThreadScheduler {
      * Checks if the thread is currently live-locked.
      *
      * @param threadId The identifier of the thread to check.
-     * @return `true` if the thread is blocked, `false` otherwise.
+     * @return `true` if the thread is live-locked, `false` otherwise.
      */
     fun isLiveLocked(threadId: ThreadId) =
         getBlockingReason(threadId) is BlockingReason.LiveLocked
+
+    /**
+     * Checks if the thread is currently parked.
+     *
+     * @param threadId The identifier of the thread to check.
+     * @return `true` if the thread is parked, `false` otherwise.
+     */
+    fun isParked(threadId: ThreadId) =
+        getBlockingReason(threadId) is BlockingReason.Parked
 
     /**
      * Checks if the thread was aborted.
@@ -236,6 +245,16 @@ open class ThreadScheduler {
      */
     fun areAllThreadsFinished() =
         threads.values.all { it.state == ThreadState.FINISHED }
+
+    /**
+     * Checks if all threads managed by the scheduler have finished or aborted.
+     * @return `true` if all threads are finished or aborted, otherwise `false`.
+     */
+    fun areAllThreadsFinishedOrAborted() =
+        threads.values.all {
+            it.state == ThreadState.FINISHED ||
+            it.state == ThreadState.ABORTED
+        }
 
     /**
      * Registers a new thread in the scheduler and assigns it a unique identifier.
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
@@ -203,6 +203,15 @@ open class ThreadScheduler {
         getThreadState(threadId) == ThreadState.BLOCKED
 
     /**
+     * Checks if the thread is currently live-locked.
+     *
+     * @param threadId The identifier of the thread to check.
+     * @return `true` if the thread is blocked, `false` otherwise.
+     */
+    fun isLiveLocked(threadId: ThreadId) =
+        getBlockingReason(threadId) is BlockingReason.LiveLocked
+
+    /**
      * Checks if the thread was aborted.
      *
      * @param threadId The identifier of the thread to check.
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -452,6 +452,10 @@ abstract class ManagedStrategy(
         check(threadId == threadScheduler.scheduledThreadId)
         // check if live-lock is detected
         val decision = loopDetector.visitCodeLocation(threadId, codeLocation)
+        // in case of live-lock, try to abort execution
+        if (decision is LoopDetector.Decision.LivelockThreadSwitch) {
+            tryAbortingUserThreads(threadId, BlockingReason.LiveLocked)
+        }
         // check if we need to switch
         val shouldSwitch = when {
             // check if a switch is required by the loop detector replay mode
@@ -553,7 +557,6 @@ abstract class ManagedStrategy(
         blockingReason: BlockingReason? = null,
         beforeMethodCallSwitch: Boolean = false,
     ): Boolean {
-        // before blocking the thread, interrupt it if the interruption flag is set
         val switchReason = blockingReason.toSwitchReason(::iThreadToDisplayNumber)
         // we create switch point on detected live-locks,
         // but that switch is not mandatory in case if there are no available threads
@@ -575,6 +578,10 @@ abstract class ManagedStrategy(
     }
 
     private fun chooseThreadSwitch(iThread: Int, mustSwitch: Boolean = false): Int {
+        // in case threads were aborted via `tryAbortingUserThreads`, no switching required
+        if (threadScheduler.areAllThreadsFinishedOrAborted()) {
+            return iThread
+        }
         if (inIdeaPluginReplayMode && collectTrace) {
             onThreadSwitchesOrActorFinishes()
         }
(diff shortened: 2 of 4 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
```diff
@@ -119,12 +122,11 @@ open class ThreadScheduler {
      *
      * @return a map from thread ids to thread instances that are currently registered.
      */
-    fun getRegisteredThreads() : ThreadMap<Thread> = mutableThreadMapOf<Thread>().apply {
-        for ((threadId, threadData) in threads) {
+    fun getRegisteredThreads() : Sequence<RegisteredThread> {
```
eupp, line 125: Why do you need to change the API here?
Why `Sequence` instead of `List`?
Why introduce `RegisteredThread`? You can already iterate through pairs `(threadId, thread)` using `ThreadMap` API (thread map is just a typealias to a regular map).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -452,6 +452,10 @@ abstract class ManagedStrategy(
         check(threadId == threadScheduler.scheduledThreadId)
         // check if live-lock is detected
         val decision = loopDetector.visitCodeLocation(threadId, codeLocation)
+        // in case of live-lock, try to abort execution
+        if (decision is LoopDetector.Decision.LivelockThreadSwitch) {
+            tryAbortingUserThreads(threadId, BlockingReason.LiveLocked)
+        }
```
eupp, line 458: Can we move it into `processLoopDetectorDecision` ?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -655,13 +656,39 @@ abstract class ManagedStrategy(
...
+     *
+     * Right now only live-locked threads are considered to be resumable.
+     *
+     * TODO: somehow refactor suspended TestThread and add them here as well.
+     */
+    private fun resumableThreads(iThread: Int): List<Int> =
+        if (runner.currentExecutionPart == PARALLEL) {
+            (0 until threadScheduler.nThreads).filter { it != iThread && threadScheduler.isLiveLocked(it) }
+        }
+        else {
```
eupp, line 677: `} else {`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -655,13 +656,39 @@ abstract class ManagedStrategy(
...
+            emptyList()
+        }
+
+    /**
+     * Returns threads that could be switched to directly from thread [iThread]
+     * or those that could be resumed in case there are no direct options.
+     */
+    protected fun availableThreads(iThread: Int): List<Int> =
+        switchableThreads(iThread).let {
```
eupp, line 686: please rewrite without `let` here

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -827,6 +854,47 @@ abstract class ManagedStrategy(
...
+     * and all running user threads are `LiveLocked` or `Parked`.
+     *
+     * This method is expected to be called before [onSwitchPoint] method in
+     * order to abort execution and not to create unnecessary switch points.
+     *
+     * Since this method only can abort executions with `LiveLocked` or `Parked`
+     * user-threads and `Finished` test threads, then invocations of this method are
+     * only meaningful near the corresponding strategy hooks
+     * ([ManagedStrategy.processLoopDetectorDecision], [ManagedStrategy.park],
+     * and [ManagedStrategy.onThreadFinish] where otentially test thread could finish).
```
eupp, line 868:
```suggestion
     * and [ManagedStrategy.onThreadFinish] where potentially test thread could finish).
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -827,6 +854,47 @@ abstract class ManagedStrategy(
...
+        val userThreadsAbortionPossible =
+            // all `TestThread`s are finished (including main: with id of zero)
+            (0 ..< scenario.nThreads).all(threadScheduler::isFinished) &&
+            // The main thread finished its execution (actually all `TestThread`s did): successfully or not, we don't care.
+            // If all user threads (those that are not `TestThread` instances) are not blocked, then abort
+            // the running user threads. Essentially treating them as "daemons", which completion we do not wait for.
+            // Thus, abort all of them and allow `runner` to process the invocation result accordingly.
+            getRegisteredThreads()
+                .map { it.threadId }
+                .filterNot { isTestThread(it) || it == threadId /* we check invoking thread separately */ }
```
eupp, line 883: Could you please extract this part (perhaps even in a method inside `ThreadScheduler`):

```kotlin
val userThreads = ...
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -827,6 +854,47 @@ abstract class ManagedStrategy(
...
+                .all{ threadScheduler.isLiveLocked(it) || threadScheduler.isParked(it) }
```
eupp, line 884:
```suggestion
                .all { threadScheduler.isLiveLocked(it) || threadScheduler.isParked(it) }
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -848,6 +916,7 @@ abstract class ManagedStrategy(
         loopDetector.onThreadFinish(threadId)
         traceCollector?.onThreadFinish()
         unblockJoiningThreads(threadId)
+        tryAbortingUserThreads(threadId, null)
```
eupp, line 919: Either make `blockingReason = null` as a default parameter value in `tryAbortingUserThreads`, or otherwise pass it by name: ` tryAbortingUserThreads(threadId, blockingReason = null)`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1956,7 +2026,9 @@ abstract class ManagedStrategy(
 
     private fun objectFqTypeName(obj: Any?): String {
         val enumPrefix = if (obj?.javaClass?.isEnum == true) "Enum:" else ""
-        return "$enumPrefix${obj?.javaClass?.name ?: "null"}"
+        val typeName = obj?.javaClass?.name ?: "null"
+        return if (enumPrefix.isEmpty()) typeName
```
eupp, line 2030: Looks like `if` is redundant here, just `$enumPrefix$typeName` would do the same.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -253,6 +254,9 @@ class SnapshotTracker {
         )
     }
 
+    private fun shouldIgnoreClass(owner: String) =
+        owner.startsWith("com.intellij.rt.debugger")
```
eupp, line 258: Could you check if this is still required?
I've added `com.intellij.rt.debugger` into a list of non-instrumented packages in some previous PR:
https://github.com/JetBrains/lincheck/blob/7b02d5410ca4952bdb07594f8c0e00377ae5b5d7/src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt#L509

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/coroutines/DaemonThreadPoolTest.kt
```diff
@@ -11,24 +11,158 @@
...
+            //                 abort background threads, when main is finished
+        }
+    }
+
+    @Test
+    fun testUselessThreadPoolWithMultipleCoroutines() {
+        runConcurrentTest(1000) {
+            val pool = Executors.newFixedThreadPool(5).asCoroutineDispatcher()
+            runBlocking(pool) {
+                val c1 = launch(pool) {}
```
eupp, line 49: Please insert some busy code into launched coroutines, like an increment of atomic integer.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/coroutines/DaemonThreadPoolTest.kt
```diff
@@ -11,24 +11,158 @@
...
             runBlocking(pool) { /* do nothing */ }
-            // pool.close() -- this call tells gpmc that threads are not in infinite spinning,
-            //                 but interleavings exploration still happens
+            // pool.close() -- we do not need this call, because lincheck will
+            //                 abort background threads, when main is finished
+        }
+    }
+
+    @Test
+    fun testUselessThreadPoolWithMultipleCoroutines() {
```
eupp, line 45: Would make sense to rename the test to something like `testNonClosedThreadPoolCoroutineDispatcher`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/coroutines/DaemonThreadPoolTest.kt
```diff
@@ -11,24 +11,158 @@
...
+                c3.join()
+                c4.join()
+                c5.join()
+            }
+            // pool.close() -- no manual closing
+        }
+    }
+
+    @Test
+    fun testInfiniteBackgroundThreads() {
```
eupp, line 66: Could you please add a short comment to each test, explaining expected behavior. 
For instance, here the behavior is that "even though background user threads do not terminate, the whole test should finish once the main thread finishes".

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: Call to `tryAbortingUserThreads` should be done before `onSwitchPoint` call. If adding `tryAbortingUserThreads` inside `processLoopDetectorDecision`, then I would also need to split calls to `onSwitchPoint` -- one in the process loop detector, and the second one before calling `shouldSwitch` after the check for `decision !is LoopDetector.Decision.Idle`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: ThreadScheduler does not differentiate between user and test threads. I added a method to the stategy

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
dmitrii-artuhov: As I profiled earlier the version `"$enumPrefix$typeName"` copies the string. But in most of the invocations of this method, I just need to return `obj?.javaClass?.name` which is already allocated

eupp: Ok, let's then wrap the entier `Enum` path into separate `if` branch:
```
private fun objectFqTypeName(obj: Any?): String {
        val typeName = obj?.javaClass?.name ?: "null"
        // Note: `if` here is important for performance reasons;
        // In common case we want to return just `typeName` without using string templates to avoid redundant string allocation.
        if (obj?.javaClass?.isEnum == true) {
            return "Enum:$typeName"
        }
        return typeName
}
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: I had this problem in the original PR, where the check in agent's `shouldTransform` was present already.
There was an attempt to add over 50k+ allocated objects in the snapshot, when I was debugging some code. Right now I am not able to reproduce it (when I remove `shouldIgnoreClass` method)

dmitrii-artuhov: checked for the second time -- it helped

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
dmitrii-artuhov: it was a memory optimizations: from 560 MB on all tests allocated -> 185 mbs.
reversed as not needed right now

eupp (approved):

PR #648, opened 2025-05-15
Move guide tests into integration tests category

eupp:


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/ObstructionFreedomViolationEventsCutTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    fun dequeue(): Int? = q.dequeue()
+
+    @Test
+    fun runModelCheckingTest() = ModelCheckingOptions()
+        .checkObstructionFreedom(true)
+        .checkImpl(this::class.java)
+        .checkLincheckOutput("obstruction_freedom_violation_events_cut")
+}
+
+class MSQueueBlocking {
```
dmitrii-artuhov, line 46: Now implementation of `MSQueueBlocking` is added to this file, but from where was it taken from? I don't see it in the diff as removed. Is there a duplicate of class `MSQueueBlocking` right now in the source code?

dmitrii-artuhov: Tests from `test-integration` can include classes from `test` module. You can define `MSQueueBlocking` in `test` and just include in guide test, because there is the same class defined in `test-integration/org/jetbrains/kotlinx/lincheck_test/guide/ObstructionFreedomViolationTest.kt`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/ObstructionFreedomViolationEventsCutTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.jetbrains.kotlinx.lincheck_test.util.checkLincheckOutput
+import org.junit.Assume.assumeFalse
+import org.junit.Before
+import org.junit.Test
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * Checks that spin-cycle repeated events are cut in case of obstruction freedom violation
+ */
+class ObstructionFreedomViolationEventsCutTest {
```
dmitrii-artuhov, line 26: why this is required to be put outside of `SpinlockEventsCutTests.kt` file? It logically belongs to other `events cut` tests

dmitrii-artuhov (changes requested):

dmitrii-artuhov:
also links to guide tests in docs should be updated (here or as a second PR after emerging this one)

dmitrii-artuhov (approved):

eupp:
> also links to guide tests in docs should be updated (here or as a second PR after merging this one)

Where are these links mentioned?

PR #649, opened 2025-05-18
Support java.io, java.nio.file in trace debugger

zhelenskiy:
Trace debugger support for file/directory-related API excluding `java.nio.Channel`-related

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -1639,7 +1639,7 @@ abstract class ManagedStrategy(
             codeLocation = codeLocation,
             methodId = methodId,
         )
-        val deterministicMethodDescriptor = getDeterministicMethodDescriptorOrNull(methodCallInfo)
+        val deterministicMethodDescriptor = getDeterministicMethodDescriptorOrNull(receiver, params, methodCallInfo)
         // get method's analysis section type
         val methodSection = methodAnalysisSectionType(receiver, className, methodName,
             atomicMethodDescriptor,
@@ -1705,7 +1705,8 @@ abstract class ManagedStrategy(
         receiver: Any?,
         params: Array<Any?>,
         result: Any?
-    ) = runInsideIgnoredSection {
+    ): Any? = runInsideIgnoredSection {
+        var newResult = result
         if (deterministicMethodDescriptor != null) {
             Logger.debug { "On method return with descriptor $deterministicMethodDescriptor: $result" }
         }
@@ -1717,9 +1718,9 @@ abstract class ManagedStrategy(
         }
 
         if (isInTraceDebuggerMode && isFirstReplay && deterministicMethodDescriptor != null) {
-            deterministicMethodDescriptor.saveFirstResultWithCast(receiver, params, KResult.success(result)) {
+            newResult = deterministicMethodDescriptor.saveFirstResultWithCast(receiver, params, KResult.success(result)) {
                 nativeMethodCallStatesTracker.setState(descriptorId, deterministicMethodDescriptor.methodCallInfo, it)
-            }
+            }.getOrElse { error("Unexpected replacement success -> failure:\n$result\n${KResult.failure<Any?>(it)}") }
         }
         val threadId = threadScheduler.getCurrentThreadId()
         // check if the called method is an atomics API method
@@ -1749,6 +1750,7 @@ abstract class ManagedStrategy(
         }
         // if the method has certain guarantees, leave the corresponding section
         leaveAnalysisSection(threadId, methodSection)
+        return newResult
     }
 
     override fun onMethodCallException(
@@ -1759,14 +1761,17 @@ abstract class ManagedStrategy(
         receiver: Any?,
         params: Array<Any?>,
         throwable: Throwable
-    ) = runInsideIgnoredSection {
+    ): Throwable = runInsideIgnoredSection {
+        var newThrowable = throwable
         if (deterministicMethodDescriptor != null) {
             Logger.debug { "On method exception with descriptor $deterministicMethodDescriptor:\n${throwable.stackTraceToString()}" }
         }
         require(deterministicMethodDescriptor is DeterministicMethodDescriptor<*, *>?)
         if (isInTraceDebuggerMode && isFirstReplay && deterministicMethodDescriptor != null) {
-            deterministicMethodDescriptor.saveFirstResult(receiver, params, KResult.failure(throwable)) {
+            newThrowable = deterministicMethodDescriptor.saveFirstResult(receiver, params, KResult.failure(throwable)) {
                 nativeMethodCallStatesTracker.setState(descriptorId, deterministicMethodDescriptor.methodCallInfo, it)
+            }.let { newResult ->
+                newResult.exceptionOrNull() ?: error("Unexpected replacement failure -> success:\n$throwable\n$newResult")
             }
         }
         val threadId = threadScheduler.getCurrentThreadId()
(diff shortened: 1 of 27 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -139,7 +139,7 @@ internal class LincheckClassVisitor(
         // We need to disable breakpoints in such a case, as the numeration will break.
         // Breakpoints are disabled as we do not instrument toString and enter an ignored section,
         // so there are no beforeEvents inside.
-        if (methodName == "<init>" || ideaPluginEnabled && methodName == "toString" && desc == "()Ljava/lang/String;") {
+        if (ideaPluginEnabled && methodName == "toString" && desc == "()Ljava/lang/String;") {
```
eupp, line 142: Reminder: this part was extracted into separate PR https://github.com/JetBrains/lincheck/pull/661

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1686,9 +1687,9 @@ abstract class ManagedStrategy(
         }
 
         if (isInTraceDebuggerMode && isFirstReplay && deterministicMethodDescriptor != null) {
-            deterministicMethodDescriptor.saveFirstResultWithCast(receiver, params, KResult.success(result)) {
+            newResult = deterministicMethodDescriptor.saveFirstResultWithCast(receiver, params, KResult.success(result)) {
                 nativeMethodCallStatesTracker.setState(descriptorId, deterministicMethodDescriptor.methodCallInfo, it)
-            }
+            }.getOrThrow()
```
eupp, line 1692: Isn't it too early to exit?
What about other stuff that happens below?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/io/FileTests.kt
```diff
@@ -0,0 +1,450 @@
...
+        require(file.exists()) { "File was not created: $file" }
+
+        try {
+            // Write initial content
+            val initialContent = "Hello, "
+            FileWriter(file).use { writer ->
+                writer.write(initialContent)
+            }
+
+            // Append content
```
eupp, line 428: Could you please also add another similar test, where file is closed and re-opened between an intial write and append.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
zhelenskiy: I expect it to never happen. Errors should not be replaced with not errors, etc.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
zhelenskiy: I know, I am waiting for its merge.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/native_calls/DeterministicFileMethodDescriptor.kt
```diff
@@ -0,0 +1,615 @@
...
+private fun <T> notSupportedInLincheck(): PureDeterministicMethodDescriptor<T>.(Any?, Array<Any?>) -> T = { _, _ ->
+    throwUnsupportedFileError()
+}
+
+private fun throwUnsupportedFileError(): Nothing {
+    error("File operations are not supported in Lincheck")
+}
+
+private fun pure(methodCallInfo: MethodCallInfo) =
+    PureDeterministicMethodDescriptor<Any?>(methodCallInfo, notSupportedInLincheck())
```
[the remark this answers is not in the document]
eupp: What do you mean? 
All I see is that this is just a function that creates `PureDeterministicMethodDescriptor` which throws error on `runFake`. 

What it has to do with `DeterministicFileMethodDescriptor` or IO in general?
It is a generic utility function.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/native_calls/DeterministicFileMethodDescriptor.kt
```diff
@@ -0,0 +1,615 @@
...
+            State(byteArray.copyOfRange(fromIndex = offset, toIndex = offset + len), returned)
+        }
+        saveState(state)
+        return result
+    }
+}
+
+private class DeterministicIteratorMethodDescriptor<T>(
+    override val methodCallInfo: MethodCallInfo,
+) : DeterministicMethodDescriptor<Result<TrackedIterator<T>>, Iterator<T>>() {
```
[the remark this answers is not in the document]
eupp: > Result<TrackableIterator> whose every action is tracked

I can see that from the function signature. I am trying to understand the semantics. 

The thing is, everything that is saved as a `State` survives between replays. 

- Saving immutable values in a `State` is fine, because they have no mutable state and are copied on replay. 
- Saving copyable values like `ByteArray` is fine, because we copy the array on replay. 

Saving a mutable object with identity between replay in general is not safe and error prone. 
Some mutations of the object can leak between different replays, leading to hard-to-debug bugs. 

That is why I am asking. 
I am trying to understand the class semantics and contract here. 
Why is it safe to save `TrackableIterator` between replays?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/native_calls/DeterministicFileMethodDescriptor.kt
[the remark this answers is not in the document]
zhelenskiy: In some cases there's not throwing implementation.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: > I expect it to never happen

Could you please then instead add a `check` here. Because otherwise you would re-throw the exception stored in the descriptor, and not the indication that something went wrong in the code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/native_calls/DeterministicFileMethodDescriptor.kt
```diff
@@ -0,0 +1,615 @@
...
+import sun.nio.ch.lincheck.Injections
+import sun.nio.ch.lincheck.MethodSignature
+import sun.nio.ch.lincheck.Types.ArrayType
+import sun.nio.ch.lincheck.Types.BOOLEAN_TYPE
+import sun.nio.ch.lincheck.Types.BYTE_TYPE
+import sun.nio.ch.lincheck.Types.INT_TYPE
+import sun.nio.ch.lincheck.Types.LONG_TYPE
+import sun.nio.ch.lincheck.Types.MethodType
+import sun.nio.ch.lincheck.Types.ObjectType
+import sun.nio.ch.lincheck.Types.VOID_TYPE
```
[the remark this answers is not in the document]
zhelenskiy: I don't want to import everything as it might break the code because of the collisions

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/trace_debugger/io/FileTests.kt
zhelenskiy: But you are already pointing at the such test: `FileAppendTest`. `use` opens and closes under-the-hood.

eupp (commented):
Let's try to merge this without waiting for https://github.com/JetBrains/lincheck/pull/661. 

One possible approach is to instrument constructor bodies only for trace debugger under a flag.

PR #655, opened 2025-05-19
Java lambda argument representation

bbrockbernd:


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/LambdaRepresentation.kt
```diff
@@ -0,0 +1,20 @@
...
+package org.jetbrains.kotlinx.lincheck_test.representation
+
+class LambdaRepresentation: BaseTraceRepresentationTest("java_lambda_argument_representation") {
+    var a = 1
+    val hm = HashMap<Int, Int>()
+    override fun operation() {
+        hm.computeIfAbsent(1) { a++ }
+    }
+
+}
```
eupp, line 20: Could you please convert this file to `.java`, so we are sure that lambdas are compiled using Java convention?

eupp (changes requested):

eupp (approved):

Issue #653, opened 2025-05-19
Do not store call stack trace in each trace point


eupp:
Depends on #652

ivandev0:
Solved with PR #667

PR #656, opened 2025-05-19
`LoopDetector` API simplification

eupp:
In https://github.com/JetBrains/lincheck/issues/653, we plan to get rid of call stacks stored in trace points.
Currently, there is only a single place where call stacks of trace points are actually used: inside `LoopDetector`, method `afterSpinCycleTraceCollected`.
It would be nice to get rid of this usage, in order to simplify the implementation of https://github.com/JetBrains/lincheck/issues/653.

This PR makes a first step towards this goal.
While it does not solve the problem, it proposes a refactoring of `LoopDetector` class to make its API and contract more comprehensible and predictable.

The next step is to try get rid of call stack usages in the `LoopDetector`.


src/jvm/test/resources/expected_logs/suspend_trace_reporting_jdk11.txt
```diff
@@ -29,12 +29,12 @@ The following interleaving leads to the error:
...
-|         installParentHandle(): ChildContinuation#1 at CancellableContinuationImpl.getResult(CancellableContinuationImpl.kt:311)                                                           |                                                                                                                                                                      |
-|         switch (reason: coroutine is suspended)                                                                                                                                           |                                                                                                                                                                      |
+|         switch                                                                                                                                                                            |                                                                                                                                                                      |
```
bbrockbernd, line 32: The switch reason seems to miss in the trace (other txts as well). Is this intended?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -93,46 +106,54 @@ internal class LoopDetector(
...
+    /**
+     * In replay mode, represents the period of spin-cycle if spin-cycle was entered.
+     */
+    private val replayModeCurrentCyclePeriod: Int
+        get() = replayModeLoopDetectorHelper?.currentCyclePeriod ?: 0
+
+    /**
+     * Is called before each interleaving processing
+     */
+    fun initialize() {
```
bbrockbernd, line 140: Nitpick: wouldn't `clear` or `reset` better describe the purpose of this method? Idem for other places with similar name, that reset state.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -218,36 +239,42 @@ internal class LoopDetector(
...
+                // spin-loop detected - switch
+                count > hangingDetectionThreshold ->
+                    Decision.LivelockThreadSwitch(hangingDetectionThreshold)
+                // live-lock detected - fail
                 totalExecutionsCount > ManagedCTestConfiguration.LIVELOCK_EVENTS_THRESHOLD ->
-                    Decision.EventsThresholdReached // live-lock detected, fail
-                else -> Decision.Idle
+                    Decision.EventsThresholdReached
+                // else - continue
+                else ->
```
bbrockbernd, line 270:
```suggestion
else -> Decision.Idle
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -281,25 +308,36 @@ internal class LoopDetector(
...
         }
-        val cyclePeriod = replayModeCurrentCyclePeriod
-        return if (detectedFirstTime || detectedEarly)
-            Decision.LivelockThreadSwitch(cyclePeriod)
-        else
-            Decision.Idle
+        if (detectedFirstTime || detectedEarly) {
+            val cyclePeriod = replayModeCurrentCyclePeriod
+            return Decision.LivelockThreadSwitch(cyclePeriod)
+        }
```
bbrockbernd, line 314:
```suggestion
if (detectedFirstTime || detectedEarly) return Decision.LivelockThreadSwitch(replayModeCurrentCyclePeriod)
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -695,13 +666,13 @@ private class ReplayModeLoopDetectorHelper(
...
         currentInterleavingNodeIndex = 0
         executionsPerformedInCurrentThread = 0
-        threadsRan.clear()
     }
 
     /**
      * Called before next execution in current thread.
      */
-    fun onNextExecution() {
+    @Suppress("UNUSED_PARAMETER")
```
bbrockbernd, line 674: Maybe a comment on why we still choose to have this parameter?

bbrockbernd (changes requested):

src/jvm/test/resources/expected_logs/suspend_trace_reporting_jdk11.txt
eupp: I think in this particular test, what has happened is that switch point position changed. 
The `installParentHandle()` method call now happens after the thread switch. 

Before, the thread switch was forced due to coroutine suspension.
Now it is just a strategy switch (i.e., thread switch inserted by the strategy itself) --- these thread switches do not have reason string.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
eupp: Done.

PR #661, opened 2025-05-20
Instrument constructor bodies

zhelenskiy:


src/jvm/test/resources/expected_logs/single_suspension_point.txt
```diff
@@ -17,7 +17,7 @@ The following interleaving leads to the error:
...
 |                                                                                                                                             |   suspendAndCauseFailure() [suspendable: Continuation#1]: 0 at SingleSuspensionPointTraceRepresentationTest.operation(MultipleSuspensionsTests.kt:49) |
 |                                                                                                                                             |     counter ➜ AtomicInteger#1 at SingleSuspensionPointTraceRepresentationTest.suspendAndCauseFailure(MultipleSuspensionsTests.kt:54)                  |
-|                                                                                                                                             |     counter.get(): 1 at SingleSuspensionPointTraceRepresentationTest.suspendAndCauseFailure(MultipleSuspensionsTests.kt:54)                           |
+|                                                                                                                                             |     this$0.counter.get(): 1 at SingleSuspensionPointTraceRepresentationTest.suspendAndCauseFailure(MultipleSuspensionsTests.kt:54)                    |
```
ndkoval, line 20: Can we avoid `this$0` here?

src/jvm/test/resources/expected_logs/run_concurrent_test/deque_jdk8.txt
```diff
@@ -12,6 +12,7 @@ The following interleaving leads to the error:
...
 |   block(): threw IllegalStateException at IncorrectConcurrentLinkedDequeRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:349)                          |                                                                                                                                                  |                                                                                                                                                 |
+|     Node#1.item.putObject(null) at ConcurrentLinkedDeque$Node.<init>(ConcurrentLinkedDeque.java:303)                                                                            |                                                                                                                                                  |                                                                                                                                                 |
```
ndkoval, line 15: Can we show constructors as method calls?

src/jvm/test/resources/expected_logs/single_suspension_point.txt
zhelenskiy: I see no way. I consider the cosmetic changes not urgent as this PR is just an extracted step for native calls.

src/jvm/test/resources/expected_logs/run_concurrent_test/deque_jdk8.txt
ndkoval: What are the possible ways?

src/jvm/test/resources/expected_logs/single_suspension_point.txt
ndkoval: It makes the trace slightly less intuitive. Could you please analyze how to fix it, please? It does not extremely complex.

src/jvm/test/resources/expected_logs/run_concurrent_test/deque_jdk8.txt
zhelenskiy: I'm not sure but it should be very complicated as the constructor call is separated into two parts.

ndkoval: @zhelenskiy, we already have such a logic for reads, wrapping them with "beforeRead" and "afterRead" methods. Would the same approach work here?

zhelenskiy: no

src/jvm/test/resources/expected_logs/single_suspension_point.txt
zhelenskiy: They already existed. Generally, it is an actual legitimate name for outer this parameter in the constructor of an inner class.

ndkoval: I see, but I've also noticed you fixed this for some cases. Can your approach be applied to all cases so that all the `this$0` references disappear?

src/jvm/test/resources/expected_logs/run_concurrent_test/deque_jdk8.txt
zhelenskiy: I have previously tried doing this for some time and failed, so I think that:
* It is not that simple.
* It is another [issue](https://youtrack.jetbrains.com/issue/IJTD-160/Support-constructors-in-MethodCallTransformer), and I don't see any reason to mix the PRs.

src/jvm/test/resources/expected_logs/single_suspension_point.txt
zhelenskiy: I mean this name is absolutely legit and necessary:
```kotlin
class A(val x: Int) {
    inner class B(val x: Int) {
        fun f() = x + this@A.x // x + this$0.x, they must be different not to confuse.
    }
}
```

ndkoval: This is not something you type in the code, so I would prefer omitting it in the trace.

zhelenskiy: > This is not something you type in the code

I saw such code many times in my life.

zhelenskiy: How would you like to refer to the outer this?

zhelenskiy:
@dmitrii-artuhov Does it fail both before and after my change?

zhelenskiy:
Then can we just not comment them out? It is unnecessary for the native calls support.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/modelchecking/UninitializedConstructorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.strategy.modelchecking
+
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck_test.strategy.modelchecking.snapshot.AbstractSnapshotTest
+
+class UninitializedConstructorTest : AbstractSnapshotTest() {
```
dmitrii-artuhov, line 16: It does not need to be `AbstractSnapshotTest` subclass

dmitrii-artuhov (changes requested):

PR #663, opened 2025-05-20
Change kotlin version to `2.1.21`

dmitrii-artuhov:


src/jvm/main/org/jetbrains/kotlinx/lincheck/util/JdkVersion.kt
```diff
@@ -0,0 +1,50 @@
...
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
+
+
+/**
+ * Represents a set of Java Development Kit (JDK) versions on which the tests are run.
+ */
+internal enum class TestJdkVersion {
```
eupp, line 17: If you move it here from the test infra code, then you need to support all major JDK versions.
Also please rename to `JdkVersion`.

src/jvm/test/resources/expected_logs/run_concurrent_test/hashmap.txt
```diff
@@ -7,80 +7,82 @@ java.lang.IllegalStateException: Check failed.
...
+|                                                                                                                                                                                 |                                                                    |         switch                                                                                                                          |
+|                                                                                                                                                                                 | run()                                                              |                                                                                                                                         |
+|                                                                                                                                                                                 |   Lambda#2.invoke() at ThreadsKt$thread$thread$1.run(Thread.kt:30) |                                                                                                                                         |
```
eupp, line 27: Why `Lambda#2.invoke()` started to apper in the short trace? Can we hide it from the short trace?

dmitrii-artuhov: fixed

PR #664, opened 2025-05-21
Hide and SILENT stdLib collections

bbrockbernd:
- Hides Collection internals like Maps, Lists, Sets etc
- Concurrent collections are silent
- All collections are collapsed in trace, but only if no switch point is present. Calls leading up to and neighboring switchpoint are kept.
- Unifies Section defenitions with shouldTransform definitions (currently should not transform is defined as ignored, but passed as silent to managedstrategy due to propagating issues)
- Adds user API (to GPMC and Lincheck) to enable/disable user library analysis. For concurrent collections this means it is moved from silent to normal. For all collections this means the function calls and internal events are not collapsed. 
- Lincheck Failure now contains test configuration. (This was the easiest and IMO most intuitive way to access test config from reporting logic) This was necessary to know when to collapse or not based on user settings.


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1823,17 +1824,20 @@ abstract class ManagedStrategy(
...
-            return silentSection
-        }
+        val section = getAnalysisSectionFor(ownerName, methodName)
         userDefinedGuarantees?.forEach { guarantee ->
             if (guarantee.classPredicate(ownerName) && guarantee.methodPredicate(methodName)) {
                 return guarantee.type
             }
         }
-        return AnalysisSectionType.NORMAL
+        return if (section == AnalysisSectionType.IGNORED) AnalysisSectionType.SILENT else section
```
eupp, line 1833: So it turned out merging `IGNORED` section and instrumentation logic was not as easy as we thought initially. 
So let's postpone this change for now, create an issue ticket with the description of the problem and revisit it later. 

See my comment in the `AnalysisSection` file with the description of how I can proceed for now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -182,23 +184,157 @@ internal inline fun <R> runOutsideIgnoredSection(block: () -> R): R {
...
-    }
-    return null
+/**
+ * Determines the analysis section type for a given class. And optionally function.
+ *
+ * @param className The name of the class to analyze.
+ * @param methodName The name of the method to analyze.
+ * @return AnalysisSectionType.NORMAL if the class should be transformed, AnalysisSectionType.IGNORED otherwise.
+ */
+internal fun getSectionDefinitionFor(className: String, methodName: String = ""): AnalysisSectionType = when {
```
eupp, line 194: Let's rollback a bit, and split analysis section computation and `shouldTransform` check back for now. At the same time, let's keep them coupled in one class --- so we have one entity that decides what should be instrumented, and what analysis section should be assigned.  

So let's implement the following API:
```kotlin
class AnalysisProfile {

    // decides what methods should be instrumented
    fun shouldTransform(className, methodName): Boolean

    // decides what analysis section should be assigned to the method
    fun getAnalysisSectionFor(className, methodName): AnalysisSectionType

    // decides what methods should be hidden from the trace
    fun shouldBeHidden(className, methodName): Boolean 

}
```

`shouldBeHidden` name is debatable, feel free to rename if you have better idea in mind.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -182,23 +184,157 @@ internal inline fun <R> runOutsideIgnoredSection(block: () -> R): R {
...
+)
+
 private fun isJavaExecutorService(className: String) =
-    className.startsWith("java.util.concurrent.AbstractExecutorService") ||
-    className.startsWith("java.util.concurrent.ThreadPoolExecutor") ||
-    className.startsWith("java.util.concurrent.ForkJoinPool")
\ No newline at end of file
+    className.startsWith("java.util.concurrent.AbstractExecutorService") 
+        || className.startsWith("java.util.concurrent.ThreadPoolExecutor") 
+        || className.startsWith("java.util.concurrent.ForkJoinPool")
```
eupp, line 340: Is this formatting change necessary?
IMO, it looks better when all lines have same indentation.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -59,7 +60,9 @@ internal class TraceReporter(
         val traceGraph = traceToGraph(fixedTrace)
 
         // Optimizes trace by combining trace points for synthetic field accesses etc..
-        val compressedTraceGraph = traceGraph.compressTrace()
+        val compressedTraceGraph = traceGraph
+            .compressTrace()
+            .let { if (failure.testConfig is ModelCheckingCTestConfiguration && !failure.testConfig.stdLibAnalysisEnabled) it.collapseLibraries() else it }
```
eupp, line 65: Please extract:
```kotlin
(failure.testConfig is ModelCheckingCTestConfiguration && !failure.testConfig.stdLibAnalysisEnabled)
```

into a separate function.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -190,6 +193,28 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
...
+    val newNode = node.copy()
+    findNonLibrarySubTrees(node).forEach {  newNode.addChild(it) }
+    return@compressNodes newNode
+}
+
+/**
+ * Finds descendants that should not be hidden.
+ * But not descendants of descendants, aka the root of the subtree that should not be hidden.
+ */
+private fun findNonLibrarySubTrees(node: TraceNode): List<TraceNode> {
```
eupp, line 213: Please rename. It is not specific to stdlibs, in future we might want to hide other stuff as well.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
```diff
@@ -12,63 +12,71 @@ package org.jetbrains.kotlinx.lincheck.strategy
...
 import org.jetbrains.kotlinx.lincheck.runner.*
+import org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedCTestConfiguration
 import org.jetbrains.kotlinx.lincheck.trace.Trace
 
 sealed class LincheckFailure(
     val scenario: ExecutionScenario,
     val results: ExecutionResult,
-    val trace: Trace?
+    val trace: Trace?,
+    val testConfig: ManagedCTestConfiguration?
```
eupp, line 22: Once you implement `AnalysisProfile` class, let's pass it instead of `ManagedCTestConfiguration`. 

Passing `ManagedCTestConfiguration` has the following drawbacks:
- It contains too much details not relevant to the trace representation.
- It is coupled with `ManagedStrategy`.
- We will likely change the configuration API in the future.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategyGuarantee.kt
```diff
@@ -75,5 +75,10 @@ class ManagedStrategyGuarantee private constructor(
          * specified method calls.
          */
         fun treatAsAtomic() = ManagedStrategyGuarantee(classPredicate, methodPredicate, AnalysisSectionType.ATOMIC)
+
+        /**
+         * The methods will be treated as SILENT, so that no switch points will be placed unless forced.
+         */
+        internal fun mute() = ManagedStrategyGuarantee(classPredicate, methodPredicate, AnalysisSectionType.SILENT)
```
eupp, line 82: I actually like the `mute` name!

We might consider renaming `SILENT` to `MUTED` (just an idea for the future, not for this PR).

src/jvm/test-integration/org/jetbrains/kotlinx/lincheck_test/guide/ConcurrentLinkedDequeTest.kt
```diff
@@ -39,5 +39,7 @@ class ConcurrentLinkedDequeTest {
 
     //@Test // TODO: Please, uncomment me and comment the line below to run the test and get the output
     @Test(expected = AssertionError::class)
-    fun modelCheckingTest() = ModelCheckingOptions().check(this::class)
+    fun modelCheckingTest() = ModelCheckingOptions()
+        .analyzeStdLib(true)
```
eupp, line 43: There are also some stdlib tests in `org/jetbrains/kotlinx/lincheck_test/verifier/linearizability`, could you please also add this flag there. 

Also please re-check by project search if there are any other affected tests (search for `ConcurrentHashMap`, `ConcurrentLinkedDequeue`, etc).

src/jvm/test/resources/expected_logs/thread_safe_collection_muted_jdk11.txt
```diff
@@ -0,0 +1,24 @@
...
+Detailed trace:
+| ------------------------------------------------------------------------------------------------------------------------- |
+|                                                         Thread 1                                                          |
+| ------------------------------------------------------------------------------------------------------------------------- |
+| operation()                                                                                                               |
+|   concurrentMap.put(1, 1): null at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:27) |
+|   a ➜ 1 at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:28)                         |
+|   a = 2 at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:28)                         |
+|   result: void                                                                                                            |
+| ------------------------------------------------------------------------------------------------------------------------- |
```
eupp, line 24: Is there any diff wrt `thread_safe_collection_muted.txt` ?

src/jvm/test/resources/expected_logs/thread_safe_collection_muted_jdk8.txt
```diff
@@ -0,0 +1,24 @@
...
+Detailed trace:
+| ------------------------------------------------------------------------------------------------------------------------- |
+|                                                         Thread 1                                                          |
+| ------------------------------------------------------------------------------------------------------------------------- |
+| operation()                                                                                                               |
+|   concurrentMap.put(1, 1): null at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:27) |
+|   a ➜ 1 at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:28)                         |
+|   a = 2 at ThreadSafeCollectionMutedRepresentationTest.operation(StdLibRepresentationTests.kt:28)                         |
+|   result: void                                                                                                            |
+| ------------------------------------------------------------------------------------------------------------------------- |
```
eupp, line 24: Is there any diff wrt `thread_safe_collection_muted.txt` ?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/StdLibRepresentationTests.kt
```diff
@@ -0,0 +1,108 @@
...
+
+import org.jetbrains.kotlinx.lincheck.annotations.Operation
+import org.jetbrains.kotlinx.lincheck.checkImpl
+import org.jetbrains.kotlinx.lincheck.strategy.managed.forClasses
+import org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingOptions
+import org.jetbrains.kotlinx.lincheck_test.util.checkLincheckOutput
+import org.junit.Test
+import java.util.concurrent.ConcurrentHashMap
+
+class ThreadSafeCollectionMutedRepresentationTest: BaseTraceRepresentationTest("thread_safe_collection_muted") {
```
eupp, line 21: Could you please inherit from `BaseRunConcurrentRepresentationTest` instead?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/StdLibRepresentationTests.kt
```diff
@@ -0,0 +1,108 @@
...
+        concurrentMap.computeIfAbsent(1) { 
+            a = 5
+            innerConcurrentMap.computeIfAbsent(1) { ++a }
+            ++a 
+        }
+        a++
+    }
+}
+
+class ThreadSafeCollectionUnmutedRepresentationTest: BaseTraceRepresentationTest("thread_safe_collection_unmuted") {
```
eupp, line 48: Please move this test to appear right after `ThreadSafeCollectionMutedRepresentationTest`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/StdLibRepresentationTests.kt
```diff
@@ -0,0 +1,108 @@
...
+    private val concurrentMap = ConcurrentHashMap<Int, Int>()
+    private var a = 1
+    
+    override fun operation() {
+        concurrentMap.put(1, 1)
+        a++
+    }
+}
+
+class ThreadSafeCollectionMutedLambdaRepresentationTest: BaseTraceRepresentationTest("thread_safe_collection_muted_lambda") {
```
eupp, line 32: Please add a variant of this test where `computeIfAbsent` is not muted.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/StdLibRepresentationTests.kt
```diff
@@ -0,0 +1,108 @@
...
+            parallel {
+                thread { actor(::put1) }
+                thread { actor(::put2) }
+            }
+        }
+        .addGuarantee(forClasses("java.util.Map", "java.util.HashMap").allMethods().mute())
+        .iterations(0)
+        .checkImpl(this::class.java) { failure ->
+            check(failure == null) { "Test should pass" }
+        }
```
eupp, line 107: Please use `runConcurrentTest` API instead of data-structures based API.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -32,8 +32,9 @@ object Lincheck {
     @JvmStatic
     fun runConcurrentTest(
         invocations: Int = DEFAULT_INVOCATIONS_COUNT,
-        block: Runnable
-    ) {
+        analyzeStdLib: Boolean = false,
```
ndkoval, line 35: Please do not change the API, add a system property instead.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
bbrockbernd: The `||` used to be at the end of previous lines which is confusing IMO. And IJ decided to use this indentation level. But do agree that this extra tab is not ideal.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -461,72 +462,11 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
         if (instrumentationMode == STRESS) {
             if (className.startsWith("java.") || className.startsWith("kotlin.")) return false
         }
-        // We should transform all eagerly instrumented classes.
-        if (isEagerlyInstrumentedClass(className)) {
-            return true
-        }
```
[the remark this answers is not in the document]
bbrockbernd: In the end this function also leads to className definitions like `java.lang.StackTraceElement`. So if we would keep this check here that would mean `java.lang.StackTraceElement` would be instrumented but ends up in IGNORED for analysis since everything in `java.` falls into ignored unless specified otherwise.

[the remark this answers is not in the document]
eupp: Yes, I agree there is a conflict and overlap here, between the analysis sections, `shouldTransform`, and eager/lazy transformation. However, for now, I do not know how to better approach this problem, so let's not keep the old behavior for now : )

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_trace_debugger.txt
```diff
@@ -5,10 +5,9 @@ The following interleaving leads to the error:
...
 | $this_runCatching.block() at BaseRunConcurrentRepresentationTest.testRunWithModelChecker$lambda$1$lambda$0(RunConcurrentRepresentationTests.kt:47) |
-|   block() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:96)                                        |
-|     q.take() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:101)                                    |
+|   block() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:98)                                        |
+|     q.take() at MainThreadBlockedRunConcurrentRepresentationTest.block(RunConcurrentRepresentationTests.kt:103)                                    |
 |       lock.lockInterruptibly() at ArrayBlockingQueue.take(ArrayBlockingQueue.java:417)                                                             |
-|       count ➜ 0 at ArrayBlockingQueue.take(ArrayBlockingQueue.java:419)                                                                            |
```
eupp, line 11: BTW, why these `count` accesses disappeared from the trace

bbrockbernd: `q.take()` is a function operated on a collection. According to our new hiding rules it shows the function (just like `add` and `put` etc..)

Now this function has three child nodes: `lock.lockInterruptibly()`, `count -> 0`, `notEmpty.await()`.

The first and last function call are not part of the collection stdLib and are therefore not hidden. However, the count -> 0 event is part of the library and therefore is hidden.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
bbrockbernd: Sounds good, done.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
bbrockbernd: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
eupp: After a discussion, decided to leave this option, but make the API internal, only expose publicly the version without additional options.

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_trace_debugger.txt
ndkoval: Shouldn't Lincheck hide either everything or nothing in this case? What is the expected semantics here?

bbrockbernd: That wouldn't work for functions taking lambda's. For instance, if we would hide everything from `computeIfAbsent` we would not see user code. But keeping everything would leave a lot of noise. 

Therefore we cannot blindly hide everything inside a function and the resulting semantics are that we only hide what is known that should be hidden (in this PR the std Collections).

Similarly, when a switchpoint does occus in one of these collections, we do not want to show everything that happened inside. Only the calls leading up to the switch event.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -182,23 +185,198 @@ internal inline fun <R> runOutsideIgnoredSection(block: () -> R): R {
...
+        // We do not need to instrument most standard Java classes.
+        // It is fine to inject the Lincheck analysis only into the
+        // `java.util.*` ones, ignored the known atomic constructs.
+        if (className.startsWith("java.")) {
+            if (className == "java.lang.Thread") return true
+            if (className.startsWith("java.util.concurrent.") && className.contains("Atomic")) return false
+            if (className.startsWith("java.util.")) return true
+            return false
+        }
+            if (className.startsWith("com.sun.")) return false
```
eupp, line 222: Indentation is broken here and in other places in this function.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -461,72 +462,12 @@ internal object LincheckClassFileTransformer : ClassFileTransformer {
...
-        if (className.startsWith("org.junit.")) return false
-        if (className.startsWith("junit.framework.")) return false
-        // Finally, we should never instrument the Lincheck classes.
-        if (className.startsWith("org.jetbrains.kotlinx.lincheck.")) return false
-        if (className.startsWith("sun.nio.ch.lincheck.")) return false
-        // All the classes that were not filtered out are eligible for transformation.
-        return true
+        if (isEagerlyInstrumentedClass(className)) return true
+
+        return AnalysisProfile(true).shouldTransform(className, "")
```
eupp, line 467:
```suggestion
        return AnalysisProfile(analyzeStdLib = true).shouldTransform(className, "")
```

eupp (commented):
Please fix few minor problems. Otherwise, looks good to me.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
bbrockbernd: Done

eupp:
Should we merge?
CC @ndkoval

eupp:
@bbrockbernd could you please make `analyzeStdLib` to be `internal` API everywhere?

src/jvm/test/resources/expected_logs/run_concurrent_test/main_thread_blocked_trace_debugger.txt
ndkoval: The question is mainly about the resulting trace we wish to show here, not the implementation. What would you, as a user, prefer to see, and how can we get there?

bbrockbernd: We could also hide all contents until we find a lambda or runnable, but a lambda can also be part of some internal logic..

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -33,6 +33,22 @@ object Lincheck {
...
+     * This method will explore different interleavings of the [block] body and all the threads created within it,
+     * searching for the first raised exception.
+     *
+     * @param lincheckSettings settings that determine linchecks behaviour like analyse std library collections.
+     * @param invocations number of different interleavings of code in the [block] that should be explored.
+     * @param block lambda which body will be a target for the interleavings exploration.
+     */
+    @JvmOverloads
+    @JvmStatic
+    internal fun runConcurrentTest(
```
ndkoval, line 48: Let's rename to `runConcurrentTestInternal`

PR #667, opened 2025-05-26
Store `MethodCallTracePoint` directly in the trace

ivandev0:


src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -538,6 +522,14 @@ internal class SpinCycleStartTracePoint(iThread: Int, actorId: Int, callStackTra
     }
 }
 
+internal class ReturnFromMethodTracePoint(iThread: Int, actorId: Int, callStackTrace: CallStackTrace, val methodName: String): TracePoint(iThread, actorId, callStackTrace) {
```
eupp, line 525: Let's rename into one of the following:
- `MethodResultTracePoint`
- `MethodReturnTracePoint`
- `MethodExitTracePoint`

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1636,26 +1626,26 @@ abstract class ManagedStrategy(
...
         if (methodSection == AnalysisSectionType.ATOMIC &&
             // do not create a trace point on resumption
             !(isTestThread(threadId) && isResumptionMethodCall(threadId, className, methodName, params, atomicMethodDescriptor))
         ) {
-            // re-use last call trace point
-            val methodCallTracePoint = callStackTrace[threadId]!!.lastOrNull()?.tracePoint
             newSwitchPoint(threadId, codeLocation, beforeMethodCallSwitch = true)
-            traceCollector?.addTracePointInternal(methodCallTracePoint)
             loopDetector.passParameters(params)
         }
```
eupp, line 1639: Could you please rebase on `develop`?
When refactoring loop detector in this PR:
https://github.com/JetBrains/lincheck/pull/656

I made a small change here, to clearly separate cases of atomic/regular method calls.

src/jvm/test/resources/expected_logs/method_handles/bind_to.txt
```diff
@@ -19,15 +19,20 @@ Detailed trace:
...
 |         Counter#1.value = 1 at Counter.increment(MethodHandlesLookupTests.kt:158)                                                                                               |
 |       methodHandle.invokeExact() at MethodHandlesBindToRepresentationTest.doTest(MethodHandlesLookupTests.kt:140)                                                               |
 |         Counter#1.value ➜ 1 at Counter.increment(MethodHandlesLookupTests.kt:158)                                                                                               |
 |         Counter#1.value = 2 at Counter.increment(MethodHandlesLookupTests.kt:158)                                                                                               |
 |       counter.value ➜ 2 at MethodHandlesBindToRepresentationTest.doTest(MethodHandlesLookupTests.kt:141)                                                                        |
+|     "Check failed.".toString(): "Check failed." at BaseMethodHandleLookupRepresentationTest.block(MethodHandlesLookupTests.kt:58)                                               |
```
eupp, line 36: Why these lines started to appear in the trace?

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -796,15 +787,6 @@ abstract class ManagedStrategy(
...
-        if (collectTrace) {
-            val tracePoint = ThreadJoinTracePoint(
-                iThread = currentThreadId,
-                actorId = currentActorId[currentThreadId]!!,
-                joinedThreadDisplayNumber = iThreadToDisplayNumber(joinThreadId),
-                callStackTrace = callStackTrace[currentThreadId]!!,
-            )
-            traceCollector!!.addTracePointInternal(tracePoint)
-        }
```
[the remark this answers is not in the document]
ivandev0: Same as ThreadStartTracePoint

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -705,22 +705,13 @@ abstract class ManagedStrategy(
...
-        val forkedThreadId = registerThread(thread, descriptor)
-        if (collectTrace) {
-            val tracePoint = ThreadStartTracePoint(
-                iThread = currentThreadId,
-                actorId = currentActorId[currentThreadId]!!,
-                startedThreadDisplayNumber = iThreadToDisplayNumber(forkedThreadId),
-                callStackTrace = callStackTrace[currentThreadId]!!,
-            )
-            traceCollector!!.addTracePointInternal(tracePoint)
-        }
```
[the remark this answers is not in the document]
eupp: So after this change, do we need `ThreadStartTracePoint` and `ThreadJoinTracePoint` at all? Can we remove these classes then?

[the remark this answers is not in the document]
ivandev0: I would say we can remove them. I didn't do it yet because it is a draft and maybe we want to create a secial renderer for start and join (like it was with custom nodes)

eupp (approved):

eupp:
Summary of the PR:
- Adds `MethodReturnTracePoint`. 
- Stores `MethodCallTracePoint` and `MethodReturnTracePoint` directly in the trace, similarly as other trace point types.
- Collects "flat" trace, tree structure is restored later at trace post-processing stage. 
- Do not store call stack traces in each trace point, removes `callStackTrace` field from almost all trace points type (except `SwitchEventTracePoint`, `SpinCycleStartTracePoint`, and `ObstructionFreedomViolationExecutionAbortTracePoint` --- these stack traces are still required for the current loop detector algorithm implementation).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -808,15 +794,6 @@ internal abstract class ManagedStrategy(
             // Switch to another thread and wait for a moment when the thread is finished
             switchCurrentThread(currentThreadId, BlockingReason.ThreadJoin(joinThreadId))
         }
-        if (collectTrace) {
-            val tracePoint = ThreadJoinTracePoint(
```
bbrockbernd, line 812: I suppose thread joins are now added as normal function calls?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -994,11 +971,40 @@ internal abstract class ManagedStrategy(
...
+        val actor = if (actorId < scenario.threads[iThread].size) scenario.threads[iThread][actorId]
+        else validationFunction
+        check(actor != null) { "Could not find current actor" }
+
+        val lastTracePoint = traceCollector?.trace?.lastOrNull()
+
+        if (lastTracePoint !is MethodReturnTracePoint
+            || lastTracePoint.methodTracePoint.methodName != actor.method.name
+            || lastTracePoint.methodTracePoint.className != actor.method.declaringClass.name) {
+            // TODO rewrite. This will create an additional return trace point if actor is a suspend function
```
bbrockbernd: Seems like an important TODO

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -82,7 +86,193 @@ internal class TraceReporter(
...
+     *    switch
+     *    ...
+     *    foo()
+     *      bar()
+     *        x = 42
+     *    ```
+     *
+     * @return A new trace instance with updated trace point ordering.
+     */
+    private fun Trace.moveStartingSwitchPointsOutOfMethodCalls(): Trace {
```
bbrockbernd, line 114: @eupp Do we need this logic for spin cycle detection? If not, wouldn't this be much easier to do after constructing the tree? And if so, what changed I think before we did it after spin cycle detection?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Collections.kt
```diff
@@ -10,6 +10,95 @@
...
+/**
+ * Finds the index of the first element that matches the given [predicate]
+ * starting from the specified [from] index.
+ *
+ * @param from the index from which to start searching. Must be within the bounds of the list.
+ * @param predicate a lambda function to test each element for a condition.
+ * @return the index of the first matching element, or -1 if no elements match the condition.
+ * @throws IllegalArgumentException if [from] is out of bounds.
+ */
+fun <T> List<T>.indexOf(from: Int, predicate: (T) -> Boolean): Int {
```
bbrockbernd: Should these be internal? Idem for all others in this file

bbrockbernd (approved):
Mostly LGTM, just some questions and left TODOS

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ivandev0: Yes they are

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
ivandev0: Yeah. We checked it and have no idea how to fix it for now. The problem is a little bit different than stated in the message (I rewrote it in the last commit). When the thread finishes as aborted, we have strange state when actor wasn't started, but we are trying to finish it. We decided to handle it later.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
eupp: > Do we need this logic for spin cycle detection?

Yes, we need it unfortunately. For instance, if the two trace transformations are re-ordered, the tests do not pass.

> If not, wouldn't this be much easier to do after constructing the tree?

Yes, some trace transformations would be easier to do on tree representation (e.g., moving an event out of its parent method calls). 

At the same time, some other trace transformations might be not so easy. 
For instance, here when we move thread switch trace point, we also need to move some trace points to the next block of events of the corresponding thread. In essence, it involves changing the global positioning of events within trace. 

I think we can return to this problem later. 
Once we figure out what primitive trace transformations exactly do we need, then we can craft the trace representation and API that would be more convenient to use.

eupp (approved):

PR #672, opened 2025-06-04
Collect actor results during trace collection

bbrockbernd:
Refactor actor results being collected in onActorFinish.

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
@@ -284,13 +284,13 @@ internal class MethodCallTracePoint(
     
     private fun StringBuilder.appendActor() {
         append("$methodName(${ parameters?.joinToString(", ") ?: "" })")
-        if (returnedValue != ReturnedValueResult.NoValue && returnedValue.actorRepresentationConfig.showAtBeginningOfActor) {
-            append(": ${returnedValue.actorRepresentationConfig.resultRepresentation}")
+        if (returnedValue != ReturnedValueResult.NoValue && returnedValue.showAtMethodCallBeginning) {
+            append(": ${returnedValue.resultRepresentation}")
         }
         
         // We only show hung for actors (not for thread starts)
         if (returnedValue == ReturnedValueResult.NoValue && isActor) {
-            append(": ${returnedValue.actorRepresentationConfig.resultRepresentation}")
+            append(": ${returnedValue.resultRepresentation}")
         }
     }
         
@@ -393,54 +393,40 @@ internal class MethodCallTracePoint(
 }
 
 
-internal sealed class ReturnedValueResult(
-    val actorRepresentationConfig: ActorResultRepresentationConfig
-) {
-    data object NoValue: ReturnedValueResult(HungActorResultConfig)
-    data object VoidResult: ReturnedValueResult(VoidActorResultConfig)
+internal sealed class ReturnedValueResult() {
+    abstract val resultRepresentation: String
+
+    open val showAtMethodCallBeginning: Boolean = true
+    open val showAtMethodCallEnd: Boolean = true
+
+    data object NoValue : ReturnedValueResult() {
+        override val resultRepresentation = "<hung>" // TODO: this is a hack, add special `HungResult` in the future
+        override val showAtMethodCallBeginning = false
+        override val showAtMethodCallEnd = false
+    }
+
+    data object VoidResult : ReturnedValueResult() {
+        override val resultRepresentation = "void"
+        override val showAtMethodCallBeginning = false
+        override val showAtMethodCallEnd = true
+    }
     
     // Should not be possible for actor
-    data object CoroutineSuspended: ReturnedValueResult(VoidActorResultConfig)
-    data class ValueResult(val valueRepresentation: String, val valueType: String): ReturnedValueResult(ValueActorResultConfig(valueRepresentation))
+    data object CoroutineSuspended : ReturnedValueResult() {
+        override val resultRepresentation = "<suspended>"
+        override val showAtMethodCallBeginning = false
+        override val showAtMethodCallEnd = true
+    }
+
+
+    data class ValueResult(val valueRepresentation: String, val valueType: String) : ReturnedValueResult() {
+        override val resultRepresentation = valueRepresentation
+    }
     
     // Only for actors
-    data class ExceptionResult(val exceptionRepresentation: String): ReturnedValueResult(ExceptionActorResultConfig(exceptionRepresentation)) {
-        var exceptionNumber: Int
-            get() = (actorRepresentationConfig as ExceptionActorResultConfig).exceptionNumber
-            set(value) { (actorRepresentationConfig as ExceptionActorResultConfig).exceptionNumber = value }
+    data class ExceptionResult(val exceptionRepresentation: String, var exceptionNumber: Int = -1) : ReturnedValueResult() {
+        override val resultRepresentation get() = "$exceptionRepresentation #$exceptionNumber"
     }
-}
-
-internal sealed interface ActorResultRepresentationConfig {
-    val resultRepresentation: String
-    val showAtBeginningOfActor: Boolean
-    val showAtEndOfActor: Boolean
-}
-
-internal object HungActorResultConfig:  ActorResultRepresentationConfig {
-    override val resultRepresentation: String = "<hung>"
-    override val showAtBeginningOfActor: Boolean = true
-    override val showAtEndOfActor: Boolean = false
-}
-
-
-internal class ValueActorResultConfig(result: String): ActorResultRepresentationConfig {
-    override val resultRepresentation: String = result
-    override val showAtBeginningOfActor: Boolean = true
-    override val showAtEndOfActor: Boolean = true
-}
-
-internal object VoidActorResultConfig:  ActorResultRepresentationConfig{
-    override val resultRepresentation = "void"
-    override val showAtBeginningOfActor = false
-    override val showAtEndOfActor = true
-}
-
-internal class ExceptionActorResultConfig(private val result: String): ActorResultRepresentationConfig {
-    override val resultRepresentation get() = "$result #$exceptionNumber"
-    override val showAtBeginningOfActor = true
-    override val showAtEndOfActor = true
-    var exceptionNumber = -1
 }
 
 internal class MonitorEnterTracePoint(
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
@@ -284,7 +284,7 @@ internal fun flattenedTraceGraphToCSV(nodeList: SingleThreadedTable<TraceNode>):
                 is ResultNode -> {
                     val beforeEventId = -1
                     val type = TracePointType.RESULT
-                    val representation = node.actorResult.actorRepresentationConfig.resultRepresentation
+                    val representation = node.actorResult.resultRepresentation
                     val exceptionNumber = (node.actorResult as? ReturnedValueResult.ExceptionResult)?.exceptionNumber ?: -1
                     "${type.ordinal};${node.iThread};${node.callDepth};${preExpandedNodeSet.contains(node)};${beforeEventId};${representation};${exceptionNumber};null;-1;[];false"
                 }
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceFlattenPolicies.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceFlattenPolicies.kt
@@ -64,7 +64,7 @@ internal class VerboseTraceFlattenPolicy : TraceFlattenPolicy {
                     currentNode.tracePoint.isActor) return emptyList()
                 
                 // Check if result node should be added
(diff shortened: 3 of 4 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -996,7 +996,13 @@ abstract class ManagedStrategy(
         enableAnalysis()
     }
 
-    override fun onActorFinish() {
+    override fun onActorFinish(iThread: Int) = runInsideIgnoredSection {
+        runner.getActorResult(iThread, currentActorId[iThread]!!)?.let { result ->
+            traceCollector?.trace
+                ?.filterIsInstance<MethodCallTracePoint>()
+                ?.firstOrNull { it.isActor && it.actorId == currentActorId[iThread] && it.iThread == iThread }
+                ?.initializeActorResult(result)
```
eupp: Why can't you use `callStackTrace` instead to obtain the actor method call trace point?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -996,7 +996,13 @@ abstract class ManagedStrategy(
         enableAnalysis()
     }
 
-    override fun onActorFinish() {
+    override fun onActorFinish(iThread: Int) = runInsideIgnoredSection {
```
eupp, line 971: Could you please then also add `runInsideIgnoredSection` into `onActorStart` for consistency?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectLabelFactory.kt
```diff
@@ -115,4 +115,5 @@ object ObjectLabelFactory {
 /**
  * Since lambdas do not have a specific type, this class is used as key for [ObjectLabelFactory.getObjectNumber].
  */
-private class Lambda
\ No newline at end of file
+private class Lambda
+private class ExceptionResult
```
eupp: Do you still use this? If not, please remove.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -150,9 +150,15 @@ fun Strategy.verify(result: InvocationResult, verifier: Verifier): LincheckFailu
         is SpinCycleFoundAndReplayRequired -> null
         is CompletedInvocationResult ->
             if (!verifier.verifyResults(scenario, result.results)) {
-                IncorrectResultsFailure(scenario, result.results, tryCollectTrace(result), AnalysisProfile(testCfg))
+                tryCollectTrace(result).let { (trace, traceResult) -> IncorrectResultsFailure(
```
eupp: Instead of `let`, let's just use:
`val (trace, traceResult) = ...`
here.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/Strategy.kt
```diff
@@ -150,9 +150,15 @@ fun Strategy.verify(result: InvocationResult, verifier: Verifier): LincheckFailu
...
+                        scenario = scenario,
+                        results = (traceResult as CompletedInvocationResult).results,
+                        trace = trace,
+                        analysisProfile = AnalysisProfile(testCfg)
+                    )
+                }
             } else null
         else -> 
-            result.toLincheckFailure(scenario, tryCollectTrace(result), AnalysisProfile(testCfg))
+            tryCollectTrace(result).let { (trace, traceResult) -> traceResult.toLincheckFailure(scenario, trace, AnalysisProfile(testCfg)) }
```
eupp: Instead of `let`, let's just use:
`val (trace, traceResult) = ...`
here.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -334,6 +335,19 @@ internal class MethodCallTracePoint(
     fun initializeOwnerName(ownerName: String) {
         this.ownerName = ownerName
     }
+    
+    fun initializeActorResult(result: Result?) {
+        check(isActor)
+        this.returnedValue = when (result) {
+            null -> ReturnedValueResult.VoidResult
+            is ExceptionResult -> when (result.throwable) {
+                is LincheckAnalysisAbortedError -> ReturnedValueResult.ActorHungResult
```
eupp: Is it always true that `LincheckAnalysisAbortedError` is thrown only in case of a hung? I think in general, it can also mean other things --- the actor could have been aborted for various reasons.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -370,22 +384,40 @@ internal sealed interface ReturnedValueResult {
...
+        override val showAtBeginningOfActor = true
+        override val showAtEndOfActor = false
+    }
+
+    data class ActorValueResult(override val resultRepresentation: String): ActorResult {
+        override val showAtBeginningOfActor = true
+        override val showAtEndOfActor = true
+    }
+    
+    data object ActorVoidResult: ActorResult {
```
eupp: We have a duplication writ to `VoidResult`, `ValueResult`, etc. 
Maybe we don't need `ActorResult` as separate sub-class? 

Any additional actor-specific data (like `showAtBeginningOfActor` and `showAtEndOfActor`) we can pass as additional object of class `actorData: ActorResultData` (or something like that) --- in addition we can decide if method result is actor result based on if `actorData != null`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -370,22 +384,40 @@ internal sealed interface ReturnedValueResult {
...
+        override val resultRepresentation = "void"
+        override val showAtBeginningOfActor = false
+        override val showAtEndOfActor = true
+    }
+    
+    data class ActorExceptionResult(val exceptionResultRepresentation: String): ActorResult {
+        override val resultRepresentation: String get() = "$exceptionResultRepresentation #$excNumber" 
+        override val showAtBeginningOfActor = true
+        override val showAtEndOfActor = true
+        var excNumber = -1
```
eupp: please keep the name `exceptionNumber`

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -27,34 +26,30 @@ internal typealias Column<T> = List<T>
 @Synchronized // we should avoid concurrent executions to keep `objectNumeration` consistent
 internal fun StringBuilder.appendTrace(
     failure: LincheckFailure,
-    results: ExecutionResult,
     trace: Trace,
     exceptionStackTraces: Map<Throwable, ExceptionNumberAndStacktrace>,
```
eupp, line 33: Do we still use `exceptionStackTraces`? 
If we assign exception numbers in `numberExceptionResults`, maybe we can get rid of it?

src/jvm/test/resources/expected_logs/validation_function_failure.txt
```diff
@@ -18,24 +18,24 @@ All operations above the horizontal line | ----- | happen before those below the
...
+|                   Thread 1                    |  Thread 2   |
+| ----------------------------------------------------------- |
+| operation()                                   |             |
+| ----------------------------------------------------------- |
+|                                               | operation() |
+| operation()                                   |             |
+| ----------------------------------------------------------- |
+| operation()                                   |             |
+| ----------------------------------------------------------- |
+| validateWithError(): IllegalStateException #1 |             |
```
eupp, line 37: Should we print exception number near validation exception stack trace (similarly as we do for other types of exceptions) ?

eupp (changes requested):
Good job overall!
Left few comments and request.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
bbrockbernd: callStackTrace happens to be empty at this point in the program

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: If that is the case we do not test for it. For what I understand from the code base threads can be aborted on live/deadlocks or when there is an internal error. In the last case I think we only show the stacktrace of the internal error and not the execution trace. So I don't think it matters.

Could you provide an example where LincheckAbortedError is the result of one of the actors in the trace and the execution was not dead/live locekd?

I am open to suggestions if you think there are better ways of detecting hung actors

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
bbrockbernd: We still use it for `appendExceptionsStackTraces`

src/jvm/test/resources/expected_logs/validation_function_failure.txt
bbrockbernd: commit `Add number to validation function stack trace.` solves this issue. However, I don't really like the solution. Also, in GPMC mode we have the same problem look at `access_field_representation_test.txt`

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: I agree there is some redundancy. However, we would also have to add ways to distinguish exception result and hung result from value results. As this influences adding results nodes during trace flattening.

Additionally, when we add an `actorData` field to `VoidResult` (and others) would require us to create seperate result instances for each method (which are currently reusing the same objects in case of void or no result). To solve this we could have one `VoidResult` instance for all methods and one for all actors. But then we would be back at the the same duplication problem stated in your comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
eupp: > Could you provide an example where LincheckAbortedError is the result of one of the actors in the trace and the execution was not dead/live locked?

I do not have concrete example, just uneasy feeling. However, I agree, current architecture of actors & aborting does not allow us easily implement a better approach, so let's leave it as is for now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
eupp: Ok, let's leave as is for now. Could you please add a `TODO` though, to unify it later.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: Ok, it is going to be called once per-actor, so probably no big deal.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -245,7 +245,16 @@ internal class MethodCallTracePoint(
     val callType: CallType = CallType.NORMAL,
     val isSuspend: Boolean
 ) : CodeLocationTracePoint(eventId, iThread, actorId, codeLocation) {
-    var returnedValue: ReturnedValueResult = ReturnedValueResult.NoValue
+    
+    private var myReturnedValue: ReturnedValueResult? = null
```
eupp: The naming convention in Kotlin in this case would be `_returnedValue` and `returnedValue` for private/public properties respectively.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -245,7 +245,16 @@ internal class MethodCallTracePoint(
...
+    var returnedValue: ReturnedValueResult 
+        get() {
+            return if (myReturnedValue != null) myReturnedValue!!
+            else if (isActor) ReturnedValueResult.ActorHungResult
```
eupp: Please add a comment to explain the logic here (and, perhaps, a TODO?)

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: This is removed after last refactor

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: This is also removed after last refactor

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: Should be unified in last commit

eupp (approved):

PR #677, opened 2025-06-06
First Trace Recorder implementation

lev-serebryakov-jetbrains:
To run something with one method traced run:

```sh
java -Dlincheck.traceRecorderMode=true -javaagent:${path-to-lincheck-fat-jar}=${class},${method},${output} YourRunnerClass 
```

This implementation dumps trace in same format as IdeaPlugin consume for now.

It can overinstrument code (lincheck agent is installed early) and sometimes (i.e. when run from `gradle`) provide some unneeded views inside of standard library, as lincheck filters out system classes on too high level (via `ensureClassHierarchyIsTransformed()`) and it happily transforms classes when called directly by instrumentation API, for example for classes de-serialized  by gradle.

Contains some (significant) code duplication with `ManagedStrategy` around processing of method calls.

PR #678, opened 2025-06-06
Refactor static agent for trace debugger

dmitrii-artuhov:


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/ModelCheckerTest.kt
```diff
@@ -32,12 +32,12 @@ internal fun modelCheckerTest(
         parallel { thread { actor(testOperation) } }
     }
     val verifier = CollectResultsVerifier()
-    withLincheckJavaAgent(InstrumentationMode.MODEL_CHECKING) {
+    withLincheckDynamicJavaAgent(InstrumentationMode.MODEL_CHECKING) {
```
ndkoval: Note: let's think about the function name

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/ModelCheckerTest.kt
```diff
@@ -32,12 +32,12 @@ internal fun modelCheckerTest(
...
         val strategy = createStrategy(testClass.java, scenario, stdLibAnalysis)
         val failure = strategy.runIteration(invocations, verifier)
         if (expectedFailure != null) {
             assert(expectedFailure.isInstance(failure))
-            return
+            return@withLincheckDynamicJavaAgent
```
dmitrii-artuhov: `@...` not needed anymore, since `withLincheckDynamicJavaAgent` is inline

buildSrc/src/main/kotlin/TraceDebuggerTasks.kt
```diff
@@ -48,7 +48,7 @@ fun Project.registerTraceDebuggerTasks() {
         manifest {
             attributes(
                 mapOf(
-                    "Premain-Class" to "org.jetbrains.kotlinx.lincheck.transformation.TraceDebuggerAgent",
```
ndkoval, line 51: Let's rename with the Trace Recorder addition

src/jvm/main/org/jetbrains/kotlinx/lincheck/Lincheck.kt
```diff
@@ -69,8 +69,8 @@ object Lincheck {
             .verifier(NoExceptionVerifier::class.java)
 
         val testCfg = options.createTestConfigurations(GeneralPurposeModelCheckingWrapper::class.java)
-        withLincheckJavaAgent(testCfg.instrumentationMode) {
-            ensureObjectIsTransformed(block)
+        withLincheckDynamicJavaAgent(testCfg.instrumentationMode) {
+            LincheckJavaAgent.ensureObjectIsTransformed(block)
```
ndkoval: Please revert the static import

src/jvm/main/org/jetbrains/kotlinx/lincheck/TraceDebugger.kt
```diff
@@ -115,7 +115,7 @@ internal object TraceDebuggerInjections {
 
 internal class TraceDebuggerStaticMethodWrapper {
     fun callStaticMethod(clazz: Class<*>, method: Method) {
-        ensureClassHierarchyIsTransformed(clazz.name)
+        LincheckJavaAgent.ensureClassHierarchyIsTransformed(clazz.name)
```
ndkoval: please revert

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgents.kt
```diff
@@ -71,16 +55,55 @@ internal enum class InstrumentationMode {
 }
 
 /**
- * LincheckJavaAgent represents the Lincheck Java agent responsible for instrumenting bytecode.
+ * [LincheckStaticJavaAgent] is an agent that is used by `TRACE_DEBUGGER` and `TRACE_RECORDER` modes of Lincheck.
+ * It appends [TraceDebuggerTransformer] in the transformers chain and then installs the agent once
+ * in the `MODEL_CHECKING` mode and never performs `uninstall`.
+ */
+internal object LincheckStaticJavaAgent : CachingJavaAgent by LincheckJavaAgent {
```
ndkoval: With "TraceDebugger" everywhere in the code, should we rename it to "TraceDebuggerJavaAgent"?

dmitrii-artuhov: renamed to TraceDebugger**Java**Agent

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgents.kt
```diff
@@ -71,16 +75,44 @@ internal enum class InstrumentationMode {
 }
 
 /**
- * LincheckJavaAgent represents the Lincheck Java agent responsible for instrumenting bytecode.
+ * [TraceDebuggerJavaAgent] is an agent that is used by `TRACE_DEBUGGER` and `TRACE_RECORDER` modes of Lincheck.
+ * It appends [TraceDebuggerTransformer] in the transformers chain and then installs the agent once
+ * in the `MODEL_CHECKING` mode and never performs `uninstall`.
+ */
+internal object TraceDebuggerJavaAgent {
```
eupp, line 82: Let's keep it in separate file.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgents.kt
```diff
@@ -11,48 +11,52 @@
...
-import java.security.ProtectionDomain
 import java.util.*
-import java.util.concurrent.ConcurrentHashMap
 import java.util.jar.JarFile
 
 
 /**
  * Executes [block] with the Lincheck java agent for byte-code instrumentation.
  */
 internal inline fun withLincheckJavaAgent(instrumentationMode: InstrumentationMode, block: () -> Unit) {
```
eupp, line 33: If we change the java-agent attach from dynamic to static in trace debugger mode, why do we need to call `withLincheckJavaAgent` at all in this case?

eupp (changes requested):
What is our plan wrt. synching this PR with https://github.com/JetBrains/lincheck/pull/677? 

Besides, I would prefer to not modify the old `LincheckJavaAgent` at all --- let's use it only for dynamic agent attach scenario. Also, due to current architectural limitations, we probably would not be able to switch to static agent easily in case of trace debugger.   

Let's consider adding static agent only for trace recorder case.

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgents.kt
dmitrii-artuhov: In fact we don't. withLincheckJavaAgent is called from many places in the source code (e.g. in tests), thus, it is made non-transparent for the client which type of agent is attached: static or dynamic. Otherwise, I would need to write ifs everywhere, whether I am using dynamic or static agent and call or don't call withLincheckJavaAgent

dmitrii-artuhov:
@eupp, hi, I have a couple of questions regarding your comment above
- Not to modify the LincheckJavaAgent: its logic is not modified. The only difference between static and dynamic agents in our implementation is how the Instrumentation object is acquired: from ByteBuddy or from the premain method. The rest of the functionality is identical for both agents
- Not be able to switch to static agent easily: what do you mean here? Plugin for trace debugger works with this PR, the static agent does not use dymanic attach anymore: it calls LincheckJavaAgent::install(MODEL_CHECKING) once and that is it. In the method withLincheckJavaAgent it only asserts that MODEL_CHECKING mode is used

src/jvm/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgents.kt
```diff
@@ -11,48 +11,52 @@
...
-    LincheckJavaAgent.install(instrumentationMode)
-    return try {
+    // in case if trace debugger attached its static agent we don't do anything
+    if (isTraceDebuggerJavaAgentAttached) {
         block()
-    } finally {
-        LincheckJavaAgent.uninstall()
+    }
+    // otherwise we perform instrumentation via ByteBuddy dynamic agent
+    else {
```
dmitrii-artuhov, line 39: } else {
// comment here

dmitrii-artuhov:
Added as part of https://github.com/JetBrains/lincheck/pull/677 via separate commits

PR #679, opened 2025-06-10
Configure sonatype publishing

eupp:


build.gradle.kts
```diff
@@ -380,11 +404,29 @@ publishing {
             artifact(sourcesJar.get()) {
                 classifier = "sources"
             }
+            // TODO: we probably will have to configure and add javadoc, as it seemingly required by Maven Central
+            // artifact(tasks.javadoc)
```
ndkoval: Please note that we need to keep the following code in `build.gradle.kts`:

```
    replace("jvmSourcesJar", Jar::class).run {
        from(sourceSets["main"].allSource)
        // Also collect sources for the injected classes to simplify debugging
        from(project(":bootstrap").file("src"))
    }
```

ndkoval: I'm afraid we need to add javadoc.

build.gradle.kts
```diff
@@ -380,11 +404,29 @@ publishing {
...
+                url.set("https://github.com/JetBrains/lincheck")
+                scm {
+                    connection.set("scm:git:https://github.com/JetBrains/lincheck.git")
+                    url.set("https://github.com/JetBrains/lincheck")
+                }
+
+                developers {
+                    developer {
+                        this.name.set("Nikita Koval")
+                        id.set("ndkoval")
```
ndkoval: If you want to configure the publication with my account, the correct id is `nikita.koval`

build.gradle.kts
```diff
@@ -1,11 +1,27 @@
...
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.asRequestBody
+import java.util.*
+
+buildscript {
+    repositories {
+        maven { url = uri("https://packages.jetbrains.team/maven/p/jcs/maven") }
+    }
+    dependencies {
+        classpath("com.jetbrains:jet-sign:45.47")
```
ndkoval, line 17: Should we put these plugin versions to `gradle.properties`?

build.gradle.kts
eupp: > Please note that we need to keep the following code in build.gradle.kts

It is there, only the name of the task has changed to `sourcesJar`.

> I'm afraid we need to add javadoc.

Added it via Dokka.

build.gradle.kts
eupp: `buildscript { ... }` block is evaluated earlier than the rest of the script, and thus have different semantics. 
For instance, the usual syntax `val okhttpVersion by project` does not work inside `buildscript`. 

I've tried to google/ask-AI the problem and got conflicting info on whether it is possible to access `gradle.properties` inside `buildscript` and what is the syntax. 
Perhaps it is possible, but since the solution is not trivial, I suggest leaving it as is. 

It's just two dependencies that are only used for signing and publishing. 
Later if we will experience some problems around updating these dependencies, we can revisit this and modify the script.

eupp:
I resolved previous comments

ndkoval:
@eupp please rebase onto `develop`

ndkoval:
Before merging the PR, please make sure the artifact is similar to the previous one when you publish it to maven local.

eupp:
> Before merging the PR, please make sure the artifact is similar to the previous one when you publish it to maven local.

Checked the generated structured. 
It mostly matches the one we had before (e.g., the lincheck-2.39 published to Sonatype). 
The only missing file(s) is `lincheck-2.39-kotlin-tooling-metadata.json`, which is probably related to Kotlin Multiplatform. 

Besides that, the artifact has:
- `lincheck-%version.jar`
- `lincheck--%version-javadoc.jar`
- `lincheck--%version-sources.jar`
- `lincheck-%version.pom`
- `lincheck-%version.module`
- `md5`, `sha1`, `sha256`, and `sha512` hashes of the above files
- `asc` files are missing because signing is not yet configured

Issue #682, opened 2025-06-11
[Bug] Tracecompression fails when return types of default pairs are not equal.


bbrockbernd:
```Kotlin
class InstantParsing {
    @Test
    fun testParsingInvalidInstants() = TraceRecorder.recordTraceInternal("bla.txt") {
        fun parseInstantLikeJavaDoes(input: String): Instant =
            DateTimeComponents.Formats.ISO_DATE_TIME_OFFSET.parse(input).apply {
                when {
                    hour == 24 && minute == 0 && second == 0 && nanosecond == 0 -> {
                        setDate(toLocalDate().plus(1, DateTimeUnit.DAY))
                        hour = 0
                    }
                    hour == 23 && minute == 59 && second == 60 -> second = 59
                }
            }.toInstantUsingOffset()
        fun formatTwoDigits(i: Int) = if (i < 10) "0$i" else "$i"
        val hour = 23
        val minute = 59
        val second = 59
        val input = "2020-03-16T${hour}:${formatTwoDigits(minute)}:${formatTwoDigits(second)}Z"
        assertEquals(
            runCatching { java.time.Instant.parse(input) }.getOrNull(),
            runCatching { parseInstantLikeJavaDoes(input).toJavaInstant() }.getOrNull()
        )
    }

}
```

has the following (partial) result

```
|     ISO_DATE_TIME_OFFSET.parse("2020-03-16T23:59:59Z"): Instant#3 at InstantParsing.testParsingInvalidInstants$lambda$3$parseInstantLikeJavaDoes(TraceRecorderInternalApiTest.kt:40)                                                                  |
|       Parser.match-impl$default(ParserStructure#1, "2020-03-16T23:59:59Z", DateTimeComponentsContents#1, 0, 4, null): DateTimeComponents#1 at AbstractDateTimeFormat.parse(DateTimeFormat.kt:134)                                                     |
|         Parser.match-impl(ParserStructure#1, "2020-03-16T23:59:59Z", DateTimeComponentsContents#1, 0): DateTimeComponentsContents#10 at Parser.match-impl$default(Parser.kt:186)                                                                      |

```

eupp:
Migrated to YT

PR #683, opened 2025-06-11
Bugfixes and trace improvements during hackathon

bbrockbernd:
- Different return types from default method pair #682
- Several trace improvements, mostly removed $ signs which are added by kotlin compiler.


src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -103,16 +117,20 @@ private fun SingleThreadedTable<TraceNode>.compressAccessPairs() = compressNodes
  * For more details check [isDefaultPair] and [isAccessPair].
  */
 private fun combineNodes(parent: CallNode, child: CallNode): TraceNode {
-    parent.tracePoint.methodName = child.tracePoint.methodName
-    parent.tracePoint.parameters = child.tracePoint.parameters
-
-    check(parent.tracePoint.returnedValue == child.tracePoint.returnedValue)
     check(parent.tracePoint.thrownException == child.tracePoint.thrownException)
+    // TODO investigate why in rare cases return values are not equal #682
+    if (parent.tracePoint.returnedValue == child.tracePoint.returnedValue) {
```
eupp: Shouldn't it be:
```kotlin
if (parent.tracePoint.returnedValue == child.tracePoint.returnedValue &&
    parent.tracePoint.thrownException == child.tracePoint.thrownException) 
{ ... }
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -21,6 +21,20 @@ internal fun SingleThreadedTable<TraceNode>.compressTrace() = this
     .compressUserThreadRun()
     .compressThreadStart()
     .removeCoroutinesCoreSuffix()
+    .compressInlineIV()
+    .compressDollarThis()
+    .replaceNestedClassDollar()
+    .compressSyntheticParameterNumbers()
+    .compressLambdaCaptureSyntheticField()
+    .compressVolatileDollar()
```
eupp, line 28: For the transformations that alter only the `ownerName` of the trace points, I have alternative suggestions on how to approach them. Let's consider them case-by-case to find the best option. See comments below.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -189,6 +207,121 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
...
+    node.tracePoint.ownerName = node.tracePoint.ownerName!!.removeSuffix("\$iv")
+    node
+}
+
+/**
+ * Removes $this. owner names.
+ */
+private fun SingleThreadedTable<TraceNode>.compressDollarThis() = compressNodes { node ->
+    if (node !is CallNode || node.tracePoint.ownerName == null) return@compressNodes node
+    if (node.tracePoint.ownerName == "\$this") node.tracePoint.ownerName = null
```
eupp: What if instead we will filter `$this` local variable from the local variables owner name resolution procedure?

You will need to modify `ShadowStack.getLastAccessVariable` method (it already has some examples of filters).

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -189,6 +207,121 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
...
+    node
+}
+
+/**
+ * Removes nested class dollars from owner names.
+ */
+private fun SingleThreadedTable<TraceNode>.replaceNestedClassDollar() = compressNodes { node ->
+    if (node is CallNode && node.tracePoint.ownerName != null) {
+        val newOwner = fixNestedClassDollar(node.tracePoint.ownerName!!)
```
eupp, line 314: I believe nested dollars signs in class names are due to the difference in `name` and `canonicalName` class name formats of inner classes in JVM:
https://www.baeldung.com/java-class-name#3-inner-classes

Currently, we get class names by converting internal class names (obtained from ASM) by substitution of `/` characters to `.` (see `toCanonicalClassName()` function):
https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2

But despite its name, it looks like our `toCanonicalClassName()` actually returns name in normal fully-qualified form (i.e., the one given by `Class::name` property, not `Class::canonicalName`).  
I think we rely on this implementation, as we sometimes want to pass the resulting name to `Class.forName(...)` function, which expects the class name in this format. 

But, to the point. I think, we can try to convert the class name to expected format directly upon its initialization of the corresponding field in the `TracePoint` class (i.e., `ownerName`).

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -189,6 +207,121 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
...
+        && node.tracePoint.ownerRepresentation?.startsWith("$") == true
+        && node.tracePoint.fieldName == "element"
+    ) {
+        node.tracePoint.fieldName = node.tracePoint.ownerRepresentation!!.removePrefix("$")
+        node.tracePoint.ownerRepresentation = null
+    }
+    node
+}
+
+private fun SingleThreadedTable<TraceNode>.compressVolatileDollar() = compressNodes { node ->
```
eupp, line 380: Please add documentation explaining what this function does and an example of before/after trace change. Also, add a link to the corresponding issue: https://github.com/JetBrains/lincheck/issues/687

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -189,6 +207,121 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
     newNode
 }
 
+/**
+ * Remove $iv from owner names, due to inline functions.
+ */
+private fun SingleThreadedTable<TraceNode>.compressInlineIV() = compressNodes { node ->
+    if (node !is CallNode || node.tracePoint.ownerName == null) return@compressNodes node
+    node.tracePoint.ownerName = node.tracePoint.ownerName!!.removeSuffix("\$iv")
```
eupp: Would it make sense to filter `$iv` directly in `TracePoint` upon initialization of `ownerName`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -21,6 +21,20 @@ internal fun SingleThreadedTable<TraceNode>.compressTrace() = this
     .compressUserThreadRun()
     .compressThreadStart()
     .removeCoroutinesCoreSuffix()
+    .compressInlineIV()
+    .compressDollarThis()
+    .replaceNestedClassDollar()
+    .compressSyntheticParameterNumbers()
+    .compressLambdaCaptureSyntheticField()
+    .compressVolatileDollar()
+
```
eupp, line 29: Besides, could you please for each transformation, add a documentation comment, containing:
- explanation of what this transformation does;
- example of trace point string before/after transformation; 
- link to the test testing this case (if present).

See, `compressDefaultPairs` documentation for example.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -153,10 +153,10 @@ internal class ObstructionFreedomViolationExecutionAbortTracePoint(
 }
 
 internal class ReadTracePoint(
-    private val ownerRepresentation: String?,
+    var ownerRepresentation: String?,
     iThread: Int, actorId: Int,
     callStackTrace: CallStackTrace,
-    private val fieldName: String,
```
eupp, line 159: Let' try to keep it `private` and `val`, and apply all necessary transformations upon initialization.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -179,10 +179,10 @@ internal class ReadTracePoint(
 }
 
 internal class WriteTracePoint(
-    private val ownerRepresentation: String?,
+    var ownerRepresentation: String?,
     iThread: Int, actorId: Int,
     callStackTrace: CallStackTrace,
-    private val fieldName: String,
+    var fieldName: String,
```
eupp: Let' try to keep it `private` and `val`, and apply all necessary transformations upon initialization.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -179,10 +179,10 @@ internal class ReadTracePoint(
 }
 
 internal class WriteTracePoint(
-    private val ownerRepresentation: String?,
+    var ownerRepresentation: String?,
```
eupp: Let' try to keep it `private` and `val`, and apply all necessary transformations upon initialization.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -227,7 +227,7 @@ internal class MethodCallTracePoint(
     var thrownException: Throwable? = null
     var parameters: List<String>? = null
     var parameterTypes: List<String>? = null
-    private var ownerName: String? = null
+    var ownerName: String? = null
```
eupp: Let' try to keep it `private` and `val`, and apply all necessary transformations upon initialization.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
```diff
@@ -153,10 +153,10 @@ internal class ObstructionFreedomViolationExecutionAbortTracePoint(
 }
 
 internal class ReadTracePoint(
-    private val ownerRepresentation: String?,
+    var ownerRepresentation: String?,
```
eupp: Let' try to keep it `private` and `val`, and apply all necessary transformations upon initialization.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/MultipleRootsInRunConcurrent.kt
```diff
@@ -0,0 +1,36 @@
...
+package org.jetbrains.kotlinx.lincheck_test.representation
+
+import org.jetbrains.lincheck.Lincheck
+import org.junit.Test
+import kotlin.concurrent.thread
+
+class MultipleRootsInRunConcurrent {
+
+    @Test
+    fun multipleRootsTest() {
```
eupp, line 25: What does this test check? Could you please add a comment?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -49,23 +49,29 @@ abstract class BaseRunConcurrentRepresentationTest<R>(private val outputFileName
...
+            check(error is LincheckAssertionError) {
+                """
             |The test should throw LincheckAssertionError, but instead it failed with:
             |${error.stackTraceToString()}
             """
-            .trimMargin()
-        }
-        if (!isFlakyTest) {
-            error.failure.checkLincheckOutput(outputFileName)
+                    .trimMargin()
```
eupp: Looks like indentation is broken after refactoring.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
```diff
@@ -189,6 +207,121 @@ private fun SingleThreadedTable<TraceNode>.compressThreadStart() = compressNodes
...
+    val before = nestedClassRepresentation.substringBefore('$')
+    val after = nestedClassRepresentation.substringAfter('$', "")
+    if (after.isEmpty()) return nestedClassRepresentation
+    val firstPart = if (before.isNotEmpty() && before[0].isUpperCase() && after[0].isUpperCase()) "$before."
+    else "$before$"
+    return firstPart + fixNestedClassDollar(after)
+}
+
+/**
+ * Removes $0 $1 etc.. These are the result of kotlin compiler dealing with overshadowing variables in lambdas.
```
eupp: BTW, I suspect `$0`, `$1`, etc, might be also related to names of anonymous classes:

https://www.baeldung.com/java-class-name#4-anonymous-classes

Basically, If there are several anonymous classes defined in the same scope, they will get the same name with `$n` being the ordering number of the class inside scope.  If we remove the `$n` suffix in such cases it can lead to confusion.  


Could you please re-check this and add a test if necessary?

eupp (changes requested):
See the comments

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
bbrockbernd: I would assume a compiler introduced method doesn't throw exceptions on its own. Which is still checked by the `check` above. Additionally, as discussed, the if statement is a temporary measure until we know the exact reason why return value types can differ #682

bbrockbernd:
@eupp Thanks for the review. Here I would like to discuss some of the comments in bulk as most of them adress the same issue, namely would we like to optimize owner names during collection or during trace optimization.

I would prefer to optimize owner names during trace compression/optmization for the following reasons:
- All trace polishing logic stays in the same place, and is therefore easier to keep track of.
- All trace polisihing happens during the same phase, and is therefore easier to debug. Put breakpoints before and after optimization and see what changes etc..
- Regarding some `tracePoint` variables becoming non-private and/or mutable: most other variables are already mutable and non-private and mutated during trace optimization. I think it would be more consistent.
- `$fieldName.elemnt` optimization needs owner name and field name so would be non trivial to solve during `lastAccessVariable` resolution.

Let me know what you think!

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
bbrockbernd: I have tried it with the following test. But it seems that `obj1` and `obj2` are caught as owner names. We only apply this transformation on owner names and call arguments which should always resolve the variable name, right? Also, I don't think anonymous classes/objects can have static members.
Please let me know if I am missing a scenario where the `className` of an anonymous class can be part of an owner name.

```kotlin
class MultipleAnonymousObjectsTest: BaseRunConcurrentRepresentationTest<Unit>("run_concurrent_test/multiple_anonymous_objects") {
    var a = 0
    override fun block() {
        
        val obj1 = object : TestInterface {
            override fun inc() {
                a++
            }

        }
        
        val obj2 = object : TestInterface {
            override fun inc() {
                a++
            }

        }
        obj1.inc()
        obj2.inc()
        check(false)
        
    }
    
    interface TestInterface {
        fun inc()
    }
}
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TracePoint.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
bbrockbernd: see https://github.com/JetBrains/lincheck/pull/683#issuecomment-2983300439

eupp:
> Regarding some tracePoint variables becoming non-private and/or mutable: most other variables are already mutable and non-private and mutated during trace optimization. I think it would be more consistent.

Yes, we have some mutable fields in `TracePoint`, but I think we should move in the opposite direction => gradually getting rid of mutable fields and `lateinit` fields. In general, mutable and `lateinit` fields are more error-prone. 

> All trace polishing logic stays in the same place, and is therefore easier to keep track of.

We can still group and isolate this code in some place. My point is that:

- it is better to fully initialize all data in the class constructor;
- if we need to store some data in some "canonical" representation, it is better to covert to this representation in the constructor of the class. 

> $fieldName.element optimization needs owner name and field name so would be non trivial to solve during lastAccessVariable resolution.

You can still apply the tranformation inside the `TracePoint` class, when you initialize `fieldName`/`ownerName`. 

====

> most of them adress the same issue, namely would we like to optimize owner names during collection or during trace optimization.

In general, I would say we need to do all string manipulations even further in the pipeline --- during trace printing. 
But right now, it is not possible without even larger refactoring. 
Thus I would say that for now it is less important --- IMO more important is to clean-up `TracePoint` class to avoid mutable and `lateinit` fields. 

Please also note that in trace recorder we implemented different architecture. There, trace points only store id-s of all data (e.g., `fieldId`, `methodId`, `classId`, etc) instead of strings. In the future, we will likely migrate to similar representation in managed strategy. With this approach, all string manipulations will be done only at trace printing stage.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
eupp: I thought just for uniformity might be a good idea to then convert both `check`-s to `if` statement.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceCompressor.kt
eupp: > But it seems that obj1 and obj2 are caught as owner names.

What if you'll try this code?
```
object : TestInterface {
     override fun inc() {
        a++
    }
}.inc()
        
object : TestInterface {
    override fun inc() {
        a++
    }
}.inc()
```

bbrockbernd: That leads to the trace below.
So it seems you are right; there can indeed be cases where `$N` is used to distinguish anonymous objects. However, in this case it would be trivial to deduct which `inc()` call is which without the number. As they can only happen in one order and are accompanied by line numbers.
```
 block(): threw IllegalStateException at MyTest.block(MyTest.kt:13)                                                                                                            |
|     block$1.inc() at MyTest.block(MyTest.kt:21)                                                                                                                                 |
|       this$0.getA(): 2 at MyTest$block$1.inc(MyTest.kt:19)                                                                                                                      |
|         a ➜ 2 at MyTest.getA(MyTest.kt:14)                                                                                                                                      |
|       this$0.setA(3) at MyTest$block$1.inc(MyTest.kt:19)                                                                                                                        |
|         a = 3 at MyTest.setA(MyTest.kt:14)                                                                                                                                      |
|     block$2.inc() at MyTest.block(MyTest.kt:27)                                                                                                                                 |
|       this$0.getA(): 3 at MyTest$block$2.inc(MyTest.kt:25)                                                                                                                      |
|         a ➜ 3 at MyTest.getA(MyTest.kt:14)                                                                                                                                      |
|       this$0.setA(4) at MyTest$block$2.inc(MyTest.kt:25)                                                                                                                        |
|         a = 4 at MyTest.setA(MyTest.kt:14)     
```

bbrockbernd: But having `block.inc()` would be confusing as well. I guess ideally we would like to somehow show we are dealing with an anonymous class..

bbrockbernd:
> it is better to fully initialize all data in the class constructor

Agreed. I didn't propose to initialize owner and field names later. I propose to polish them later. So initialize it in it's raw form during collection, and optimize during optimization. 

> You can still apply the tranformation inside the TracePoint class, when you initialize fieldName/ownerName.

Okay so then we make polishing transformations, during TracePoint initialization, during last access variable resolution, during compression and during string building. This feels counter intuitive to me.. (last one is sadly unavoidable due to stackTraceElement being immutable, and can be fixed once we have new string storing architecture)

> In general, I would say we need to do all string manipulations even further in the pipeline --- during trace printing.
But right now, it is not possible without even larger refactoring.

Currently, it happens right before we start string building.

> IMO more important is to clean-up TracePoint class to avoid mutable and lateinit fields.

I agree we should get rid of the lateinits. About avoiding mutability: I remember we had a discussion where we hypothesized that part of the bad performance was due to all (deep)copies during optimization, which is a direct result of immutability. 

Lastly, as we have seen at a different comment about `$N` cleaning up the trace can be error prone. These kind of things are IMO easier to debug if all polishing happens during the same phase in the pipeline.



In any case, I would like to merge this soon and with the new architecture coming up it doesn't matter much in the end. So if you decide you'd still like to see the owner name optimizations in `ShadowStackFrame` and `TracePoint` constructor. I'll make the refactor👍

eupp:
> In any case, I would like to merge this soon and with the new architecture coming up it doesn't matter much in the end.

Ok, let's settle on middle-ground then : )

Let's only expose publically `val` properties in `TracePoint`, and make `private set`. 
Let's add some kind of explicit method like `updateOwnerName`/`updateFieldName`, add a TODO comment on why it is there and that we are going to change that, and use it in your transformations during trace compressing stage. 

This way we would not directly expose mutability (`var` field), so hopefully, other people would not start to accidentally rely on mutability of these fields and would not start using it in other places of the codebase.

eupp (approved):

PR #695, opened 2025-06-17
Copy data-structures testing API to the `org.jetbrains.lincheck` package

eupp:


src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -10,32 +10,33 @@
...
-import org.jetbrains.kotlinx.lincheck.verifier.*
-import org.jetbrains.kotlinx.lincheck.strategy.stress.*
 import org.jetbrains.kotlinx.lincheck.util.DEFAULT_LOG_LEVEL
+import org.jetbrains.kotlinx.lincheck.util.LoggingLevel
 import kotlin.reflect.*
 
 /**
  * This class runs concurrent tests.
  */
 class LinChecker(private val testClass: Class<*>, options: Options<*, *>?) {
```
ndkoval: Let's deprecate this constructor as well.

src/jvm/main/org/jetbrains/lincheck/strategy/managed/modelchecking/ModelCheckingStrategyConfiguration.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.jetbrains.kotlinx.lincheck.transformation.InstrumentationMode
+import org.jetbrains.kotlinx.lincheck.transformation.InstrumentationMode.MODEL_CHECKING
+import org.jetbrains.kotlinx.lincheck.verifier.Verifier
+import java.lang.reflect.Method
+import kotlin.collections.ifEmpty
+
+/**
+ * Options for the model checking strategy.
+ */
+class ModelCheckingOptions : ManagedOptions<ModelCheckingOptions, ModelCheckingCTestConfiguration>() {
```
ndkoval: Do we want to move it to another package, e.g., `org.jetbrains.lincheck.datastructures.*`?

src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
```diff
@@ -99,9 +106,44 @@ object Lincheck {
...
+     * Runs the specified concurrent tests.
+     *
+     * If [options] is null, then the `@...CTest` annotations provided on the testing class are used
+     * to specify the test parameters.
+     *
+     * @throws LincheckAssertionError if any of the tests fails.
+     */
+    @JvmStatic
+    @JvmOverloads
+    fun check(testClass: Class<*>, options: Options<*, *>? = null) {
```
ndkoval: Let's remove this function and keep `Options` only

src/jvm/test-integration/org/jetbrains/kotlinx/lincheck_test/guide/BasicCounterTest.kt
```diff
@@ -8,13 +8,20 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+/*
+ * Checks that the old data-structures testing API from `org.jetbrains.kotlinx.lincheck` is still working.
+ */
+
+// Supress deprecation of the old API.
+@file:Suppress("DEPRECATION")
```
ndkoval: Please do not use deprecated API in the guide tests and move this test to another package

src/jvm/test-integration/org/jetbrains/lincheck_test/guide/ObstructionFreedomViolationTest.kt
```diff
@@ -8,11 +8,11 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package org.jetbrains.kotlinx.lincheck_test.guide
```
ndkoval, line 11: For the guide tests, please also update the corresponding links to them in the guide

src/jvm/main/org/jetbrains/kotlinx/lincheck/annotations/Operation.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlin.reflect.KClass
+
+/**
+ * Mark your method with this annotation to use it in concurrent testing as an operation.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY_GETTER, AnnotationTarget.PROPERTY_SETTER)
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "Use org.jetbrains.lincheck.annotations.Operation instead.",
```
ndkoval: What do you think about `replaceWith = ReplaceWith("Operation", "org.jetbrains.lincheck.annotations.Operation")` instead of this message? This will enable a quick-fix in IntelliJ IDEA

src/jvm/main/org/jetbrains/kotlinx/lincheck/annotations/StateRepresentation.kt
```diff
@@ -19,4 +19,8 @@ package org.jetbrains.kotlinx.lincheck.annotations
  */
 @Retention(AnnotationRetention.RUNTIME)
 @Target(AnnotationTarget.FUNCTION)
+@Deprecated(
+    level = DeprecationLevel.WARNING,
+    message = "Use org.jetbrains.lincheck.annotations.StateRepresentation instead.",
```
ndkoval: Should we keep it at all? :)

eupp: I can remove it from the new API.

src/jvm/main/org/jetbrains/kotlinx/lincheck/annotations/Operation.kt
eupp: I've tried using `ReplaceWith`. Unfortunately, it does not work in case when we substitute identifier with the same identifier from different package:
https://youtrack.jetbrains.com/issue/KTIJ-20773/ReplaceWith-doesnt-work-for-classes-with-the-same-name-but-in-different-package

The best we can do is to replace with fully qualified name `org.jetbrains.lincheck.annotations.Operation`, but I think results in a more verbose code.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/quiescent/LockFreeTaskQueueTest.kt
```diff
@@ -11,7 +11,7 @@
 package org.jetbrains.kotlinx.lincheck_test.verifier.quiescent
 
 import org.jetbrains.kotlinx.lincheck.*
-import org.jetbrains.kotlinx.lincheck.annotations.*
+import org.jetbrains.lincheck.annotations.*
```
ndkoval: Would it be more consistent to move the whole API related to data structures to a single package?

src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
```diff
@@ -99,9 +106,28 @@ object Lincheck {
...
+    internal const val DEFAULT_INVOCATIONS = CTestConfiguration.DEFAULT_INVOCATIONS
 }
 
+/**
+ * This is a short-cut for the following code:
+ * ```
+ * Lincheck.check(testClass, options)
+ * ```
+ */
+fun <O : Options<O, *>> O.check(testClass: Class<*>) =
```
ndkoval, line 135: Could you please add a test showing how this function can be called from Java?

src/jvm/main/org/jetbrains/kotlinx/lincheck/annotations/Operation.kt
ndkoval: Got it. Let's then keep the current messages.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/verifier/quiescent/LockFreeTaskQueueTest.kt
eupp: Done

src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
eupp: From Java, it would require to use `LincheckKt` name. 
So I partly reverted this commit: https://github.com/JetBrains/lincheck/pull/695/commits/40254c1b5fd1c681201decc6c688ef64b4033abe and added back `Lincheck.check(options, class)` function that can be conveniently used from Java. 

I've added corresponding Java test.

eupp:
@ndkoval are we ready to merge?

Issue #674, opened 2025-06-04
[Bug] Lincheck cannot find thread to switch to


bbrockbernd:
Test:
```Kotlin
class ReentrantLockMuliLockRepresentationTest: BaseRunConcurrentRepresentationTest<Unit>("reentrant_lock_multi_lock") {
    private val lock = ReentrantLock()

    override fun block() {
        lock.lock()
        val t = thread {
            lock.lock()
            lock.unlock()
        }
        lock.unlock()
        t.join()

    }
    
    override val analyzeStdLib: Boolean = true
}
```

Output:
```
Wow! You've caught a bug in Lincheck.
We kindly ask to provide an issue here: https://github.com/JetBrains/lincheck/issues,
attaching a stack trace printed below and the code that causes the error.

Exception stacktrace:
java.lang.IllegalStateException: Trying to switch the execution to thread 1,
but only the following threads are eligible to switch: []
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$Interleaving.chooseThread(ModelCheckingStrategy.kt:365)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.chooseThread(ModelCheckingStrategy.kt:116)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.chooseThreadSwitch(ManagedStrategy.kt:608)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread(ManagedStrategy.kt:580)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.switchCurrentThread$default(ManagedStrategy.kt:571)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPoint(ManagedStrategy.kt:493)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.newSwitchPoint$default(ManagedStrategy.kt:464)
	at org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy.beforeReadField(ManagedStrategy.kt:1279)
	at sun.nio.ch.lincheck.Injections.beforeReadField(Injections.java:334)
	at java.base/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:698)
[...]

Issue #673, opened 2025-06-04
[Bug] No trace displayed when deadlocked with reentrantlock


bbrockbernd:
No trace displayed when deadlock with reentrantlock

```Kotlin
class ReentrantLockMuliLockRepresentationTest: BaseRunConcurrentRepresentationTest<Unit>("reentrant_lock_multi_lock") {
    private val lock = ReentrantLock()

    override fun block() {
        lock.lock()
        val t = thread {
            lock.lock()
        }
        t.join()

    }
}
```
results in (without any more informtation):
```
= Concurrent test has hung =
```

dmitrii-artuhov:
Hi! Here is the same issue as in https://github.com/JetBrains/lincheck/issues/674. 
When rewriting `lock` as a local variable inside `block` the trace is displayed correctly.

Related to https://github.com/JetBrains/lincheck/issues/504

Closing this issue as a duplicate

PR #696, opened 2025-06-17
Implement lambda capture memory restoring logic

dmitrii-artuhov:
Implements feature required by
- https://github.com/JetBrains/lincheck/issues/673
- https://github.com/JetBrains/lincheck/issues/674


bbrockbernd (commented):
LGTM and nice to have this fix.

I see some representation tests have failed, but this result is expected with the proposed change.

src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
```diff
@@ -79,9 +80,15 @@ object Lincheck {
             val verifier = testCfg.createVerifier()
             val wrapperClass = GeneralPurposeModelCheckingWrapper::class.java
             testCfg.createStrategy(wrapperClass, scenario, null, null).use { strategy ->
+                check(strategy is ModelCheckingStrategy)
+
+                // save fields referenced by lambda for restoring by the snapshot tracker
+                block.javaClass.declaredFields
+                    .mapNotNull { readFieldSafely(block, it).getOrNull() }
+                    .forEach(strategy::updateSnapshotWithRootObject)
```
eupp: Probably worth moving this code to some other place, e.g., to `ManagedStrategy` constructor/initialization, to make sure it is called from all places that run `ManagedStrategy` over the lambda. 

Because, for instance, `runPluginReplay` currently also creates an instance of the `ManagedStrategy`, and there you would not add the fields to the snapshot.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
```diff
@@ -27,15 +27,22 @@ import java.util.concurrent.atomic.AtomicReferenceArray
...
+ *
+ * Also, manual addition of required to restore objects is possible.
  */
 class SnapshotTracker {
     private val trackedObjects = IdentityHashMap<Any, MutableList<MemoryNode>>()
 
+    /**
+     * Stores manually saved root objects, which will be the starting objects during the traversal when restoring snapshot.
+     */
+    private val rootObjects = Collections.newSetFromMap(IdentityHashMap<Any, Boolean>())
```
eupp: Why do you need separate set for `root` objects? Why not just store them indise `trackedObjects`?

src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
dmitrii-artuhov: done

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/SnapshotTracker.kt
dmitrii-artuhov: true, there is not need to start restoring from `root` objects and from `Class<*>` classes

dmitrii-artuhov:
Measured gpmc tests time with and without restoring of lambda's reference to test class:
- WITH — 2m 34s
- WITHOUT — 2m 37s

So there is no difference, probably because there are only like 9 tests which have some fields to restore between invocations among total of 60-ish

PR #700, opened 2025-06-18
Improve storage of TRObjects.

lev-serebryakov-jetbrains:
Store all class names into a cache, use ids.

This change reduces 3.5GiB trace dump into 1.6GiB trace dump.

Before:
Trace collected in 6750 ms
Trace dumped in 9759 ms
Dump size: 3550MiB

After:
Trace collected in 8252 ms
Trace dumped in 6194 ms
Dump size: 1671MiB

Trace collection is slower (as class names are «internalized») but total time is better and file size decreases significantly.


src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
```diff
@@ -270,30 +270,37 @@ fun loadTRTracePoint(inp: DataInput): TRTracePoint {
     return loader(inp, threadId, codeLocationId, eventId)
 }
 
+internal val classNameCache = IndexedPool<String>()
```
eupp, line 273: Note for the future: we likely want to have `ClassDescriptor` class instead, similarly as we have `FieldDescriptor` and `MethodDescriptor`.

lev-serebryakov-jetbrains: And group all "caches" in one file with only them.

PR #702, opened 2025-06-19
Change object numeration implementation

eupp:
- Instead of relying on object graph traversal to enumerate objects, assign object number from `afterNewObjectCreation` injection. 

- Store all information about registered objects in one place inside `ObjectEntry` class. 

- Make object numeration part of the `ObjectTracker` API.

- Improve `ObjectTracker` API. Add possibility of efficient bi-directional querying: `ObjectId -> Object` and `Object -> ObjectId` (and `Object`/`ObjectId` -> `ObjectEntry`). 

- In `ObjectTracker` implementation store objects via `WeakReference` to improve GC and avoid potential memory leaks.

- Do not store object numeration map as global map, instead make it part of `ManagedStrategy` state => less chance of introducing various bugs with memory leaks or objects numeration not being reset between invocations.   
 
 Should resolve: https://github.com/JetBrains/lincheck/issues/593


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
```diff
@@ -61,7 +62,7 @@ internal class ModelCheckingStrategy(
     private var isReplayingSpinCycle = false
 
     // Tracker of objects' allocations and object graph topology.
-    override val objectTracker: ObjectTracker? = if (isInTraceDebuggerMode) null else LocalObjectManager()
```
bbrockbernd, line 64: Does this mean we now track objects when in normal mode? And does that affect performance?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -99,7 +96,8 @@ internal abstract class ManagedStrategy(
     internal val loopDetector: LoopDetector = LoopDetector(settings.hangingDetectionThreshold)
 
     // Tracker of objects' allocations and object graph topology.
-    protected abstract val objectTracker: ObjectTracker?
+    // TODO: make private again
```
bbrockbernd: Looks like this still needs to be done?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -10,8 +10,35 @@
...
+ * The registered objects are associated with the registry entries [ObjectEntry],
+ * keeping object's serial number, its identity hash code, a weak reference to the object,
+ * and, potentially, other meta-data (defined by the concrete implementations of the interface).
+ *
+ * The tracker allows retrieving the registry entry either by the object reference itself or by unique object id.
+ * In this way, the tracker allows to:
+ *   - get an object by its object id: `ObjectId -> Object`;
+ *   - get an object id by the object itself: `Object -> ObjectId`.
  */
 interface ObjectTracker {
```
bbrockbernd, line 43: Is this interface necessary? Idem for the abstract class. I see only one actual implementation. Or are there plans to write more implementations for this interface and even more for the abstract class?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -23,38 +50,448 @@ interface ObjectTracker {
...
+
+/**
+ * Represents an entry for the tracked object.
+ *
+ * @property objectNumber A unique serial number for the object.
+ * @property objectHashCode The identity hash code of the object.
+ * @property objectDisplayNumber The number used in string representation of the object.
+ * @property objectReference A weak reference to the associated object.
+ */
+open class ObjectEntry(
```
bbrockbernd, line 157: Only one subclass. Is this extracted super class necessary?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -23,38 +50,448 @@ interface ObjectTracker {
...
+ * Enumerates all objects currently tracked by the object tracker,
+ * assigning each object its display number as determined by the object tracker.
+ *
+ * Enumeration is required for the Plugin as we want to see on the diagram if some object was replaced by a new one.
+ * Uses the same a numeration map as the trace reporter via [getObjectDisplayNumber] method, so objects have the
+ * same numbers, as they have in the trace.
+ *
+ * @return a map associating each reachable object with its display number.
+ */
+internal fun ObjectTracker.enumerateAllObjects(): Map<Any, Int> {
```
bbrockbernd, line 217: Minor suggestionk: something like the following would seem more idiomatic to me
```kotlin
internal fun ObjectTracker.enumerateAllObjects(): Map<Any, Int> = enumerateObjectEntries()
    .filter { it.objectReference.get() != null }
    .associate { it.objectReference.get()!! to it.objectDisplayNumber }

```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -23,38 +50,448 @@ interface ObjectTracker {
...
+    }
+
+    override fun reset() {
+        objectCounter = 0
+        objectIndex.clear()
+        referenceQueue.clear()
+        perClassObjectNumeration.clear()
+    }
+
+    private fun cleanup() {
```
bbrockbernd, line 488: What is the difference between reset, clear and clean?

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Collections.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
+
+fun <K, V> MutableMap<K, V>.update(key: K, default: V, transform: (V) -> V): V {
+    // TODO: could it be done with a single lookup in a map?
```
bbrockbernd: How about?
```Kotlin
fun <K, V> MutableMap<K, V>.update(key: K, default: V, transform: (V) -> V): V =
    compute(key) { _, current -> transform(current ?: default) }!!

```

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -435,9 +438,11 @@ private fun visualizeTrace(): Array<Any>? = runCatching {
  *
  * The Debugger uses this information to enumerate objects.
  */
-private fun createObjectToNumberMapAsArray(testObject: Any?): Array<Any> {
+@Suppress("UNUSED")
+private fun ObjectTracker.createObjectToNumberMapAsArray(testObject: Any?): Array<Any> {
     val resultArray = arrayListOf<Any>()
-    val numbersMap = if (testObject != null) enumerateObjects(testObject) else enumerateObjects()
+    // val numbersMap = if (testObject != null) enumerateReachableObjects(testObject) else enumerateAllObjects()
```
bbrockbernd, line 444: Commented code

src/jvm/test/resources/expected_logs/trace_debugger/random_bytes_trace_debugger.txt
```diff
@@ -12,22 +12,3 @@ The following interleaving leads to the error:
 | operation() |
 | ----------- |
 
-Detailed trace:
```
bbrockbernd: This trace is gone

bbrockbernd (changes requested):
Hi, I left a few questions about naming and the necessity of inheritance. Also looked like some todos and commented code was forgotten.

Also please note that I am not very familiar with the old object tracking logic

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
eupp: Left it commented, because I wasn't sure about this one. 
It allows to avoid object graph traversal when plugin code calls it (on each stopped trace point) --- this was the issue we discussed in slack. But on the other side, in theory, it can return larger map. 

But if it works on tests from the course, probably we can keep only the `enumerateAllObjects()` version.

bbrockbernd: Hmm, I'll give it another go to be sure. Since I tested with the verison before you rebased and updated. I remember agent going out of memory when clicking a tracepoint.

bbrockbernd: Still works

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
eupp: - `reset()` resets the state of the object tracker between invocations.

- `clear()` is standard method of maps that is called from `reset()` to clear all the internal data structures of the tracker. 

- `cleanup()` is a custom method that is used to remove object entries registered in the object tracker whose objects were garbage collected. 

I've added comments to the `cleanup` methods, hopefully they will help to understand the code better.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/Collections.kt
eupp: applied suggestion

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
eupp: Yes, it is better stylistically, but intermediate `filter` would result in additional intermediate list creation (which can be quite large). I propose to leave the loop code for now.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
eupp: Yes, I agree, it is over-engineering here. Let's keep it as is for now to merge it quicker, and revisit this later.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
eupp: Yes, I agree, it is over-engineering here. Let's keep it as is for now to merge it quicker, and revisit this later.

src/jvm/test/resources/expected_logs/trace_debugger/random_bytes_trace_debugger.txt
eupp: Fixed

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/modelchecking/ModelCheckingStrategy.kt
eupp: Thanks for spotting this, there was an issue here. 

I enabled `LocalObjectManager` only for model checking mode, and for trace debugger mode use base object tracker. 

> Does this mean we now track objects when in normal mode? 

In trace debugger mode, we tracked them via `ObjectIdentityHashCodeTracker` and indirectly through global object numeration map.

PR #703, opened 2025-06-19
Change the artifact groupId and fix the guide tests

ndkoval:
Please note it is OK for the Kotlin Coroutines build to fail with the artifact location change


gradle.properties
```diff
@@ -8,7 +8,7 @@
 # with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
 
-group=org.jetbrains.kotlinx
+group=org.jetbrains.lincheck
```
eupp, line 11: Looks like coroutine tests on CI still search for the old `groupId`.

ndkoval: Yes, but we cannot make the changes atomically. I will merge this PR and find a way to keep the build green.

PR #706, opened 2025-06-19
Group all trace-recorder related classes (but not trace data classes) into separate package.

lev-serebryakov-jetbrains:
It allows to extract it as separate module in the future.


ndkoval:
@lev-serebryakov-jetbrains please merge when the tests pass.

PR #707, opened 2025-06-19
Fix saving and loading of CodeLocations: fix internal/canonical class names

lev-serebryakov-jetbrains:


ivandev0 (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/Serialization.kt
```diff
@@ -242,3 +245,17 @@ private fun DataInput.readInternalizedString(internalizer: (Int) -> String?): St
...
+ */
+private fun String.toCanonicalClassName() =
+    this.replace('/', '.')
+
+/**
+ * Converts a string representing a class name in canonical format (e.g., "com.example.MyClass")
+ * into an internal class name format with (e.g., "com/example/MyClass").
+ */
+private fun String.toInternalClassName() =
+    this.replace('.', '/')
```
eupp, line 261: Why need to copy these functions?
They should already be available somewhere in Lincheck codebase (either `utils` or `transformation` sub-package).

lev-serebryakov-jetbrains: Because this package should be dependence-free to make it easy to extract to separate module.

eupp: Ok, I see. 

We would need to revisit this problem later. I think we need some sort of `core` package, that should include `utils` and perhaps some other classes, and should be distributed along `tracedata` jar (as well as other potential separate jars).

PR #711, opened 2025-06-19
Add trace recorder integration tests

dmitrii-artuhov:
This PR generalizes API for writing integrations tests for trace agent (trace debugger, trace recorder).

Left TODOs:
- [x] Add a config on teamcity to run trace recorder tests
- [x] Change packages of the integration tests so they do not have 'kotlinx' anymore
- [x] Fix trace recorder trace, so that lamdas' addresses are not printed + enable constant hashcodes as in trace debugger via jvm arg
- [ ] Select which tests to run with trace recorder (right now only all methods of `tests.contract.list.ImmutableListTest` from kotlinx-collections-immutable project)
- [ ] Change the integration tests folder structure to use `gradle subprojects` instead of `gradle source sets` (*could be done in separate PR, as it might be a more difficult task*)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -282,7 +282,6 @@ tasks {
             extraArgs.add("-Dlincheck.traceDebuggerMode=true")
             exclude("**/lincheck_test/guide/*")
         }
-        // TODO: do we need testInTraceRecorderMode in the gradle.properties then?
         val dumpTransformedSources: String by project
         if (dumpTransformedSources.toBoolean()) {
             extraArgs.add("-Dlincheck.dumpTransformedSources=true")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -380,7 +380,7 @@ tasks {
         classpath = sourceSets["traceRecorderTest"].runtimeClasspath
 
         outputs.upToDateWhen { false } // Always run tests when called
-        dependsOn(traceAgentIntegrationTestsPrerequisites) // TODO: do we need it here?
+        dependsOn(traceAgentIntegrationTestsPrerequisites)
     }
 
     check {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -78,6 +78,15 @@ sourceSets {
         }
     }
 
+    create("traceRecorderTest") {
+        java.srcDir("src/jvm/test-trace-recorder-integration")
+        configureClasspath()
+
+        resources {
+            srcDir("src/jvm/test-trace-recorder-integration/resources")
+        }
+    }
+
     dependencies {
         // main
         val kotlinVersion: String by project
@@ -101,10 +110,13 @@ sourceSets {
         val junitVersion: String by project
         val jctoolsVersion: String by project
         val mockkVersion: String by project
+        val slf4jVersion: String by project
+        val gradleToolingApiVersion: String by project
 
         testImplementation("junit:junit:$junitVersion")
         testImplementation("org.jctools:jctools-core:$jctoolsVersion")
         testImplementation("io.mockk:mockk:${mockkVersion}")
+        testImplementation("org.gradle:gradle-tooling-api:${gradleToolingApiVersion}")
 
         // integrationTest
         val integrationTestImplementation by configurations
@@ -114,13 +126,20 @@ sourceSets {
         integrationTestImplementation("org.jctools:jctools-core:$jctoolsVersion")
 
         // traceDebuggerTest
-        val gradleToolingApiVersion: String by project
         val traceDebuggerTestImplementation by configurations
         val traceDebuggerTestRuntimeOnly by configurations
 
         traceDebuggerTestImplementation("junit:junit:$junitVersion")
         traceDebuggerTestImplementation("org.gradle:gradle-tooling-api:${gradleToolingApiVersion}")
-        traceDebuggerTestRuntimeOnly("org.slf4j:slf4j-simple:1.7.10")
+        traceDebuggerTestRuntimeOnly("org.slf4j:slf4j-simple:$slf4jVersion")
+
+        // traceRecorderTest
+        val traceRecorderTestImplementation by configurations
+        val traceRecorderTestRuntimeOnly by configurations
+
+        traceRecorderTestImplementation("junit:junit:$junitVersion")
+        traceRecorderTestImplementation("org.gradle:gradle-tooling-api:${gradleToolingApiVersion}")
+        traceRecorderTestRuntimeOnly("org.slf4j:slf4j-simple:$slf4jVersion")
     }
 }
 
@@ -156,6 +175,13 @@ tasks {
     named<KotlinCompile>("compileTraceDebuggerTestKotlin") {
         setupKotlinToolchain()
     }
+
+    named<JavaCompile>("compileTraceRecorderTestJava") {
+        setupJavaToolchain()
+    }
+    named<KotlinCompile>("compileTraceRecorderTestKotlin") {
+        setupKotlinToolchain()
+    }
 }
 
 fun JavaCompile.setupJavaToolchain() {
@@ -181,6 +207,9 @@ kotlin {
         configureAssociation()
     }
     target.compilations.named("traceDebuggerTest") {
+        configureAssociation()
+    }
+    target.compilations.named("traceRecorderTest") {
         configureAssociation()
     }
 }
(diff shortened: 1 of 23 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
```diff
@@ -84,11 +84,9 @@ class TRMethodCallTracePoint(
         }
         sb.append(')')
         if (exceptionClassName != null) {
-            sb.append(": THROWS EXCEPTION")
-                .append(exceptionClassName)
+            sb.append(": threw $exceptionClassName")
         } else if (result != TR_OBJECT_VOID) {
-            sb.append(": ")
-            sb.append(result.toShortString())
+            sb.append(": ${result.toShortString()}")
```
lev-serebryakov-jetbrains: Why?! Template in StringBuilder argument looks very strange. StringBuilder is used to avoid templating/concatenation cost for sure.

 Yo Dawg, Heard You concatenate strings so we put template into StringBuilder so you can concatenate strings when you concatenate strings!

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/AbstractIntegrationTest.kt
```diff
@@ -0,0 +1,314 @@
...
+import java.io.File
+import java.io.FileInputStream
+import java.io.FileOutputStream
+import java.lang.reflect.Method
+import java.lang.reflect.Modifier
+import java.net.URL
+import java.net.URLClassLoader
+import java.nio.file.Paths
+
+abstract class AbstractIntegrationTest {
```
eupp: `AbstractTraceIntegrationTest`

build.gradle.kts
```diff
@@ -253,6 +282,7 @@ tasks {
             extraArgs.add("-Dlincheck.traceDebuggerMode=true")
             exclude("**/lincheck_test/guide/*")
         }
+        // TODO: do we need testInTraceRecorderMode in the gradle.properties then?
```
eupp: Don't think we need it. The flag was mostly used to selectively disable some Lincheck tests that are not yet supported in trace debugger mode. Since trace recorder does not re-use `ManagedStrategy`, there should be no overlap between the tests.

src/jvm/test-trace-recorder-integration/org/jetbrains/trace_recorder/integration/TraceDebuggerExamplesTraceRecorderIntegrationTest.kt
```diff
@@ -0,0 +1,37 @@
...
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.trace_recorder.integration
+
+import org.junit.Test
+import java.nio.file.Paths
+
+class TraceDebuggerExamplesTraceRecorderIntegrationTest : AbstractTraceRecorderIntegrationTest() {
```
eupp, line 16: Name `TraceDebuggerExamplesTraceRecorderIntegrationTest` looks very confusing.

eupp (commented):
Left few minor comments

eupp:
> Select which tests to run with trace recorder (right now only all methods of tests.contract.list.ImmutableListTest from kotlinx-collections-immutable project)

In this PR it is not important. For now it is more important to set-up the infrastructure, we can add/configure tests later. 

>  Fix trace recorder trace, so that lamdas' addresses are not printed + enable constant hashcodes as in trace debugger via jvm arg

Could you fix it before we merge this PR?

> Change the integration tests folder structure to use gradle subprojects instead of gradle source sets

Separate PR.

src/jvm/test-trace-recorder-integration/org/jetbrains/trace_recorder/integration/TraceDebuggerExamplesTraceRecorderIntegrationTest.kt
dmitrii-artuhov: this is a pattern of naming such integration tests -- starting with the downloaded project name

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
```diff
@@ -450,9 +447,15 @@ data class TRObject internal constructor (
         } else if (classNameId == TR_OBJECT_VOID_CLASSNAME) {
             "void"
         } else {
-            className.substringAfterLast(".") + "@" + identityHashCode
+            adornedRepresentation()
```
eupp, line 450: Please partly revert to:
```
adornedClassNameRepresentation(className) + "@" + identityHashCode
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
```diff
@@ -450,9 +447,15 @@ data class TRObject internal constructor (
...
         }
     }
+
+    fun adornedRepresentation(): String =
+        className.substringAfterLast(".").let {
+            if (isJavaLambdaClass(it)) it.substringBeforeLast('/')
+            else it
+        } + "@" + identityHashCode
```
eupp: @lev-serebryakov-jetbrains WDYT, are we free to perform string manipulations like this one during trace printing without significan perf. overhead?

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
dmitrii-artuhov: `adornedRepresentation` is used in 3 different places, in each of them I would need to write code like this. I can extract a separate `adornedClassNameRepresentation` method, I don't see why we would want to rewrite it to explicit concatenation everywhere

eupp: >  I can extract a separate `adornedClassNameRepresentation` method

Yes, please.

lev-serebryakov-jetbrains: We've discussed this with Nikita a lot. I don't like it at all: adornedStringRepresentation depends on other methods, and I don't like that this code will depend on trace data structures package AT ALL.

I've specifically added comment with TODO to `TRObject`, that when we will have `core` module we'll move all such utility code & re-use it everywhere. Now I prefer to left this packag as dependencyt-free as possible and to have as little code as possible to depend on it for UTILITIES not realted to its main functions.

eupp: As far as I understand, the only dependency is `isJavaLambdaClass` method. 
Let's copy it here for now, and revisit this code later once we will have `core` module.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: I think, pretty-prinitng is not performance-critical code at all.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: I've discussed about `adornedObjectRepresentation()` which is much more complex, but still.

eupp: `adornedObjectRepresentation` is not used here now. 
There are just two new methods `adornedRepresentation` and `adornedClassNameRepresentation` defined in this file below : )

dmitrii-artuhov: is it critical to copy `isJavaLambdaClass` method here?
@lev-serebryakov-jetbrains

dmitrii-artuhov: - Will leave as is and create a separate YouTrack task

dmitrii-artuhov:
@eupp please approve PR, so I can merge it

PR #716, opened 2025-06-20
Add primitive types.

lev-serebryakov-jetbrains:
Performance changes are not statistically-significant.

Now text output contains real values of all primitives.


src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/TraceRecorderTracePoints.kt
```diff
@@ -412,21 +414,53 @@ fun loadTRTracePoint(inp: DataInput): TRTracePoint {
...
 ) {
-    constructor(className: String, identityHashCode: Int):
-            this(classNameCache.getOrCreateId(className), identityHashCode)
-
-    val className get() = classNameCache[classNameId]
+    val className: String  get() = primitiveValue?.javaClass?.name ?: classNameCache[classNameId]
+    val isPrimitive: Boolean get() = primitiveValue != null
+    val value: Any? get() = primitiveValue
+
+    override fun toString(): String {
```
ndkoval, line 431: Let's reuse `ObjectLabelFactory.adornedStringRepresentation`, moving the function to another place and making it possible to not use the object enumeration logic.

lev-serebryakov-jetbrains: What do you mean by "moving the function to another place"?

lev-serebryakov-jetbrains: Also, it is bad idea to allow multiline strings! And `adornedStringRepresentation` doesn't bother. Should we teach it?

lev-serebryakov-jetbrains: Decided to add only TODO for future (private conversation)

zhelenskiy (approved):

PR #731, opened 2025-06-25
Refactor lincheck integration tests

dmitrii-artuhov:
Changes:
- Moved conceptually lincheck integration tests from `test` source set to `test-lincheck-integration`
- Updated lincheck integration tests package name not to contain `kotlinx` anymore

TODOs:
- [ ] [Another PR] Renamed lincheck integration tests to match trace-debugger/trace-recorder naming pattern: gradle task name change `integrationTest` -> `lincheckIntegrationTest` + modify team-city tasks for running lincheck integration tests (because gradle task name changed)


src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/linearizability/BufferedChannelTest.kt
```diff
@@ -1,18 +1,19 @@
...
  *
- * Copyright (C) 2019 - 2023 JetBrains s.r.o.
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
  *
  * This Source Code Form is subject to the terms of the
  * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
-package org.jetbrains.kotlinx.lincheck_test.verifier.linearizability
+package org.jetbrains.lincheck_test.verifier.linearizability
```
eupp: Let's move to `org.jetbrains.lincheck_test.datastructures.linearizable`

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/quiescent/LockFreeTaskQueueTest.kt
```diff
@@ -1,24 +1,23 @@
...
- * Copyright (C) 2019 - 2023 JetBrains s.r.o.
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
  *
  * This Source Code Form is subject to the terms of the
  * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package org.jetbrains.kotlinx.lincheck_test.verifier.quiescent
+package org.jetbrains.lincheck_test.verifier.quiescent
```
eupp: Let's move to `org.jetbrains.lincheck_test.datastructures.quiescent`

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/FaaQueueTest.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.lincheck_test
```
eupp: Let's move to `org.jetbrains.lincheck_test.datastructures`

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/serializability/SerializableQueueTest.kt
```diff
@@ -8,7 +8,7 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package org.jetbrains.kotlinx.lincheck_test.verifier.serializability
+package org.jetbrains.lincheck_test.verifier.serializability
```
eupp: Let's move to `org.jetbrains.lincheck_test.datastructures.serializable`

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/PrintingTest.kt
```diff
@@ -19,7 +19,7 @@ import kotlin.concurrent.thread
  * This test aims to test the effects of printing to `stdout` and `stderr`.
  * The test used to hang because of prints, see https://github.com/JetBrains/lincheck/issues/409.
  *
- * See [CyclicBarrierTest] for identical test but with no printing.
+ * See `CyclicBarrierTest` for identical test but with no printing.
  */
 class PrintingTest {
```
eupp, line 24: If you moved the `CyclicBarrierTest` to integration tests, would it make sense to also move this one? 


Or, in general, are there any other gpmc tests worth moving into integration tests?
I am trying to understand our long-term strategy here: what tests should we move to integration tests? The moving itself we can perform later in separate PR.

eupp (changes requested):
Let's fix few minor comments, then it is good to go.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/PrintingTest.kt
dmitrii-artuhov: In general, most of the gpmc tests look like they are integrational, since they all use at some point end-to-end functionality of gpmc API (the 'runConcurrentTest' function). But I think, that gpmc tests which aim to replicate some bug that we had (like IntrinsicCandidateMethodTest/PrintingTest) or to test specific use cases (like in InterruptionTest/HangTest/DaemonThreadPoolTest) should not be treated as integrational.

While the tests that just use some existing data structure and just test for finding/not finding the assertion failure - those should be treated as integrational (like CountDownLatchTest/CyclicBarrierTest). Actually, I think tests from DataStructuresTests.kt are also integrational, but they use a special method for executing the test -- 'runModelChecking' instead of the API function, so I left them in the unit tests.

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/linearizability/BufferedChannelTest.kt
dmitrii-artuhov: I thought 'datastructures' package was for the implementation of data-structures-under-test and not for the tests themselves. Why should we change that approach?

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/FaaQueueTest.kt
dmitrii-artuhov: As I have commented above, the implementation of data-structure-under-test (FAAQueue.kt) is in the 'datastructures' package, but the test is not

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/linearizability/BufferedChannelTest.kt
eupp: The idea is to eventually in Lincheck implementation move everything related to data structures testing to `org.jetbrains.lincheck.datastructures` package. Then it also makes sense to move all data structures related tests to corresponding test package.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/gpmc/PrintingTest.kt
eupp: > but they use special method for executing the test -- 'runModelChecking' instead of the API function, so I left them in the unit tests.

We should migrate them to normal GPMC API at some point later.

src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/verifier/linearizability/BufferedChannelTest.kt
eupp: Also `verifier.linearizability` package does not make much sense in this context, since it is not only the verifier is tested here, but rather the whole testing framework.

PR #733, opened 2025-06-26
Do not check representation test outputs on non-default JDKs by default

dmitrii-artuhov:
This PR makes sure that only default jdk (17) outputs are stored, compare when tests are run and overwritten when required. Each representation test, though, has the special flag to enable output comparison for any provided jdk


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -46,9 +46,13 @@ internal inline fun <reified E: Exception> Options<*, *>.checkFailsWithException
  * Checks that failure output matches the expected one stored in a file.
  *
  * @param expectedOutputFilePrefix name of file stored in resources/expected_logs, storing the expected lincheck output.
+ * @param handleOnlyDefaultJdkOutput flag which disables comparing and overwriting representation output files for non-default jdks.
  */
-internal fun LincheckFailure?.checkLincheckOutput(expectedOutputFilePrefix: String) {
+internal fun LincheckFailure?.checkLincheckOutput(expectedOutputFilePrefix: String, handleOnlyDefaultJdkOutput: Boolean = true) {
```
eupp, line 51: Do we want to use system property instead?

Also, if we assume `handleOnlyDefaultJdkOutput = true` is a new default, should we remove all non-default expected trace log files?
Otherwise, if we do not re-check and update them on each PR, they will become obsolete pretty soon.

dmitrii-artuhov: Motivation for the per-thread setup of this flag was to eventually leave only a couple of non-default jdk representation tests and the rest would be just jdk-17. Thus, global flag is not needed.

I agree that we can remove non-default JDK representation outputs. I'll do it in this PR

eupp: Ok

eupp (approved):

eupp:
@dmitrii-artuhov Please don't forget to also update the `overwrite_representation_tests.sh` script.

ndkoval:
@dmitrii-artuhov could you please also remove the script for generating the outputs and revise the PR title and description?

dmitrii-artuhov:
Hi!

@ndkoval I don't see the reason to remove the script, it is still useful for overriding default jdk output for representation tests with and without trace-debugger flag

@eupp I have updated the description of the script and removed all jdks from list except for the "17"

ndkoval:
@dmitrii-artuhov and @eupp, do we want to keep trace debugger representation tests? 

With the Trace Recorder project, we showed that writing a custom `EventTracker` is a significantly more promising approach, which means Trace Debugger will very unlikely share the Lincheck infrastructure in the future. In this case, we do not need these representation tests, and can safely remove them, remove the script, and even remove the TeamCity build.

eupp:
> do we want to keep trace debugger representation tests?

I don't know, let's discuss on the next sync. In the meantime, I propose to merge this PR and not wait for the decision on trace debugger tests. We can always disable them too, it is not hard/long to do so in a separate PR.

ndkoval (approved):

PR #738, opened 2025-06-30
New on-disk trace format: Supports streaming write and lazy loading.

lev-serebryakov-jetbrains:
It doesn't support streaming write from multiple threads yet.

Documentation will be added later today or tomorrow.


src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/Serialization.kt
```diff
@@ -12,261 +12,343 @@ package org.jetbrains.kotlinx.lincheck.tracedata
...
+
+    /**
+     * Must be called when the method call is ended and the call stack is popped.
+     *
+     * @param callTracepoint tracepoint popped from the call stack.
+     */
+    fun callEnded(callTracepoint: TRMethodCallTracePoint)
+
+    /**
+     * Must be called when trace is finished
```
ndkoval: What should happen if the program gets aborted while collecting the trace?

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/SerDesCommon.kt
```diff
@@ -0,0 +1,180 @@
+/*
```
ndkoval: Please consider renaming the file to `SerializationCommon`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/Serialization.kt
lev-serebryakov-jetbrains: File (and especially index!) will be unfinished. Now there is no recovery for this condition, but I'll add ability to read such files together with multithread support.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracedata/SerDesCommon.kt
lev-serebryakov-jetbrains: But it is not only Serialization! :-)

ndkoval: Serialization implies deserialization.

lev-serebryakov-jetbrains: Done.

trace/README.md
```diff
@@ -0,0 +1,380 @@
...
+
+ These objects are called «reference objects» or «thread-independent objects»
+in this document, and these names are used interchangeably.
+
+ Binary storage format follows this principle and makes the best effort to store 
+all such entities only once, no matter how many other entities reference them.
+
+ To allow on-the-fly writing («streaming») of trace, a binary format doesn't group
+all reference data in one place like the header or footer of the file but intermixes
+it with trace points as needed.
```
eupp, line 29: I wonder why this approach was chosen instead of storing the reference data in header/footer or in a separate file altogether. Even with streaming approach, the reference data can be saved all-at-once at the end of tracing (e.g., in some `finally` block), either in the footer of the file or in some separate file.  

It would solve problems with races mentioned in this document later. And IMO if "reference" data and trace points are stored separately, it makes the whole format  cleaner.

trace/README.md
```diff
@@ -0,0 +1,380 @@
...
+  the data file. These offsets point to places from which corresponding object, starting with
+  object kind, can be read without additional calculations. API checks consistency between
+  index and data file and reports an error when the kind or id of the object in the data file 
+  is not equal to the corresponding kind and/or id in the index cell. In such cases the index
+  is thought to be broken and discarded. **End offset** is not used for such objects and ignored.
+
+- For containers (which are all thread-dependent now) the **start offset** points right
+  after a fixed header object of the container (to the first byte of its variable-size content).
+  **End offset** points to last byte of content, right before `TRACEPOINT_FOOTER` on-disk object.
+  Both of these offsets are *logical*, as opposed to *physical* offsets of other cells. It means
```
eupp, line 224: It's not clear from the document how the logical positions can be converted to physical. 

Is there any index that stores all thread blocks physical offsets? 
Otherwise, what is an intended way to collect all thread blocks into a single stream of thread's trace points?

trace/src/main/org/jetbrains/lincheck/trace/AtomicBitmap.kt
```diff
@@ -0,0 +1,66 @@
...
+        do {
+            val b = bitmap.get()
+            // Another thread was faster
+            if (b.length() > idx) return
+
+            val newLen = max(idx + 1, b.length() + b.length() / 2)
+            val newB = AtomicLongArray(newLen)
+            for (i in 0..< b.length()) {
+                newB.lazySet(i,b.get(i))
+            }
```
eupp: Not sure that the resizing operation is linearizable w.r.t. concurrent `set` operations. 
The problem, I think, is that  copying the current bitmap in the loop could potentially lost some concurrent updates.  

For instance, imagine you copying the bitmap of 4 longs, and first 2 longs were already copied. 
At the same time, concurrent `set` changes the 0-bit of the first long. This update will be lost after successfull CAS of the `bitmap`.

trace/src/main/org/jetbrains/lincheck/trace/Printing.kt
```diff
@@ -29,8 +38,26 @@ private fun printTRPoint(output: PrintStream, node: TRTracePoint, depth: Int, ve
     output.print(" ".repeat(depth * 2))
     output.println(node.toText(verbose))
     if (node is TRMethodCallTracePoint) {
-        node.events.forEach {
-            printTRPoint(output, it, depth + 1, verbose)
+        var unloaded = 0
+        node.events.forEach { event ->
+            if (event == null) {
```
eupp, line 43: IDEA reports `event == null` is always false.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+import kotlin.io.path.Path
+
+private const val INPUT_BUFFER_SIZE: Int = 16 * 1024 * 1024
+
+private typealias CallStack = MutableList<TRMethodCallTracePoint>
+private typealias TracePointReader = (DataInput, TraceContext, CodeLocationsContext) -> Boolean
+
+private interface TracepointConsumer {
+    fun tracePointRead(parent: TRMethodCallTracePoint?, tracePoint: TRTracePoint)
+    fun footerStarted(tracePoint: TRMethodCallTracePoint) = Unit
```
eupp: IMO, simply:

```kotlin
fun footerStarted(tracePoint: TRMethodCallTracePoint) {}
```

is clearer

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+    init {
+        require(start >= 0) { "start must be non-negative" }
+        require(start < end)  { "block cannot be empty" }
+        require(aaccDataSize >= 0) { "accumulated data size cannot be negative" }
+        physicalStart = start
+        physicalEnd = end
+        accDataSize = aaccDataSize
+    }
+
+    fun coversPhysicalOffset(offset: Long): Boolean = offset in physicalStart ..<physicalEnd
```
```suggestion
    fun coversPhysicalOffset(offset: Long): Boolean = offset in physicalStart ..< physicalEnd
```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+
+    fun coversLogicalOffset(offset: Long): Boolean = offset in logicalDataStart ..<logicalDataStart + dataSize
```
```suggestion
    fun coversLogicalOffset(offset: Long): Boolean = offset in logicalDataStart ..< logicalDataStart + dataSize
```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+
+private interface BlockConsumer {
+    fun blockStarted(threadId: Int) = Unit
+    fun blockEnded(threadId: Int) = Unit
+}
+
+private class DataBlock(
+    start: Long,
+    end: Long,
+    aaccDataSize: Long
```
eupp: What does `aacc` stands for?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+        }
+    }
+}
+
+private fun loadTracePointDeep(
+    input: DataInput,
+    context: TraceContext,
+    codeLocs: CodeLocationsContext,
+    stack: CallStack,
+    consumer: TracepointConsumer
```
eupp, line 658: BTW, why not using Kotlin's `Stream`-s instead of consumer interfaces?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+        require(count in 1 .. parent.events.size - from) { "Count $count must be in range 1..${parent.events.size - from}" }
+
+        data.seek(calculatePhysicalOffset(parent.threadId, parent.getChildAddress(from)))
+        loadTracePoints(parent.threadId, count,this::readTracePointWithChildAddresses) { idx, tracePoint, _ ->
+            parent.loadChild(idx + from, tracePoint)
+        }
+
+    }
+
+    private fun loadTracePoints(threadId: Int, maxRead: Int, reader: () -> TRTracePoint, consumer: (Int, TRTracePoint, Long) -> Unit) {
```
eupp: The type of `consumer: (Int, TRTracePoint, Long) -> Unit` is not clear from the context. 
What are `Int` and `Long` components denote?

Maybe worth introducing class with proper field names instead of triple `(Int, TRTracePoint, Long)`?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+            parent.loadChild(idx, tracePoint)
+        }
+
+        val actualFooterPos = data.position() - 1 // 1 is size of object kind
+        check(actualFooterPos == calculatePhysicalOffset(parent.threadId, positions.second)) {
+            "Input contains broken data: expected Tracepoint Footer for event ${parent.eventId} at position ${positions.second}, got $actualFooterPos"
+        }
+    }
+
+    fun loadChild(parent: TRMethodCallTracePoint, childIdx: Int) = loadChildrenRange(parent, childIdx, 1)
```
eupp: For `Unit`-return methods, it's better to wrap the body into `{ ... }` instead of using `= ...`. 
Or, at least, add a return type `: Unit` annotation explicitly.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TRPlayground.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.jetbrains.kotlinx.lincheck.tracedata.LazyTraceReader
+import org.jetbrains.lincheck.trace.TRACE_CONTEXT
+import org.jetbrains.lincheck.trace.TRMethodCallTracePoint
+import org.jetbrains.lincheck.trace.printRecorderTrace
+
+fun main(args: Array<String>) {
+    // val (context, trace) = loadRecordedTrace(FileInputStream("C:\\home\\lev\\Projects\\kotlinx.collections.immutable\\core\\output.bin"))
+
+    val context = TRACE_CONTEXT
+    val reader = LazyTraceReader("C:\\home\\lev\\dada\\FundsTest_testTransfers1056732866102702014.bin")
```
eupp: Better replace the actual path with some placeholder.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TRPlayground.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package org.jetbrains.kotlinx.lincheck_test.util
+
+import org.jetbrains.kotlinx.lincheck.tracedata.LazyTraceReader
+import org.jetbrains.lincheck.trace.TRACE_CONTEXT
+import org.jetbrains.lincheck.trace.TRMethodCallTracePoint
+import org.jetbrains.lincheck.trace.printRecorderTrace
+
+fun main(args: Array<String>) {
+    // val (context, trace) = loadRecordedTrace(FileInputStream("C:\\home\\lev\\Projects\\kotlinx.collections.immutable\\core\\output.bin"))
```
eupp: Should the commented code be removed?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -0,0 +1,771 @@
...
+
+    private companion object {
+        fun wrapStream(input: InputStream?): DataInputStream? {
+            if (input == null) return null
+            return DataInputStream(input.buffered(INPUT_BUFFER_SIZE))
+        }
+    }
+}
+
+fun loadRecordedTrace(inp: InputStream): Pair<TraceContext, List<TRTracePoint>> {
```
eupp: Worth adding a data class with proper field names instead of  `Pair<TraceContext, List<TRTracePoint>>`

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
```diff
@@ -75,7 +74,7 @@ private class ThreadData(
 }
 
 enum class TraceCollectorOutputType {
-    BINARY, TEXT, VERBOSE
+    BINARY, `BINARY-MEM`, TEXT, VERBOSE
```
eupp: What does `BINARY-MEM` stands for?

Why not simply `BINARY_MEM`?

trace/README.md
eupp: Discussed: we want (?) to support "hard" crash scenario, and thus need to support on-the-fly backup of reference data.

trace/README.md
lev-serebryakov-jetbrains: Yep, all data blocks are published to index, it is described in document («For *data blocks*, the **start offset** points to the beginning of corresponding `BLOCK_START`...»).

trace/src/main/org/jetbrains/lincheck/trace/AtomicBitmap.kt
lev-serebryakov-jetbrains: You are right, it is possible. I'll fix that

trace/README.md
lev-serebryakov-jetbrains: To support hard crashes.

trace/src/main/org/jetbrains/lincheck/trace/Printing.kt
lev-serebryakov-jetbrains: It is very strange. It doesn't for me... And it could be null, as `evenyts` property has type `List<TRTracePoint?>`.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TRPlayground.kt
lev-serebryakov-jetbrains: Of course! My bad.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TRPlayground.kt
lev-serebryakov-jetbrains: It is example of two APIs, to switch between them quickly. It is playground for quick debugging, I'm not sure it should be code-style-proof.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: Ok!

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
lev-serebryakov-jetbrains: I need to add documentation, yes. `BINARY-MEM` is strategy to collect-and-dump, as opposite to `BINARY` which is «streaming» API. I don't like `_` in configs, so I choose to use `-`. 
I'm open to better name for this mode, I don't like `binary-mem` much, to be honest.

trace/src/main/org/jetbrains/lincheck/trace/AtomicBitmap.kt
```diff
@@ -0,0 +1,79 @@
...
+        if (chunk >= chunkCount) {
+            resizeBitmap(chunk)
+        }
+
+        val ch = bitmap.get(chunk)
+        do {
+            val value = ch.get(idx)
+            val newValue = value or bit
+        } while (ch.compareAndSet(idx, value, newValue))
+    }
```
eupp, line 69: Missed empty line.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
```diff
@@ -74,23 +73,55 @@ private class ThreadData(
...
+     * without code locations.
+     */
+    TEXT,
+
+    /**
+     * Collect full trace in memory and print it as text to the output file,
+     * with code locations.
+     */
+    VERBOSE
 }
```
eupp, line 110: What about the following naming?:
```kotlin
enum class TraceCollectorMode { 
    BINARY_STREAMING,
    BINARY_DUMP,
    TEXT,
    TEXT_VERBOSE,
}
```

`_` can be changed to `-`

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
eupp: > BINARY-MEM is strategy to collect-and-dump, as opposite to BINARY which is «streaming» API

Left a comment below with a proposal.

>  I don't like _ in configs, so I choose to use -.

By configs, you mean the way it is passed from CLI, as `binary-mem` ?
If so, then I think the enum entry name is not necessarily should correspond to the way it is passed from CLI or other configs. For instance, you can just modify the `toTraceCollectorOutputType()` method implementation, or override `TraceCollectorMode.toString()`, etc. 

IMO, for enum class name it is better to use underscore `BINARY_MEM`, because otherwise we need to wrap the name into quotation everywhere, which is less convenient.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: aAccumulatedDataSize

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: TracepointConsumer is more complicated than 1 method

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: Create class instances to pass 3 arguments? I don't like this idea at all... I can create proper interface with one method to name parameters.

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
lev-serebryakov-jetbrains: I don't like that `binary_streaming` is very long...

src/jvm/main/org/jetbrains/kotlinx/lincheck/tracerecorder/TraceCollectingEventTracker.kt
```diff
@@ -74,23 +73,55 @@ private class ThreadData(
...
+     * It is the default mode if a parameter is not passed or cannot be
+     * recognized.
+     */
+    BINARY,
+
+    /**
+     * Collect full trace in the memory and dump to the output file at the end
+     * of the run.
+     */
+    `BINARY-MEM`,
```
harrytmthy: Q: Is there a reason for naming this enum entry \`BINARY-MEM\` using backticks?

Using a hyphen requires quoting and makes usage awkward:

```kotlin
TraceCollectorMode.`BINARY-MEM`
```

Why not renaming it to `BINARY_MEM` or `BUFFERED_BINARY` for consistency with enum naming conventions and easier usage in code?

lev-serebryakov-jetbrains: It is refactored already.

PR #739, opened 2025-06-30
Integration tests modularization

dmitrii-artuhov:
This PR extracts integration tests of `lincheck`, `trace-debugger`, and `trace-recorder` into a separate sub-project. Meaning the following structure:

```txt
:
\
  :integration-test
    :common (common code for the rest of the subprojects)
    :lincheck
    :trace-debugger
    :trace-recorder

: - (sub)project
```


integration-test/build.gradle.kts
```diff
@@ -0,0 +1,204 @@
...
+        lincheckIntegrationTestImplementation("junit:junit:$junitVersion")
+        lincheckIntegrationTestImplementation("org.jctools:jctools-core:$jctoolsVersion")
+
+        // traceDebuggerIntegrationTest
+        val traceDebuggerIntegrationTestImplementation by configurations
+        val traceDebuggerIntegrationTestRuntimeOnly by configurations
+
+        traceDebuggerIntegrationTestImplementation("junit:junit:$junitVersion")
+        traceDebuggerIntegrationTestImplementation("io.mockk:mockk:${mockkVersion}") // transitively includes junit 5, which is needed to
+                                                                                     // parse junit-5 '@Test' annotations from method definitions
```
dmitrii-artuhov: There are better solutions:
- explicitly depend on different junit version that we plan to support (`4.x`, `5.x`, any more?) in the integration-test subproject
- extract runtime classpath from the test that we compile (already done in the 2nd init-script), and access the `method.annotations` property, so that classloading is done via a custom URI classloader which uses that runtime classpath

ndkoval:
@dmitrii-artuhov, could you please clarify why `src/test` and `src/main` should be present in the `integration-test` core?

dmitrii-artuhov:
@ndkoval I wrote them just as an example, in fact only `src/test` is present because it stores common code for trace-recorder/debugger (`AbstractTraceIntegrationTest.kt` class).
Not sure if it should be `test` source set and not `main`. But this is changed easily, so does not really matter which one I use, imho

ndkoval:
@dmitrii-artuhov what do you think about an alternative structure without having a module containing both source code and submodules?

```
:integration-test
    * common
    * lincheck
    * trace-debugger
    * trace-recorder
```

dmitrii-artuhov:
@ndkoval well, I guess I am not very good at drawing diagrams) The structure right now is exactly what you have written. The only difference is that `common` source set is now called `src/test`, I can rename it 

By `:root` I mean the root of the project, basically `:`

integration-test/lincheck/org/jetbrains/lincheck_test/guide/ConcurrentMapTest.kt
```diff
@@ -41,6 +43,6 @@ class ConcurrentSkipListMapTest {
     @Test
     fun modelCheckingTest() = ModelCheckingOptions()
         .checkObstructionFreedom(true)
-        .analyzeStdLib(true)
+        .analyzeStdLib(true) //  TODO: cannot access internal APIs of the main source set, old way does not work between projects
```
dmitrii-artuhov: Remove `TODO`, it is now fixed

settings.gradle.kts
```diff
@@ -26,3 +26,5 @@ rootProject.name = name
 
 include(":bootstrap")
 include(":trace")
+
```
dmitrii-artuhov: Remove new-line here

ndkoval:
@dmitrii-artuhov, should common, lincheck, trace-debugger, and trace-recorder be source sets or submodules? I would suggest the latter, as it would (1) give more flexibility in how we run these tests, (2) ensure these parts do not depend on each other (besides `common`), and (3) make it easy to move them later if needed.

buildSrc/src/main/kotlin/Toolchain.kt
```diff
@@ -37,4 +40,49 @@ fun KotlinCompile.setupKotlinToolchain(javaToolchains: JavaToolchainService, jdk
     kotlinJavaToolchain.toolchain.use(javaToolchains.launcherFor {
         languageVersion.set(JavaLanguageVersion.of(jdkToolchainVersion))
     })
+}
+
+fun Test.configureJvmTestCommon(project: Project) {
```
eupp: Let's move to a new file `Testing.kt`

eupp (changes requested):
Let's organize integration tests into sub-projects, instead of source sets, as @ndkoval suggested. 

Otherwise looks good, thanks!

integration-test/trace-debugger/org/jetbrains/test/trace/debugger/AbstractTraceDebuggerIntegrationTest.kt
```diff
@@ -8,13 +8,12 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package org.jetbrains.trace_debugger.integration
+package org.jetbrains.test.trace.debugger
```
dmitrii-artuhov: Here we can name the testing packages differently. As we discussed, no `_` underscores should be left in a code base. The `test` keyword here is first, because later when we change the `*.lincheck_test` package we cannot make it `*.lincheck.test`, because then these classes will not be transformed. However, we can introduce additional check in all necessary places, that if package name starts with `*.lincheck.test`, then we transform such classes. Or just name it `org.jetbrains.test.lincheck` to match the pattern I used here

dmitrii-artuhov: What do you think? Which looks better:
- `org.jetbrains.lincheck.test` 
   `org.jetbrains.trace.debugger.test` 
   `org.jetbrains.trace.recorder.test`
   
- `org.jetbrains.test.lincheck` 
   `org.jetbrains.test.trace.debugger`
    `org.jetbrains.test.trace.recorder`

eupp: To me `lincheck.test` is the most logical think, so I would instead change the instrumentation check. 
But let's do this in a separate PR to speed-up the process and avoid accidental bugs. 
For now I would propose to keep `lincheck_test` everywhere, so we can easily fix it everywhere all-at-once later.

integration-test/trace-debugger/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    processResources {
+        duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    }
+
+    registerTraceAgentIntegrationTestsPrerequisites()
+
+    val traceDebuggerIntegrationTest = register<Test>("traceDebuggerIntegrationTest") {
+        configureJvmTestCommon(project)
+        group = "verification"
+        include("org/jetbrains/test/trace/debugger/*")
```
eupp: Now that integration tests are split into sub-projects, do you still need this?

integration-test/trace-recorder/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    processResources {
+        duplicatesStrategy = DuplicatesStrategy.INCLUDE
+    }
+
+    registerTraceAgentIntegrationTestsPrerequisites()
+
+    val traceRecorderIntegrationTest = register<Test>("traceRecorderIntegrationTest") {
+        configureJvmTestCommon(project)
+        group = "verification"
+        include("org/jetbrains/test/trace/recorder/*")
```
eupp: Now that integration tests are split into sub-projects, do you still need this?

integration-test/lincheck/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+    main {
+        java.srcDirs("src/main")
+    }
+
+    dependencies {
+        val junitVersion: String by project
+        val jctoolsVersion: String by project
+
+        implementation(project(":"))
+        implementation(rootProject.sourceSets["test"].output)
```
eupp, line 13: Newer saw dependencies declared like this before, is it ok?
Anyway, what exactly it depends on?
Depending on the answer, we may consider moving it to `common`.

integration-test/trace-recorder/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+
+        testClassesDirs = sourceSets["main"].output.classesDirs
+        classpath = sourceSets["main"].runtimeClasspath
```
eupp, line 41: I think previously it was for `rootProject`'s `main` source set. Do we still need it?

integration-test/trace-debugger/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+
+        testClassesDirs = sourceSets["main"].output.classesDirs
+        classpath = sourceSets["main"].runtimeClasspath
```
eupp, line 41: I think previously it was for rootProject's main source set. Do we still need it?

eupp:
Please also rebase on `develop` and align with previous modularization PR.

integration-test/trace-recorder/build.gradle.kts
dmitrii-artuhov: I need it, because classes of 'main' source set of the current project should be treated as tests. Changing the source sets of all integration tests to 'test' instead of 'main' would probably help, but then ':test' task will include all integration tests which we don't want

integration-test/lincheck/build.gradle.kts
eupp: I meant what definitions from `rootProject` test source set are required in integration tests? We might wanna move them to `common` instead.

integration-test/trace-recorder/build.gradle.kts
eupp: So ':test' task runs tests from all subprojects, right?
Is there any simpler way to disable this behavior, apart from moving tests to `main` source set?

eupp (commented):
See the comments in threads.

PR #742, opened 2025-07-01
Handle inline functions  non-local return

eupp:


lev-serebryakov-jetbrains:
Please, review as lev-serebryakov/inline-methods-revamp and delete this PR.

PR #744, opened 2025-07-04
Modularization of the project

dmitrii-artuhov:
TODOs:
- [x] fix publication and lincheck artifact contents (it misses `bootstrap.jar`, classes from the modules that `:core` depends on, etc.)
- [x] refactor plugins to account for different task names and fat-jars -- PR https://jetbrains.team/p/concurrency-tools/reviews/65/timeline, cc @ivandev0 
- [x] fix idea plugin integration
- [x] try friend modules

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -136,7 +136,8 @@ sourceSets {
         val atomicfuVersion: String by project
 
         compileOnly(project(":bootstrap"))
-        api(project(":common"))
+        api(project(":common")) // TODO: `api` is used here in order to allow users of `lincheck` to access `LoggingLevel` enum class stored there,
+                                //  but later this enum will be marked as deprecated and then hidden, after that `api` should be changed to `implementation`
         implementation(project(":jvm-agent"))
         implementation(project(":trace"))
 


build.gradle.kts
```diff
@@ -40,479 +16,21 @@ java {
...
-                println("Upload status code: $statusCode")
-                println("Upload result: ${response.body!!.string()}")
-                if (statusCode != 201) {
-                    error("Upload error to Central repository. Status code $statusCode.")
-                }
-            }
-        }
-    }
+    into(file("${project(":jvm-agent").layout.buildDirectory.get()}/resources/main"))
 }
```
eupp: Could we please leave the Lincheck's code in the root's project for now.  Moving it into a separate subproject is a separate task.

There are several problems with moving is as for now. 

1) The publication is broken (from what I understood from the comments to this PR). Setting it up was not a trivial task, let's please not break it immediately.  
2) The code was moved to `core`, but we decided on the teams' meeting we want to use different naming scheme.

It's better to postpone moving it than doing it in a rush 
The task of moving Lincheck has smaller priority than extracting trace-recorder jar, so we are not in a rush here.

buildSrc/src/main/kotlin/TraceAgentTasks.kt
```diff
@@ -27,13 +27,17 @@ fun Project.registerTraceAgentTasks() {
     val mainSourceSet = javaPluginExtension.sourceSets.getByName("main")
 
     val runtimeClasspath = configurations.getByName("runtimeClasspath")
+    val compileClasspath = configurations.getByName("compileClasspath")
+    println("Runtime classpath: ${runtimeClasspath.resolve().joinToString("\n")}")
+    println("\n\n")
+    println("Compile classpath: ${compileClasspath.resolve().joinToString("\n")}")
```
eupp: Leftover `println`-s ?

buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
```diff
@@ -63,6 +63,18 @@ fun Project.registerTraceAgentIntegrationTestsPrerequisites() {
...
     }
+}
+
+fun Project.copyTraceAgentFatJar(fromProject: Project): TaskProvider<Copy> {
+    val copyTraceAgentFatJar = tasks.register<Copy>("${fromProject.name}_copyTraceAgentFatJar") {
+        dependsOn(traceAgentIntegrationTestsPrerequisites)
+        val fatJarFile = fromProject.layout.buildDirectory.file("libs/lincheck-fat.jar")
+        from(fatJarFile)
+        into(layout.buildDirectory.dir("libs"))
+    }
```
eupp, line 85: What does this task do? Why there is a need to copy fat jar?

settings.gradle.kts
```diff
@@ -25,4 +25,9 @@ val name: String by settings
 rootProject.name = name
 
 include(":bootstrap")
+include(":jvm-agent")
+include(":core")
+include(":common")
 include(":trace")
+include(":trace-recorder")
+include("trace-debugger")
```
eupp: `include(":trace-debugger")`

trace-recorder/src/main/org/jetbrains/kotlinx/lincheck/traceagent/TraceAgent.kt
```diff
@@ -0,0 +1,59 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.traceagent
```
eupp: We discussed on the meeting we want to have the following package names:
- `trace.agent`
- `trace.recorder`
- `trace.debugger`

etc.

Are there any technical problems with using these package names?

jvm-agent/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+
+    dependencies {
+        // main
+        val kotlinVersion: String by project
+        val asmVersion: String by project
+        val byteBuddyVersion: String by project
+
+        compileOnly(project(":bootstrap"))
+        api(project(":common"))
+        api(project(":trace"))
```
eupp: I think`trace` shouldn't be a dependency of `jvm-agent`.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/traceagent/TraceAgentTransformer.kt
```diff
@@ -0,0 +1,105 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.traceagent
```
eupp: I am a bit confused here. 

Some of the ASM-related classes are moved into `jvm-agent` subprojects, others are moved into `trace-recorder` or `trace-debugger` subprojects. What is the general rule here? 

Could we use a more consistent approach? So either:

- Keep all transformation classes in `jvm-agent` subproject, or
- Keep all trace-recorder/debugger specific transformers in their respective subprojects.

jvm-agent/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+        setupJavaToolchain(project)
+    }
+    named<KotlinCompile>("compileTestKotlin") {
+        setupKotlinToolchain(project)
+    }
+
+    withType<Jar> {
+        dependsOn(":bootstrapJar")
+    }
+}
```
eupp, line 90: Let's please follow the `trace` subproject template, and treat each sub-project as a separate artifact, setting-up for it:

- documentation
- maven pom
- publication 

E.g., for `jvm-agent` subproject, we should set-up a separate artifact `org.jetbrains.lincheck:lincheck-jvm-agent`. Same for other subprojects.

docs/topics/constraints.md
```diff
@@ -94,7 +94,7 @@ Here is an example of the scenario generated for this test:
 Note that all consuming `poll()` and `peek()` invocations are performed from a single thread, thus satisfying the
 "single-consumer" restriction.
 
-> [Get the full code](https://github.com/JetBrains/lincheck/blob/master/src/jvm/test-lincheck-integration/org/jetbrains/lincheck_test/guide/MPSCQueueTest.kt).
+> [Get the full code](https://github.com/JetBrains/lincheck/blob/master/common/src/test-lincheck-integration/org/jetbrains/lincheck_test/guide/MPSCQueueTest.kt).
```
eupp: Why the location changed here? 
It looks like the guide tests are not located in `common` ?

common/src/main/org/jetbrains/kotlinx/lincheck/util/TraceModes.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.util
+
+private const val TRACE_DEBUGGER_MODE_PROPERTY = "lincheck.traceDebuggerMode"
+private const val TRACE_RECORDER_MODE_PROPERTY = "lincheck.traceRecorderMode"
+val isInTraceDebuggerMode by lazy { System.getProperty(TRACE_DEBUGGER_MODE_PROPERTY, "false").toBoolean() }
+val isInTraceRecorderMode by lazy { System.getProperty(TRACE_RECORDER_MODE_PROPERTY, "false").toBoolean() }
```
eupp: Should it be kept in `common` ? Can we move it to trace-related subprojects ?

core/build.gradle.kts
```diff
@@ -0,0 +1,500 @@
...
+
+        enableAssertions = true
+        testLogging.showStandardStreams = true
+        outputs.upToDateWhen { false } // Always run tests when called
+
+        forkEvery = 1
+    }
+
+    // TODO: rename to match trace-debugger/recorder gradle task naming pattern to 'lincheckIntegrationTest'
+    val lincheckIntegrationTest = register<Test>("integrationTest") {
```
eupp: BTW, another problem with moving files to `core` module in this PR is that it is probably going to conflict with integration tests re-organization PR https://github.com/JetBrains/lincheck/pull/739

trace/build.gradle.kts
```diff
@@ -2,8 +2,6 @@ import org.gradle.kotlin.dsl.named
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
-    java
-    kotlin("jvm")
```
eupp, line 6: Why these plugins were removed?

common/build.gradle.kts
```diff
@@ -0,0 +1,41 @@
...
+    main {
+        java.srcDirs("src/main")
+    }
+
+    dependencies {
+        // main
+        val kotlinVersion: String by project
+
+        compileOnly(project(":bootstrap"))
+        api(project(":trace")) // TODO: can this be fixed?
```
eupp: I think `common` should not depend on `trace`, it should be the other way around.

eupp (changes requested):

buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
dmitrii-artuhov: TODO: rewrite trace-recorder/debugger into completely different gradle tasks and rename agent classes properly, change manifests and do an obviously separation between them instead of the current implicit approach

jvm-agent/build.gradle.kts
dmitrii-artuhov: TODO: from `:trace` module all stuff related to `Types`, `MethodDecriptors`, etc. should be put in `:common` module instead. But be careful with #738 (contact Lev)

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/traceagent/TraceAgentTransformer.kt
dmitrii-artuhov: TODO: add clarifying comment everywhere

common/src/main/org/jetbrains/kotlinx/lincheck/util/TraceModes.kt
dmitrii-artuhov: TODO: Flags related to the `TraceModes`, `ideaPluginEnabled` move into a single file inside a `:common` module

dmitrii-artuhov:
Try using friend-paths/associate-source-sets in order to handle the visibility of the classes inside `:common` (some, like `LogLevel`, must be available to users, while others, like `isInTraceDebugger` flag, should not).
Research that. If it does not work, try searching for `:private-common`/`:public-common` separation, if it is a thing in gradle world

common/src/main/org/jetbrains/kotlinx/lincheck/trace/CodeLocations.kt
```diff
@@ -8,7 +8,7 @@
  * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-package org.jetbrains.lincheck.trace
+package org.jetbrains.kotlinx.lincheck.trace
```
eupp: Let's rename package to `org.jetbrains.lincheck.descriptors`

common/src/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -10,10 +10,6 @@
 
 package org.jetbrains.kotlinx.lincheck.util
```
eupp: Let's change the package to `org.jetbrains.lincheck.util`

common/src/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ShadowStack.kt
```diff
@@ -11,11 +11,11 @@
 package org.jetbrains.kotlinx.lincheck.strategy.managed
```
eupp: Let's change package to `org.jetbrains.lincheck.analysis`

common/build.gradle.kts
```diff
@@ -0,0 +1,75 @@
...
+            this.artifactId = commonArtifactId
+            this.version = commonVersion
+
+            from(components["kotlin"])
+            artifact(sourcesJar)
+            artifact(javadocJar)
+
+            configureMavenPublication {
+                name.set(commonArtifactId)
+                description.set("Lincheck common classes library")
```
eupp: Let's say: `"Lincheck common utilities library"` or `"Lincheck common components library"`

jvm-agent/build.gradle.kts
```diff
@@ -0,0 +1,90 @@
...
+            this.artifactId = jvmAgentArtifactId
+            this.version = jvmAgentVersion
+
+            from(components["kotlin"])
+            artifact(sourcesJar)
+            artifact(javadocJar)
+
+            configureMavenPublication {
+                name.set(jvmAgentArtifactId)
+                description.set("Lincheck jvm agent library")
```
eupp: Maybe let's spell JVM with capital letters? What about?:
`"Lincheck JVM agent instrumentation library"`

buildSrc/src/main/kotlin/Documentation.kt
```diff
@@ -10,22 +10,23 @@
...
             named("main") {
-                sourceRoots.from(file("src/jvm/main"))
+                sourceRoots.from(file(mainSourceSetPath))
                 reportUndocumented.set(false)
             }
         }
     }
 
     val javadocJar = tasks.register<Jar>("javadocJar") {
+        archiveBaseName.set(this@createJavadocJar.name)
```
eupp: Are you sure this is required? Before this change, the name of the generated javadoc was e.g., `lincheck-trace-javadoc.jar` for the `trace` artifact, which was expected behavior I think. What would be the name after this change?

buildSrc/src/main/kotlin/Toolchain.kt
```diff
@@ -51,4 +52,17 @@ private fun KotlinCompile.setupKotlinToolchain(javaToolchains: JavaToolchainServ
...
         languageVersion.set(JavaLanguageVersion.of(jdkToolchainVersion))
     })
+}
+
+fun KotlinCompile.getAccessToInternalDefinitionsOf(vararg projects: Project) {
+    projects.forEach { project ->
+        val mainSourceSet = project.extensions
+            .getByType(JavaPluginExtension::class.java).sourceSets
+            .getByName("main").output.classesDirs
+        val jarArchive = Paths.get(project.buildDir.absolutePath, "libs",
```
eupp: Please fix deprecated `buildDir`. There should already be some examples in the main `build.gradle.kts` script.

buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
```diff
@@ -63,6 +63,18 @@ fun Project.registerTraceAgentIntegrationTestsPrerequisites() {
 
     traceAgentIntegrationTestsPrerequisites = tasks.register("traceAgentIntegrationTestsPrerequisites") {
         prerequisite.forEach { dependsOn(it) }
-        dependsOn("traceAgentFatJar")
+        dependsOn(":trace-debugger:traceDebuggerFatJar")
+        dependsOn(":trace-recorder:traceRecorderFatJar")
     }
+}
+
+fun Project.copyTraceAgentFatJar(fromProject: Project, fatJarName: String): TaskProvider<Copy> {
```
eupp, line 79: Please add a comment why it is needed.

buildSrc/build.gradle.kts
```diff
@@ -23,4 +24,9 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
     implementation("org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion")
     implementation("de.undercouch.download:de.undercouch.download.gradle.plugin:5.6.0")
+
+    implementation(gradleApi())
+    implementation(gradleKotlinDsl())
+    implementation(kotlin("gradle-plugin"))
```
eupp: Isn't it the same as the line above?:
```kotlin
implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
```

buildSrc/build.gradle.kts
```diff
@@ -23,4 +24,9 @@ dependencies {
...
+    implementation("org.gradle.kotlin:gradle-kotlin-dsl-plugins:4.2.1")  // Add this line
```
eupp: What is the purpose of `// Add this line` comment?

buildSrc/build.gradle.kts
```diff
@@ -23,4 +24,9 @@ dependencies {
...
 }
```
eupp: Please also sort imports from more generic one to more specific. 
I.e. start with gradle and kotlin imports, then say dokka and download.

src/jvm/main/org/jetbrains/kotlinx/lincheck/util/MethodSignature.kt
```diff
@@ -10,8 +10,8 @@
 
 package org.jetbrains.kotlinx.lincheck.util
 
-import org.jetbrains.lincheck.trace.MethodSignature
-import org.jetbrains.lincheck.trace.Types
+import org.jetbrains.kotlinx.lincheck.trace.MethodSignature
+import org.jetbrains.kotlinx.lincheck.trace.Types
 import org.objectweb.asm.commons.Method
```
eupp: Can we move this file to `common` ?

common/src/main/org/jetbrains/kotlinx/lincheck/util/AnalysisSections.kt
```diff
@@ -304,12 +308,6 @@ internal class AnalysisProfile(val analyzeStdLib: Boolean) {
     @Suppress("UNUSED_PARAMETER") // methodName is here for uniformity and might become useful in the future
     fun shouldBeHidden(className: String, methodName: String): Boolean = 
         !analyzeStdLib && (isConcurrentCollectionsLibrary(className) || isCollectionsLibrary(className))
-
-    companion object {
-        val DEFAULT = AnalysisProfile(
-            analyzeStdLib = ManagedCTestConfiguration.DEFAULT_STDLIB_ANALYSIS_ENABLED
-        )
-    }
```
eupp: Can we keep it here? You can invert the dependencies and instead re-define `ManagedCTestConfiguration.DEFAULT_STDLIB_ANALYSIS_ENABLED = AnalysisProfile.DEFAULT.analyzeStdLib`

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -37,7 +44,7 @@ sealed class TRTracePoint(
     abstract fun toText(verbose: Boolean): String
 }
 
-class TRMethodCallTracePoint(
+internal class TRMethodCallTracePoint(
```
eupp: These classes should not be `internal`. They are used on the plugin side to de-serialize data from trace dump back to objects.

trace/src/main/org/jetbrains/lincheck/trace/Serialization.kt
```diff
@@ -10,14 +10,24 @@
...
+import org.jetbrains.kotlinx.lincheck.trace.VariableDescriptor
 import java.io.*
 
 private const val OUTPUT_BUFFER_SIZE: Int = 16*1024*1024
 
 const val TRACE_MAGIC : Long = 0x706e547124ee5f70L
 const val TRACE_VERSION : Long = 6
 
-fun saveRecorderTrace(out: OutputStream, context: TraceContext, rootCallsPerThread: List<TRTracePoint>) {
+internal fun saveRecorderTrace(out: OutputStream, context: TraceContext, rootCallsPerThread: List<TRTracePoint>) {
```
eupp: I think these should also be kept public. 
Please rollback all `internal` changes in this module.

trace-debugger/src/main/org/jetbrains/kotlinx/lincheck/trace/debugger/TraceDebuggerAgent.kt
```diff
@@ -0,0 +1,55 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.trace.debugger
```
eupp: We can now safely switch the package to  `"org.jetbrains.lincheck.trace.debugger"` in this module.

trace-recorder/src/main/org/jetbrains/kotlinx/lincheck/trace/recorder/TraceRecorderAgent.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck.trace.recorder
```
eupp: We can now safely switch the package to  `"org.jetbrains.lincheck.trace.recorder"` in this module.

eupp (changes requested):

common/build.gradle.kts
dmitrii-artuhov: I like `components` more, since there are not only utilities in this subproject

build.gradle.kts
```diff
@@ -96,7 +136,9 @@ sourceSets {
         val atomicfuVersion: String by project
 
         compileOnly(project(":bootstrap"))
-        api(project(":trace"))
+        api(project(":common")) // TODO: contains some classes from public lincheck API, refactor so that it does not expose essentially internal classes
```
dmitrii-artuhov: Right now `:common` subproject has one class that should be available to the users of lincheck, which is `enum class LogLevel`. However, the rest of the symbols in it, are implementation-only. There are 2 options how to deal with that:
- [Currently used] make all symbols in `:common` internal (except for `LogLevel`) and include in the root project the subproject via `api(project(":common"))`. The problem here is that it is done via friend paths, which lack support in intellij idea, which then marks all usages on `internal` classes from `:common` reads and does not provide completion and auto-importing
- Create separate "common" modules, one called `:api`, which will contain `LogLevel` (and probably other classes in the future, the second one `:common` with all classes marked as `public`. Now, the root project will include them as follows: `api(project(":api")); implementation(":common")`, which allows users to access `LogLevel` enum and disallows all classes from `:common`. It also supports proper highlighting and auto-imports as regular.

eupp: For now, let's just make `public` everything that need to be public to work properly both in gradle and IDEA. 
Please create a YT issue to set-up `internal` isolation visibility properly. It is not critical right now, but would be good to fix before next Lincheck release.

eupp: > For now, let's just make public everything that need to be public to work properly both in gradle and IDEA.

Or use any other solution that would make code work (both in gradle and IDEA) right now, without further design/implementation.

PR #747, opened 2025-07-09
Always collect event ids and initialize them in `TracePoint` constructor

eupp:


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -747,8 +717,9 @@ internal abstract class ManagedStrategy(
         // scenario threads are handled separately by the runner itself
         if (isTestThread(currentThreadId)) return
 
+        val eventId = getNextEventId()
```
ivandev0, line 734: Minor. I think extracting event ID in a separate variable is excessive because it is not used anywhere else unlike thread ID which can be reused.

eupp: I extracted `val eventId = getNextEventId()` (almost) everywhere for two purposes:

1. Uniformity with `threadId` and between all the injection methods. 

2. As a part of the new contract (documented in one of the comments in the code), `beforeEvent` assumes that each time it is called, previous injection method has advanced the `eventId` for the new event. But some injections (e.g., `afterReadField`) create `TracePoint` conditionally. 

Thus, by moving  `val eventId = getNextEventId()` at the beginning of the methods, I ensure the id allocated for each injection/event, no matter if trace point was created or not.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -296,6 +261,7 @@ internal abstract class ManagedStrategy(
         monitorTracker.reset()
         parkingTracker.reset()
         constants.clear()
+        currentEventId = -1
```
zuevmaxim, line 263: Why not in `initializeReplay`?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1254,11 +1249,10 @@ internal abstract class ManagedStrategy(
     /**
```
zuevmaxim: the comment is outdated

zuevmaxim (approved):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: `initializeInvocation` is called before each new invocation (both replay of current interleaving and exploring a new interleaving), while `initializeReplay()` is only called on replays of the same interleaving.

PR #748, opened 2025-07-11
Add gradle tasks to simplify Space Packages publishing

eupp:


jvm-agent/gradle.properties
```diff
@@ -9,4 +9,4 @@
 #
 
 jvmAgentArtifactId=lincheck-jvm-agent
-jvmAgentVersion=0.1-SNAPSHOT
\ No newline at end of file
+jvmAgentVersion=3.1-SNAPSHOT
```
dmitrii-artuhov, line 12: Is it expected that `:jvm-agent` version jumped to `3.1`, while `:common` is only at `3.0.2`

trace/gradle.properties
```diff
@@ -11,4 +11,4 @@
 # use `trace` prefix as otherwise gradle resolves the version property to root's version,
 # for artifactId it resolves correctly, however, we still use the prefix for uniformity
 traceArtifactId=lincheck-trace
-traceVersion=0.1-SNAPSHOT
\ No newline at end of file
+traceVersion=3.0.2-SNAPSHOT
```
dmitrii-artuhov, line 14: Same question for the `:trace`, should it be `3.1-SNAPSHOT` instead?

dmitrii-artuhov (approved):

PR #749, opened 2025-07-11
Add function for setting up manifest's version and copyrights

dmitrii-artuhov:


buildSrc/src/main/kotlin/Toolchain.kt
```diff
@@ -66,4 +68,18 @@ fun KotlinCompile.getAccessToInternalDefinitionsOf(vararg projects: Project) {
         ).toFile()
         friendPaths.from(mainSourceSet + jarArchive)
     }
+}
+
+fun Manifest.appendMetaAttributes(project: Project) {
```
eupp: Please move to `Publishing.kt`. 
`Toolchain.kt` is about setting-up compiler(s) toolchain.

eupp (changes requested):

PR #752, opened 2025-07-11
Lev serebryakov/tr bugfixes on kotlin tests

lev-serebryakov-jetbrains:
Fix problems when thread buffer is overflown.

Limit string size to buffer size, too.

This fix Kotlin's `:compiler:tests-integration:test --tests "org.jetbrains.kotlin.cli.CliTestGenerated$Jvm.testExtraHelp" `

But this test generates 10GiB of Strings, so it will need A LOT of memory to load trace.

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
@@ -575,8 +575,9 @@ fun TRObjectOrVoid(obj: Any?): TRObject? =
     else TRObjectOrNull(obj)
 
 
-private fun trimString(s: CharSequence): String = s.take(min(50, s.length)).toString()
+const val MAX_TROBJECT_STRING_LENGTH = 50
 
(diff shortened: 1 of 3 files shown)


trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -573,6 +574,10 @@ fun TRObjectOrVoid(obj: Any?): TRObject? =
     if (obj == INJECTIONS_VOID_OBJECT) TR_OBJECT_VOID
     else TRObjectOrNull(obj)
 
+
+private fun trimString(s: CharSequence): String = s.take(min(50, s.length)).toString()
```
eupp: Should we make `50` a configurable parameter? Or at least extract it into a constant.

lev-serebryakov-jetbrains: We decided to add options as separate feature (see Slack channel), but constant is good idea.

ivandev0 (approved):

ndkoval:
@lev-serebryakov-jetbrains Please change the PR title to indicate what is fixed.

Issue #717, opened 2025-06-20
Investigate flaky bugs caused by instrumentation of `java.util.Arrays` class


eupp:
Looks like, when `java.util.Arrays` is instrumented and mutable lists (`ArrayList`) is used somewhere in instrumentation code it leads to some complex flaky bug. The reason also could be related to the fact that `Arrays` has hotspot intrinsic candidate methods, and perhaps the issue is somehow caused by jitting of instrumented methods.

We've already observed similar bugs in the past:
- https://github.com/JetBrains/lincheck/issues/508
- https://github.com/JetBrains/lincheck/issues/470

PR #758, opened 2025-07-14
Refactor `LincheckClassVisitor`

eupp:
This is just a refactoring of ``LincheckClassVisitor.kt` to clean it up before further development. 

- Group together coupled transformer chains and extract their application into methods. 
- Extract `shouldWrapInIgnoredSection` and `shouldNotInstrument` to unify ignored/non-instrumented cases.
- Rename `ManagedStrategyMethodVisitor` into `LincheckBaseMethodVisitor`.
- Minor clean-up.


jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -291,81 +232,91 @@ internal class LincheckClassVisitor(
...
-        visitCode()
-        invokeStatic(Injections::enterIgnoredSection)
-        visitTryCatchBlock(tryBlock, catchBlock, catchBlock, null)
-        visitLabel(tryBlock)
+    private fun shouldNotInstrument(className: String, methodName: String, descriptor: String): Boolean {
+        // Do not instrument `ClassLoader` methods.
+        if (isClassLoaderClassName(className.toCanonicalClassName()))
+            return true
+        // Instrumentation of `java.util.Arrays` class causes some subtle flaky bugs.
+        // See details in https://github.com/JetBrains/lincheck/issues/717.
```
dmitrii-artuhov, line 283: Must be the intrinsics problem, some classes which do not have `@IntrinsicCandidate` are still in fact intrinsified and delete our instrumentation mid-execution

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -153,7 +153,10 @@ internal class TraceReporter(
                 (it is MethodReturnTracePoint) ||
                 it is SpinCycleStartTracePoint
             }
-            val isThreadJoinSwitch = (remainingTracePoints.firstOrNull()?.isThreadJoin() == true)
+            val isThreadJoinSwitch = (movedTracePoints.lastOrNull()?.isThreadJoin() == true) &&
+                // check that Thread.join() call has return value set, otherwise it is a hung join
+                (movedTracePoints.last() as MethodCallTracePoint).returnedValue != ReturnedValueResult.NoValue
```
dmitrii-artuhov, line 160: Is there a test that distinguishes these behaviours? If yes, please add it

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -111,97 +111,59 @@ internal class LincheckClassVisitor(
...
+            mv = IgnoredSectionWrapperTransformer(fileName, className, methodName, mv.newAdapter())
             return mv
         }
-        /*
-         * Instrumentation of `java.util.Arrays` class causes some subtle flaky bugs.
-         * See details in https://github.com/JetBrains/lincheck/issues/717.
-         */
-        if (isJavaUtilArraysClass(className.toCanonicalClassName())) {
-            // `java.util.Arrays` contains intrinsic methods --- we need to process them
+        if (shouldNotInstrument(className, methodName, desc)) {
```
dmitrii-artuhov, line 148: should this `if` go before the previous one with `shouldWrapInIgnoredSection` call?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: This comment was here before this PR, I just moved & re-organized some code. 
The tracking issue is this one: https://github.com/JetBrains/lincheck/issues/717

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
eupp: `LivelockRunConcurrentRepresentationTest` distinguish this case. 
In fact, it is because this test was failing I had to add this line.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: Currently, the code logic assumes `shouldWrapInIgnoredSection` happens first. 

For instance, consider `<clinit>`. Currently ignored section wrapping would happen for `<clinit>` of every class. However, if we re-order the two cases, then it will not be wrapped for classes for which `shouldNotInstrument` returns `false`. 

I agree it is sub-optimal solution and the logic is not entirely clear here.
However, it was the same behavior before this refactoring. I decided to preserve it just in case, to avoid accidental bugs.

lev-serebryakov-jetbrains (approved):
Looks good.

PR #759, opened 2025-07-15
Wrap IntelliJ runtime debugger/coverage agent methods into ignored sections

eupp:
Quickfix for a bug in debugger plugin integration.


jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/TransformationUtils.kt
```diff
@@ -651,6 +651,24 @@ internal fun isJavaUtilArraysClass(className: String): Boolean =
...
+/**
+ * Checks if the given class name belongs to the IntelliJ runtime coverage agent package.
+ */
+internal fun isIntellijRuntimeCoverageAgentClass(className: String) =
+    className.startsWith("com.intellij.rt.coverage")
+
+/**
+ * Checks if the given class name belongs to the IntelliJ runtime agents.
+ */
+internal fun isIntellijRuntimeAgentClass(className: String) =
```
dmitrii-artuhov: Can you please search for these package names in entering codebase,I think this is also user in some code related to snapshot. That place should be rewritten using this function.

Apart from that lgtm

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassFileTransformer.kt
```diff
@@ -170,10 +170,12 @@ object LincheckClassFileTransformer : ClassFileTransformer {
...
+        // IntelliJ runtime agents, to wrap all their methods into the ignored section.
+        isIntellijRuntimeAgentClass(className) ||
         // `ThreadContainer` classes, to detect threads started in the thread containers.
         isThreadContainerClass(className) ||
         // TODO: instead of eagerly instrumenting `DispatchedContinuation`
         //  we should try to fix lazy class re-transformation logic
         isCoroutineDispatcherInternalClass(className) ||
         isCoroutineConcurrentKtInternalClass(className)
-}
+}
```
dmitrii-artuhov: Do we need this change?

dmitrii-artuhov (approved):

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/TransformationUtils.kt
dmitrii-artuhov: for snapshot it was removed as it seems, in one of the old PRs

PR #762, opened 2025-07-16
Simple trace polishing

dmitrii-artuhov:
This PR applies the same transformations on trace as for regular lincheck trace points. But these transformations do not modify the tree structure: e.g., do not remove a node and reappend its children to the parent, etc -- only per-trace-point modifications


trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrettifiers.kt
```diff
@@ -0,0 +1,119 @@
...
+
+internal sealed interface TRTracePointPrettifier {
+    fun dump(dest: Appendable, verbose: Boolean)
+}
+
+internal class TRMethodCallTracePointPrettifier(
+    val point: TRMethodCallTracePoint,
+    val parameters: List<TRObject?>,
+    val methodDescriptor: MethodDescriptor,
+): TRTracePointPrettifier {
```
eupp: If we are really into making an architecture (c) here, I propose a slightly different design based on visitors and foldables. 

Introduce a set of interfaces:
```kotlin
class TRMethodCallTracePointPrinter {
    fun Appendable.append(tracePoint: MethodCallTracePoint)
}
```

for each trace point type. 

Then you can have a base class implementation, that "prints the skeleton", delegating printing different subparts to an implementation:

```kotlin
class AbstractTRMethodCallTracePointPrinter() {
    override fun Appendable.append(tracePoint) {
        dest.appendClassName(tracePoint)
        dest.append('.')
        dest.appendMethodName(tracePoint)
        dest.append('(')

        ...
    }
    
    abstract fun Appendable.appendClassName(tracePoint: MethodCallTracePoint)
    abstract fun Appendable.appendMethodName(tracePoint: MethodCallTracePoint)
    ...
}
```

Then you can provide an implementation, that takes all the options (e.g., `verbose` flag, `removeCoroutinesCoreSuffix` flag, etc) as arguments:

```kotlin
class TRMethodCallTracePointPrinterImpl(
    val verbose: Boolean,
    val removeCoroutinesCoreSuffix: Boolean, 
    ...
) : AbstractTRMethodCallTracePointPrinter() { ... }
```

Also, it would be better to then make the default printer a static global:
```kotlin
val DefaultTRMethodCallTracePointPrinter = ...
```

So that by default, in the `toText()` method implementation in `MethodCallTracePoint` you avoid object allocation.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrettifiers.kt
```diff
@@ -0,0 +1,119 @@
...
+    fun removeCoroutinesCoreSuffix(): TRMethodCallTracePointPrettifier {
+        methodName = methodName.removeSuffix("\$kotlinx_coroutines_core")
+        return this
+    }
+
+    override fun dump(dest: Appendable, verbose: Boolean) {
+        dest.append(className)
+            .append('.')
+            .append(methodName)
+            .append('(')
```
eupp: Here and in other similar places `with(dest) { ... }` can help.

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.jetbrains.lincheck.util.isJavaLambdaClass
+
+/**
+ * Replaces nested class dollars (if present) from string with dots.
+ */
+fun replaceNestedClassDollar(nestedClassRepresentation: String): String {
+    val before = nestedClassRepresentation.substringBefore('$')
+    val after = nestedClassRepresentation.substringAfter('$', "")
+    if (after.isEmpty()) return nestedClassRepresentation
+    val firstPart = if (before.isNotEmpty() && before[0].isUpperCase() && after[0].isUpperCase()) "$before."
```
eupp: BTW, can we extract this check `before.isNotEmpty() && before[0].isUpperCase() && after[0].isUpperCase()` into `val ... = ` or in a separate function and give it a meaningful name?

Otherwise it is unclear what it checks for.

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
```diff
@@ -0,0 +1,33 @@
...
+    else "$before$"
+    return firstPart + replaceNestedClassDollar(after)
```
eupp: Can we avoid recursion here?

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
```diff
@@ -0,0 +1,33 @@
...
+package org.jetbrains.lincheck.trace
+
+import org.jetbrains.lincheck.util.isJavaLambdaClass
+
+/**
+ * Replaces nested class dollars (if present) from string with dots.
+ */
+fun replaceNestedClassDollar(nestedClassRepresentation: String): String {
+    val before = nestedClassRepresentation.substringBefore('$')
+    val after = nestedClassRepresentation.substringAfter('$', "")
```
eupp: BTW, these `$` signs are an artifact of our not-fully-correct internal-to-canonical class name conversion (implemented in `toCanonicalClassName()` method):

https://www.baeldung.com/java-class-name#3-inner-classes


Let's create an issue for that, because it is not the first time we encounter this.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrettifiers.kt
```diff
@@ -0,0 +1,119 @@
...
+    fun dump(dest: Appendable, verbose: Boolean)
+}
+
+internal class TRMethodCallTracePointPrettifier(
+    val point: TRMethodCallTracePoint,
+    val parameters: List<TRObject?>,
+    val methodDescriptor: MethodDescriptor,
+): TRTracePointPrettifier {
+    var className: String = point.obj?.adornedRepresentation() ?: methodDescriptor.className.substringAfterLast(".")
+    var methodName: String = methodDescriptor.methodName
```
eupp: Please avoid mutable properties. See my comment about visitor pattern.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrettifiers.kt
eupp: Even a bit better, you can share `className`/`methodName`/`varName` printing logic between different trace points via special printer for class/method/variable names:

```kotlin
interface DescriptorsNamePrinter {
    fun Appendable.appendClassName(clazz: ClassDescriptor)
    fun Appendable.appendMethodName(method: MethodDescriptor)
    fun Appendable.appendLocalVariableName(variable: VariableDescriptor)
    ...
}
```

eupp:
Please also note that for syntax highlight on the plugin side, `toText()` method is not used. Instead, a custom `toColoredText()` method is used (it depends on classes from IntelliJ API and thus cannot be extracted from the plugin). 

I am afraid it would not be possible to avoid the copy-paste there entirely. However, the solution with visitors might help to reduce it. At the very least, it would be possible to re-use `DescriptorsNamePrinter`.

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
dmitrii-artuhov: https://youtrack.jetbrains.com/agiles/153-5670/current?issue=DR-256

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    return result.toString()
+}
+
+/**
+ * Removes java lambdas runtime address from classname.
+ */
+fun removeJavaLambdaRuntimeAddress(classNameRepresentation: String): String {
+    if (!isJavaLambdaClass(classNameRepresentation)) return classNameRepresentation
+    return classNameRepresentation.substringBeforeLast('/')
```
eupp: Please re-check if this indeed removes the hash codes from java lambda class names. 
In the file `ImmutableListTest_empty.txt` there are still examples when it is left in the trace, for instance, on line 34:  `ImmutableListTest$$Lambda$100@1`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+
+package org.jetbrains.lincheck.trace
+
+import org.jetbrains.lincheck.descriptors.CodeLocations
+import org.jetbrains.lincheck.descriptors.FieldDescriptor
+import org.jetbrains.lincheck.descriptors.MethodDescriptor
+import org.jetbrains.lincheck.descriptors.VariableDescriptor
+
+
+interface TRAppendable {
```
eupp, line 19: Nice idea to support both regular text and colored text!

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+    fun appendClassName(prettyClassName: String, receiver: TRObject?): TRAppendable = append(prettyClassName)
+    fun appendMethodName(prettyMethodName: String, md: MethodDescriptor): TRAppendable = append(prettyMethodName)
```
eupp, line 21: Why `prettyMethodName` instead of just `methodName`? 
Also, lets be consistent in naming: if we have `methodName: String` then we should have `methodDescriptor: MethodDescriptor`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+package org.jetbrains.lincheck.trace
+
+import org.jetbrains.lincheck.descriptors.CodeLocations
+import org.jetbrains.lincheck.descriptors.FieldDescriptor
+import org.jetbrains.lincheck.descriptors.MethodDescriptor
+import org.jetbrains.lincheck.descriptors.VariableDescriptor
+
+
+interface TRAppendable {
+    fun appendClassName(prettyClassName: String, receiver: TRObject?): TRAppendable = append(prettyClassName)
```
eupp, line 20: Why this one does not take `classDescriptor` as an argument?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+    fun appendVariableName(prettyVariableName: String, vd: VariableDescriptor): TRAppendable = append(prettyVariableName)
+    fun appendArray(arr: TRObject): TRAppendable = append(arr.adornedRepresentation())
+    fun appendArrayIndex(index: Int): TRAppendable = append(index.toString())
+    fun appendObject(obj: TRObject?): TRAppendable = append(obj.toString())
+    fun appendKeyword(keyword: String): TRAppendable = append(keyword)
+    fun appendSpecialSymbol(symbol: String): TRAppendable = append(symbol)
+    fun append(text: String?): TRAppendable
+}
+
+internal class DefaultTRAppendable: TRAppendable {
```
eupp: You can make it `object` if it is singleton.
Also, let's rename to `DefaultTRTextAppendable`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+    private lateinit var destination: Appendable
```
eupp: Let's avoid `lateinit`. You can just declare it as: 
```kotlin
var destination: Appendable? = null
```

then in `append` add `check(destination != null) { ... }` at the beginning of the method. 

Also can add `fun reset() { destination = null }` for convenience.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+            append(" ")
+            append(tracePoint.exceptionClassName)
+        } else if (tracePoint.result != TR_OBJECT_VOID) {
+            append(": ")
+            appendObject(tracePoint.result)
+        }
+        return this
+    }
+
+    private fun String.removeCoroutinesCoreSuffix(): String = removeSuffix("\$kotlinx_coroutines_core")
```
eupp: Let's move `removeCoroutinesCoreSuffix` to `Utils.kt` and here instead declare:
```kotlin
private fun String.prettifyMethodName(): String = 
    removeSuffix(...)
   // can add more prettification here in the future
```

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+        return tracePoint.className.startsWith("kotlin.jvm.internal.Ref$") && tracePoint.name == "element"
+    }
+
+    private fun String.removeVolatileDollar(): String = removeSuffix("\$volatile\$FU")
+}
+
+internal class DefaultTRFieldTracePointPrinter: AbstractTRFieldTracePointPrinter() {
+    private val app = DefaultTRAppendable()
+
+    fun print(dest: Appendable, tracePoint: TRFieldTracePoint, verbose: Boolean) =
```
eupp: Why not `Appendable.append(tracePoint: TRFieldTracePoint, verbose: Boolean)` for consistency with other methods?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+        }
+}
+
+
+abstract class AbstractTRLocalVariableTracePointPrinter {
+
+    protected fun TRAppendable.append(tracePoint: TRLocalVariableTracePoint): TRAppendable {
+        val vd = tracePoint.variableDescriptor
+
+        appendVariableName(vd.name.removeInlineIV().removeDollarThis().removeLeadingDollar(), vd)
```
eupp: Same suggestions as for method names: let's add `prettifyVariableName()` method.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+        append(" ")
+        appendObject(tracePoint.value)
+        return this
+    }
+
+    private fun String.removeInlineIV(): String = removeSuffix("\$iv")
+
+    private fun String.removeDollarThis(): String = if (this == "\$this") this else removePrefix("\$this")
+
+    private fun String.removeLeadingDollar(): String = removePrefix("$")
```
eupp: Same suggestion as for method names: let's move to `Utils.kt`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+}
+
+internal class DefaultTRLocalVariableTracePointPrinter: AbstractTRLocalVariableTracePointPrinter() {
+    private val app = DefaultTRAppendable()
+
+    fun print(dest: Appendable, tracePoint: TRLocalVariableTracePoint, verbose: Boolean) =
```
eupp: Same here: why not `Appendable.append(...)` ?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+        append(" ")
+        appendObject(tracePoint.value)
+        return this
+    }
+}
+
+internal class DefaultTRArrayTracePointPrinter: AbstractTRArrayTracePointPrinter() {
+    private val app = DefaultTRAppendable()
+
+    fun print(dest: StringBuilder, tracePoint: TRArrayTracePoint, verbose: Boolean) {
```
eupp: Same here: why not `Appendable.append(...)` ?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+            append(": ")
+            appendObject(tracePoint.result)
+        }
+        return this
+    }
+
+    private fun String.removeCoroutinesCoreSuffix(): String = removeSuffix("\$kotlinx_coroutines_core")
+}
+
+internal class DefaultTRMethodCallTracePointPrinter(): AbstractTRMethodCallTracePointPrinter() {
```
eupp: Same here: can make it `object`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    private fun isLambdaCaptureSyntheticField(tracePoint: TRFieldTracePoint): Boolean {
+        return tracePoint.className.startsWith("kotlin.jvm.internal.Ref$") && tracePoint.name == "element"
+    }
+
+    private fun String.removeVolatileDollar(): String = removeSuffix("\$volatile\$FU")
+}
+
+internal class DefaultTRFieldTracePointPrinter: AbstractTRFieldTracePointPrinter() {
```
eupp: Same here: can make it `object`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    private fun String.removeInlineIV(): String = removeSuffix("\$iv")
+
+    private fun String.removeDollarThis(): String = if (this == "\$this") this else removePrefix("\$this")
+
+    private fun String.removeLeadingDollar(): String = removePrefix("$")
+}
+
+internal class DefaultTRLocalVariableTracePointPrinter: AbstractTRLocalVariableTracePointPrinter() {
```
eupp: Same here: can make it `object`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -0,0 +1,239 @@
...
+        appendSpecialSymbol("]")
+        append(" ")
+        appendSpecialSymbol(tracePoint.accessSymbol())
+        append(" ")
+        appendObject(tracePoint.value)
+        return this
+    }
+}
+
+internal class DefaultTRArrayTracePointPrinter: AbstractTRArrayTracePointPrinter() {
```
eupp: Same here: can make it `object`.

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
dmitrii-artuhov: this removes the java-lambda runtime address, which would be like this:
`ImmutableListTest$$Lambda$100/0x000000@1`, but in your example output is as expected. The hashcodes are appended by the `adornedRepresentation(...)` function separately

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
dmitrii-artuhov: because I also pass the `MethodDescriptor`/`FieldDescriptor`/.., which themselves contain the method/field names, but not those that the user of the `TRAppendable` is supposed to print. Thus, `pretty...` names are specified first and then the additional information

dmitrii-artuhov: as for naming I thought it would be over-expressive: the parameter name fully matches its type

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
dmitrii-artuhov: the `ClassDescriptor` contains only the original (non-pretty) class name, which there is no need for. What is needed in fact is the `receiver`

eupp: > ClassDescriptor  contains only the original (non-pretty) class name

This may change in the future. And for uniformity of the API it would be nice too. But okay.

common/src/main/org/jetbrains/lincheck/trace/Utils.kt
eupp: In `ImmutableListTest$$Lambda$100@1` string the `$100` part is still platform dependent.
But ok, let's fix in separate PR.

PR #764, opened 2025-07-17
Decrease memory consumption on trace load.

lev-serebryakov-jetbrains:
Three things takes a lot of memory:
 
1. Index of all method call trace points. For each method call trace point loader need to store two `Long` (start and end offsets) to provide quick lazy load. If there is millions of trace points, this could take gigabytes of memory.
2. Each loaded method call trace point has addresses (`Long`, again) of each child. When all millions of trace points are child of one, that could takes gigabytes of memory too.
3. Even unloaded children were stored as `null`s in parent.

 Implement staged indices for both trace points and children: after some limit (hard-coded for now) flush indices to disk into temporary files.

 Index of trace points themselves are sparse file where each pair of Longs are stored at fixed place, indexed by trace point Id. Method call trace point Ids goes not sequentially, so a lot of space wasted, but it provide very fast to create & to read index. All other schemes are much more complicated. This trades space for speed and simplicity of code. SQLIte creates file half size of my solution, but works ten times slower. I don't think it is worth it to implement something like BTree by hands for this task.

 Index of method call trace point children is dense, as it is simple list of `Long`'s.

 To store children list consisted mostly of `null`s, chunked list is implemented. It organized as list of fixed-size lists, and doesn't store empty chunks, but remember its virtual size and can provide `null` at any index which is not filled.

 All disk indices are implemented via memory-mapped files. It is quickest way of all I've tried (including SQLite).

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/AddressIndex.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/AddressIndex.kt
@@ -13,9 +13,9 @@ package org.jetbrains.lincheck.trace
 import java.nio.ByteBuffer
 import java.nio.channels.SeekableByteChannel
 import java.nio.file.Files
-import java.nio.file.StandardOpenOption.CREATE_NEW
 import java.nio.file.StandardOpenOption.DELETE_ON_CLOSE
 import java.nio.file.StandardOpenOption.READ
+import java.nio.file.StandardOpenOption.TRUNCATE_EXISTING
 import java.nio.file.StandardOpenOption.WRITE
 import kotlin.math.min
 import kotlin.math.roundToInt
@@ -25,10 +25,28 @@ internal sealed class AddressIndex {
         protected set
 
     abstract operator fun get(index: Int): Long
+
     abstract fun add(address: Long): AddressIndex
+    open fun finishWrite() {}
 
     internal companion object {
-        fun create(): AddressIndex = MemoryAddressIndex()
+        fun create(): AddressIndex = NullAddressIndex()
+    }
+}
+
+internal class NullAddressIndex: AddressIndex() {
+    override fun get(index: Int): Long = -1L
+
+    override fun add(address: Long): AddressIndex {
+        if (address < 0) {
+            size++
+            return this
+        }
+        var replacement: AddressIndex = MemoryAddressIndex()
+        repeat(size) {
+            replacement = replacement.add(-1)
+        }
+        return replacement.add(address)
     }
 }
 
@@ -44,6 +62,8 @@ internal class MemoryAddressIndex: AddressIndex() {
     }
 
     override fun add(address: Long): AddressIndex {
+        if (address < 0) return this
+
         if (size == MAX_MEM_INDEX_SIZE) {
             val fileIndex = FileAddressIndex(this.storage, size)
             fileIndex.add(address)
@@ -59,8 +79,7 @@ internal class MemoryAddressIndex: AddressIndex() {
 
 }
 
-private const val READ_BUFFER_SIZE = MAX_MEM_INDEX_SIZE / 2
-private const val WRITE_BUFFER_SIZE = MAX_MEM_INDEX_SIZE / 2
+private const val LONG_SIZE_SHIFT = 3 // log2(Long.SIZE_BYTES)
 
 internal class FileAddressIndex(
     copyFrom: LongArray,
@@ -68,54 +87,61 @@ internal class FileAddressIndex(
 ) : AddressIndex() {
     private val storage: SeekableByteChannel
 
-    private var writeBufferStart: Int = 0
-    private val writeBuffer: ByteBuffer = ByteBuffer.allocate(WRITE_BUFFER_SIZE * Long.SIZE_BYTES)
-
-    private var readBufferStart: Int = 0
-    private var readBufferCount: Int = 0
-    private val readBuffer: ByteBuffer = ByteBuffer.allocate(READ_BUFFER_SIZE * Long.SIZE_BYTES)
+    private var bufferStart: Int = 0
+    private var bufferCount: Int = 0
+    private val buffer: ByteBuffer = ByteBuffer.allocate(MAX_MEM_INDEX_SIZE shl LONG_SIZE_SHIFT)
+    private var writeFinished = false
 
     init {
-        val tmp = Files.createTempFile("trace-recorder-method-call-index", "idx")
-        storage = Files.newByteChannel(tmp, CREATE_NEW, DELETE_ON_CLOSE, READ, WRITE)
+        val tmp = Files.createTempFile("trace-recorder-method-call-index", ".idx")
+        storage = Files.newByteChannel(tmp, TRUNCATE_EXISTING, DELETE_ON_CLOSE, READ, WRITE)
 
         val prestored = ByteBuffer.allocate(count * Long.SIZE_BYTES)
         prestored.asLongBuffer().put(copyFrom, 0, count)
         storage.write(prestored)
 
-        writeBufferStart = count
+        bufferStart = count
         size = count
     }
 
     override fun get(index: Int): Long {
+        check(writeFinished) { "Cannot read from unfinished index" }
         require(index in 0 ..< size) { "Index: $index, Size: $size" }
-        if (index >= writeBufferStart) {
-            return writeBuffer.getLong(index - writeBufferStart)
+
+        if (index in bufferStart ..< bufferStart + bufferCount) {
+            return buffer.getLong((index - bufferStart) shl LONG_SIZE_SHIFT)
         }
 
-        if (index in readBufferStart ..< readBufferStart + readBufferCount) {
-            return readBuffer.getLong(index - readBufferStart)
-        }
-
-        readBuffer.clear()
-        readBufferStart = index * Long.SIZE_BYTES
-        storage.position(readBufferStart.toLong())
-        storage.read(readBuffer)
-        readBufferCount = readBuffer.position() / Long.SIZE_BYTES
-        readBuffer.flip()
-        return readBuffer.getLong(0)
+        buffer.clear()
+        bufferStart = index
+        storage.position(bufferStart.toLong() shl LONG_SIZE_SHIFT)
+        storage.read(buffer)
+        bufferCount = buffer.position() shr LONG_SIZE_SHIFT
+        buffer.flip()
+        return buffer.getLong(0)
     }
 
     override fun add(address: Long): AddressIndex {
-        if (writeBuffer.remaining() == 0) {
-            writeBuffer.flip()
-            storage.position((writeBufferStart * Long.SIZE_BYTES).toLong())
-            storage.write(writeBuffer)
-            writeBufferStart = size
-            writeBuffer.clear()
+        check(!writeFinished) { "Cannot write to finished index" }
+        if (buffer.remaining() == 0) {
+            buffer.flip()
+            // We don't need seek, as nothing could move position from last write
+            storage.write(buffer)
+            bufferStart = size
+            buffer.clear()
         }
-        writeBuffer.putLong(address)
+        buffer.putLong(address)
         size++
         return this
+    }
+
+    override fun finishWrite() {
+        if (writeFinished) return
+        buffer.flip()
+        storage.write(buffer)
+        buffer.clear()
+        bufferStart = 0
+        bufferCount = 0
+        writeFinished = true
     }
 }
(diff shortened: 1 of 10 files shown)


trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+
+        return true
+    }
+
+    fun clear() {
+        chunks.clear()
+        totalSize = 0
+    }
+
+    fun forgetAll() {
```
eupp, line 51: What is the difference wrt. `clear()`?

Currently it is used only in `TRMethodCallTracePoint::unloadAllChildren` which is itself unused.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+        chunks.clear()
+    }
+
+    fun forget(index: Int): Unit = set(index, null)
```
eupp, line 58: Unused method. Do we need it?

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    fun forget(from: Int, to: Int) {
```
eupp, line 63: Unused method. Do we need it?

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+            chunks[chunk] = ArrayList()
+        }
+        val ch = chunks[chunk]!!
+        while (idx >= ch.size) {
+            ch.add(null)
+        }
+        return ch
+    }
+
+    private fun forgetInOneChunk(chunk: Int, fromIdx: Int, toIdx: Int) {
```
eupp: Would make sense to move this method near other `forget` methods. Also, naming it just `forget` as other methods would be fine --- it is clear from the signature that elements are forgotten in the single chunk.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    private fun cleanupTail() {
+        while (chunks.isNotEmpty() && chunks.last() == null) {
+            chunks.removeLast()
+        }
+    }
+
+    private inner class ChunkedListIterator(
+        var idx: Int
+    ): ListIterator<T?>
```
eupp:
```suggestion
    ) : ListIterator<T?> {
```

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+    {
+        override fun next(): T? {
+            if (idx !in 0..<totalSize)
+                throw NoSuchElementException("Index: $idx, Size: $totalSize")
+            return get(idx++)
+        }
+
+        override fun hasNext(): Boolean = idx < totalSize
+
+        override fun hasPrevious(): Boolean  = idx > 0
```
eupp:
```suggestion
        override fun hasPrevious(): Boolean = idx > 0
```

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+        val lastEmptyChunk = if (toIdx + 1 >= (chunks[toChunk]?.size ?: 0)) toChunk else toChunk - 1
+
+        for (chunk in firstEmptyChunk .. lastEmptyChunk) {
+            chunks[chunk] = null
+        }
+
+        forgetInOneChunk(fromChunk, fromIdx, if (fromChunk == toChunk) toIdx else CHUNK_SIZE)
+        if (fromChunk != toChunk) {
+            forgetInOneChunk(toChunk, 0, toIdx)
+        }
```
eupp, line 87: IMO, would be cleaner to just split two cases by `if`:
```kotlin
if (fromChunk == toChunk) {
    ...
} else {
    ...
}
```

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    fun forget(index: Int): Unit = set(index, null)
+
+    fun forget(from: Int, to: Int) {
+        checkRange(from)
+        checkRange(to - 1)
+
+        val fromChunk = from shr CHUNK_SHIFT
+        val fromIdx = from and CHUNK_MASK
```
eupp, line 68: This pattern is repeated constantly. What about extracting chunk and index calculation into separate functions (in case the implementation will change in future)?

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
```diff
@@ -0,0 +1,231 @@
...
+
+package org.jetbrains.lincheck.trace
+
+import kotlin.math.min
+
+private const val CHUNK_SIZE = 1024
+private const val CHUNK_SHIFT = 10 // log2(CHUNK_SIZE)
+private const val CHUNK_MASK = (1 shl CHUNK_SHIFT) - 1
+
+internal class ChunkedList<T>: List<T?>, RandomAccess {
```
eupp, line 19: Why it implements `List<T?>` rather than `List<T>` ?
Besides, shouldn't it also implement `MutableList` ?

trace/src/main/org/jetbrains/lincheck/trace/RangeIndex.kt
```diff
@@ -0,0 +1,131 @@
...
+        return map[id]
+    }
+
+    override fun addRange(id: Int, start: Long, end: Long): RangeIndex {
+        check(!closed) { "Index already closed." }
+        require(id >= 0) { "Id $id must be non-negative" }
+        require(start >= 0) { "Id $id: Start $start must be non-negative" }
+        require(end >= start) { "Id $id: End $end must be larger or equal than start $start" }
+
+        if (map.size >= MAX_HASHMAP_SIZE) {
```
eupp: What if instead, you declare another `RangeIndex` implementation:
```kotlin
private class DynamicRangeIndex: RangeIndex(...) {
    val indexImpl : RangeIndex = ... 
    
    override fun addRange(id: Int, start: Long, end: Long) {
    
        if (indexImpl is HashMapRangeIndex && indexImpl.map.size >= MAX_HASHMAP_SIZE) {
            // replace `indexImpl` to memmap based
            ...
        }
    }
}
```

This approach allows to:
- split the range implementation from the dynamic switch policy;
- simplifies the `RangeIndex` class API -- no need to return new index from `addRange` method.

trace/src/main/org/jetbrains/lincheck/trace/RangeIndex.kt
```diff
@@ -0,0 +1,131 @@
...
+package org.jetbrains.lincheck.trace
+
+import java.nio.file.Files
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+internal data class Range(val start: Long, val end: Long)
+
+internal sealed class RangeIndex(
+    aOpenRanges: MutableMap<Int, Long>,
```
eupp: Why not declaring property directly in the constructor?:
```kotlin
protected val openRanges: MutableMap<Int, Long>
```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -208,7 +209,10 @@ class LazyTraceReader(
     }
 
     fun readRoots(): List<TRTracePoint> {
+        // var start = System.currentTimeMillis()
         loadContext()
+        // System.err.println("Context loaded in ${System.currentTimeMillis() - start} ms")
+        // start = System.currentTimeMillis()
```
eupp: Please remove left-over debug print code (or replace with logging)

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -238,14 +242,16 @@ class LazyTraceReader(
                 roots[threadId] = tracepoints.first()
             }
         }
+        // System.err.println("Roots loaded ${System.currentTimeMillis() - start} ms")
```
eupp: Please remove left-over debug print code (or replace with logging)

trace/src/main/org/jetbrains/lincheck/trace/AddressIndex.kt
```diff
@@ -0,0 +1,123 @@
...
+    override fun get(index: Int): Long {
+        if (index !in 0..<size) throw IndexOutOfBoundsException("Index: $index, Size: $size")
+        return storage.get(index)
+    }
+
+    override fun add(address: Long): AddressIndex {
+        if (address < 0) return this
+
+        if (size == MAX_MEM_INDEX_SIZE) {
+            val fileIndex = MemoryMapAddressIndex(this.storage, size)
```
eupp: Same suggestion as for `RangeIndex`. 
Also, here it would help to get rid of separate class `NullAddressIndex` as you can store `indexImpl: AddressIndex?`

eupp (changes requested):
Please also don't forget to update `trace/README.md` to accomodate for the changes in the serialization format. Otherwise it would be hard for the rest of us to keep track on the current status of serialization format.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: `clear()`  makes list clear (no elements). `forgetAll()` unload content but list is still «contains» elements, only as `null`s.

I add comment.

About unusage: I prefer to have full API from the beginning. Now plugin doen't unload children in closed branches, but it could be useful in future.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: Again, fullness of API.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: The same.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: Do we prefer group methods by usage or by access (public/protected/private/internal)?

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: Because it can return `null`, it is why it is exist!

It SHOULD NOT implement MutableList, as all operations of ML is not feasible for this list. It allows only adding to end and replacing existing element. No other operations of MutableList is supported on purpose.

trace/src/main/org/jetbrains/lincheck/trace/RangeIndex.kt
lev-serebryakov-jetbrains: But it adds one level of inderection! :-) Ok, I'll do this.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
lev-serebryakov-jetbrains: I don't like idea to construct `Pair<>` only to return two numbers: it is a lot of boxing/unboxing.

lev-serebryakov-jetbrains:
> Please also don't forget to update `trace/README.md` to accomodate for the changes in the serialization format. Otherwise it would be hard for the rest of us to keep track on the current status of serialization format.

Serialization format is the same. Only transient data which is created only on lazy load is changed.

trace/src/main/org/jetbrains/lincheck/trace/ChunkedList.kt
eupp: Kotlin style guide recommends to order methods logically:
https://kotlinlang.org/docs/coding-conventions.html#class-layout

PR #768, opened 2025-07-23
Trace polishing: add trace preprocessor for node combining

dmitrii-artuhov:
Missing trace compressors:
- `compressSuspendImpl`: we don't have multithreading so we see no `$suspendImpl` method suffixes, because `runBlocking` executes code in another thread
- `compressUserThreadRun`: no multithreading
- `compressThreadStart`: we do not collect trace from kotlin standard library, so we see no children to combine with


lev-serebryakov-jetbrains (approved):
LGTM!

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -32,10 +32,14 @@ private val EVENT_ID_GENERATOR = AtomicInteger(0)
...
 
 sealed class TRTracePoint(
-    val codeLocationId: Int,
+    codeLocationId: Int,
     val threadId: Int,
     val eventId: Int
 ) {
+    private var _codeLocationId: Int = codeLocationId
+    val codeLocationId: Int
+        get() = _codeLocationId
```
eupp:
```kotlin
var codeLocationId: Int = codeLocationId
    private set
```

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.lincheck.trace
+
+
+interface TracePostprocessor {
+    fun postprocess(reader: LazyTraceReader, tracePoint: TRMethodCallTracePoint): TRTracePoint
+}
```
eupp, line 26: I have few questions about this interface:

- If the `postrprocess` method takes `TRMethodCallTracePoint`, shouldn't this class be called `TRMethodCallTracePointPostprocessor` ?

- Or otherwise, should we add `postprocess` methods for each type of trace points?  

- Also, why it takes `TRMethodCallTracePoint` as argument, but returns just `TRTracePoint`?
  Maybe then it should take `TRTracePoint` as argument too?
  In the implementation class, you can always check for the required subclass, and return `this` earlier if it does not fit.

eupp, line 26: Also, could you please add a short KDoc for this interface and `postprocess` method.

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+     * will be collapsed into:
+     *
+     * ```
+     * A.callMe(3, "Hey") at A.operation(A.kt:23)
+     * ```
+     *
+     */
+    private fun TRMethodCallTracePoint.compressDefaultPairs(reader: LazyTraceReader): TRMethodCallTracePoint {
+        if (events.size != 1) return this
+        val singleChild = loadChild(reader, tracePoint = this, 0)
```
eupp:
```suggestion
        val singleChild = loadChild(reader, tracePoint = this, 0) ?: return this
```

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+     * ```
+     *
+     */
+    private fun TRMethodCallTracePoint.compressAccessPairs(reader: LazyTraceReader): TRMethodCallTracePoint {
+        if (events.size != 1) return this
+        val singleChild = loadChild(reader, tracePoint = this, 0)
+        if (singleChild == null) return this
+
+        if (singleChild !is TRMethodCallTracePoint) return this
+        if (!haveSameOwner(className, singleChild.className) || !isAccessPair(methodName, singleChild.methodName)) return this
```
eupp: Line is too long

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+     * will be collapsed into:
+     *
+     * ```
+     * A.callMe() at A.operation(A.kt:N)
+     * ```
+     *
+     */
+    private fun TRMethodCallTracePoint.compressAccessPairs(reader: LazyTraceReader): TRMethodCallTracePoint {
+        if (events.size != 1) return this
+        val singleChild = loadChild(reader, tracePoint = this, 0)
```
eupp:
```suggestion
val singleChild = loadChild(reader, tracePoint = this, 0) ?: return this
```

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+     *
+     */
+    private fun TRMethodCallTracePoint.compressSyntheticFieldAccess(reader: LazyTraceReader): TRTracePoint {
+        if (events.size != 2 && events.size != 3) return this
+        if (!isSyntheticFieldAccess(methodName)) return this
+
+        // get and set method versions have different number of children (2 and 3, correspondingly)
+        // the correct access node is stored at the last position
+        val actualNode = if (isSyntheticGetFieldAccess(methodName)) loadChild(reader, tracePoint = this, 1) ?: return this
+                         else loadChild(reader, tracePoint = this, 2) ?: return this
```
eupp: Please reformat this code fragment, it is hard to understand in the current form. 

- First line is too long.

- Using `?: return this` here clutters the code, it is unclear whether it is related to `else` branch, or the whole `if-then-else` expression.

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        actualNode.replaceCodeLocationId(codeLocationId)
```
eupp: Why do you need to change `codeLocationId` ?

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    private fun loadChild(reader: LazyTraceReader, tracePoint: TRMethodCallTracePoint, childIdx: Int): TRTracePoint? {
+        if (childIdx !in 0..tracePoint.events.lastIndex) return null
+        return tracePoint.events[childIdx] ?: reader.getChildAndRestorePosition(tracePoint, childIdx)
+    }
+
+    private fun combineCallNodes(parent: TRMethodCallTracePoint, child: TRMethodCallTracePoint): TRMethodCallTracePoint {
+        val newNode = TRMethodCallTracePoint(
+            child.threadId,
+            parent.codeLocationId,
```
eupp, line 173: Why code location is taken from parent, but all other properties from `child`?

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
eupp: Btw, wouldn't it be better to create new trace point, similarly as it is done in `combineCallNodes` ?

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * This is different from `fun$access`, which is addressed in [compressAccessPairs].
+     *
+     */
+    private fun TRMethodCallTracePoint.compressSyntheticFieldAccess(reader: LazyTraceReader): TRTracePoint {
+        if (events.size != 2 && events.size != 3) return this
+        if (!isSyntheticFieldAccess(methodName)) return this
+
+        // get and set method versions have different number of children (2 and 3, correspondingly)
+        // the correct access node is stored at the last position
```
eupp, line 145: How these children look like? Could you please add an example from the trace to the comment here.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -169,19 +169,28 @@ private class CodeLocationsContext {
...
             this(
                 dataFileName = baseFileName,
-                index = wrapStream(openExistingFile(baseFileName + INDEX_FILENAME_SUFFIX))
+                index = wrapStream(openExistingFile(baseFileName + INDEX_FILENAME_SUFFIX)),
+                postprocessor = postprocessor
             )
 
+    private fun readTracePointWithPostprocessor(readCallback: () -> TRTracePoint): TRTracePoint =
+        readCallback().let { tracePoint ->
+            if (tracePoint is TRMethodCallTracePoint)
```
eupp: See the comment about `TracePostprocessor` interface. 
If we change its API to accept just `tracePoint`, this method implementation will become almost trivial: 

```kotlin
postprocessor.postprocess(
    reader = this@LazeTraceReader,
    tracePoint = readCallback(),
)
```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -257,7 +266,7 @@ class LazyTraceReader(
         loadTracePoints(
             threadId = parent.threadId,
             maxRead = Integer.MAX_VALUE,
-            reader = this::readTracePointWithChildAddresses,
+            reader = { readTracePointWithPostprocessor(this::readTracePointWithChildAddresses) },
```
eupp: BTW, why not just pass it as `this::readTracePointWithPostprocessor` and remove lambda argument from `readTracePointWithPostprocessor` (it looks like you always pass the same read callback anyway).

common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
```diff
@@ -72,4 +72,32 @@ fun String.removeLeadingDollar(): String = removePrefix("$")
...
 
-fun String.removeVolatileDollarFU(): String = removeSuffix("\$volatile\$FU")
\ No newline at end of file
+fun String.removeVolatileDollarFU(): String = removeSuffix("\$volatile\$FU")
+
+/**
+ * Used for compressing default pairs and access pairs.
+ */
+fun haveSameOwner(currentClassName: String, nextClassName: String): Boolean =
+    currentClassName == nextClassName
```
eupp: After refactoring, having `haveSameOwner` as separate method kinda makes no sense, because it is just equality comparison.

trace/README.md
```diff
@@ -334,18 +334,18 @@ seekable stream to support random access to data.
...
+ After that the same instance of lazy loader can be used to load any children of any
 already loaded method call tracepoint. It is possible to load all children at once
 or load only some range of children.
 
  Method call trace point with children which were not yet loaded will have `null`
 at corresponding positions in the children list (`events` property). This list
 will have a proper fixed number of elements from the beginning.
 
- Also, method call trace point which is loaded in such way can «forget» any children
+ Also, method call trace point which is loaded in such a way can «forget» any children
```
eupp, line 348: Thanks for fixing typos and grammar!

eupp (changes requested):

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
dmitrii-artuhov: In order for clicking in the trace sidebar to work and point to the correct code location (open correct file and select correct line of code)

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
dmitrii-artuhov: for proper clicking in the trace sidebar

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
eupp: Please add a comment then.

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
eupp: Please add a comment then.

trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
dmitrii-artuhov: Changed its API to accept and return `TRTracePoint`, the checking for correct type point is now part of each of the compression method

eupp (approved):

PR #773, opened 2025-07-24
Refactor `TRAppendable::appendClassName` method not to include `receiver` object

dmitrii-artuhov:


common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
```diff
@@ -56,6 +56,24 @@ fun removeJavaLambdaRuntimeAddress(classNameRepresentation: String): String {
...
 }
 
+/**
+ * Removes outer class name for primitive types when they are wrapped with `Ref` class.
+ * E.g. `kotlin.jvm.internal.Ref$IntRef@xxxxx` -> `IntRef@xxxxx`.
+ *
+ * *Note*: should be called on fully qualified class name.
+ */
+fun removeRefWrapperClassName(classNameRepresentation: String): String {
+    val prefix = "kotlin.jvm.internal.Ref$"
```
eupp: Maybe left only `Ref$` part, to lift the restriction of this function to be called on fqn?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -75,8 +75,12 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
     }
 
     protected fun TRAppendable.appendClassName(tracePoint: TRMethodCallTracePoint, methodDescriptor: MethodDescriptor): TRAppendable {
```
eupp: Let's rename to `appendReceiver` or `appendOwner`. 
Because that's what this function actually does: prints receiver --- either object itself, or the class name in case of static field/method.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -144,13 +148,12 @@ abstract class AbstractTRFieldTracePointPrinter {
         return this
     }
 
-    protected fun TRAppendable.appendClassName(tracePoint: TRFieldTracePoint, isLambdaCaptureSyntheticField: Boolean): TRAppendable {
-        val className = tracePoint.obj?.adornedRepresentation() ?: tracePoint.fieldDescriptor.className.substringAfterLast(".")
-        if (!isLambdaCaptureSyntheticField) {
-            appendClassName(className.removePrefix("Ref$"), tracePoint.obj)
+    protected fun TRAppendable.appendClassName(tracePoint: TRFieldTracePoint): TRAppendable {
```
eupp: Let's rename to `appendReceiver` or `appendOwner`.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -75,8 +75,12 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
     }
 
     protected fun TRAppendable.appendClassName(tracePoint: TRMethodCallTracePoint, methodDescriptor: MethodDescriptor): TRAppendable {
-        val className = tracePoint.obj?.adornedRepresentation() ?: methodDescriptor.className.substringAfterLast(".")
-        appendClassName(className, tracePoint.obj)
+        if (tracePoint.obj != null) {
+            appendObject(tracePoint.obj)
+        }
+        else {
+            appendClassName(methodDescriptor.className.substringAfterLast("."))
```
eupp: Shouldn't you call `adornedClassNameRepresentation` here?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -144,13 +148,12 @@ abstract class AbstractTRFieldTracePointPrinter {
...
-        val className = tracePoint.obj?.adornedRepresentation() ?: tracePoint.fieldDescriptor.className.substringAfterLast(".")
-        if (!isLambdaCaptureSyntheticField) {
-            appendClassName(className.removePrefix("Ref$"), tracePoint.obj)
+    protected fun TRAppendable.appendClassName(tracePoint: TRFieldTracePoint): TRAppendable {
+        if (tracePoint.obj != null) {
+            appendObject(tracePoint.obj)
         }
         else {
-            appendClassName(className, tracePoint.obj)
+            appendClassName(tracePoint.fieldDescriptor.className.substringAfterLast("."))
```
eupp: Shouldn't you call `adornedClassNameRepresentation` here?

common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
dmitrii-artuhov: User's wrapper classes, named `Ref`, will be removed from class name then, which we do not want, I think

PR #777, opened 2025-07-25
Owner name analyzer

eupp:
This PR implements static ASM-based bytecode analyzer for field access and method call owner/receiver name resolution (for trace prettifying). 

Remaining tasks:
- ~~Proper integration with trace polishing (class/method/field/variable name polishing)~~ (done).
- ~~Serialization and de-serialization of owner names for binary trace format~~ (will be done in separate PR).


common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -95,33 +97,41 @@ class TraceContext {
...
         return id
     }
 
     fun stackTrace(codeLocationId: Int): StackTraceElement {
-        // actors do not have a code location (for now)
         if (codeLocationId == UNKNOWN_CODE_LOCATION_ID) return EMPTY_STACK_TRACE
-        var loc = locations[codeLocationId]
+        val loc = locations[codeLocationId]
+        if (loc == null) {
+            error("Unknown code location $codeLocationId")
```
dmitrii-artuhov: nit: `Code location does not exist`? Unknown code location is a valid state and processed above

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -95,33 +97,41 @@ class TraceContext {
...
+        }
+        return loc.stackTraceElement
+    }
+
+    fun accessPath(codeLocationId: Int): AccessPath? {
+        if (codeLocationId == UNKNOWN_CODE_LOCATION_ID) return null
+        val loc = locations[codeLocationId]
         if (loc == null) {
             error("Unknown code location $codeLocationId")
```
dmitrii-artuhov: Same here

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariablesTracker.kt
```diff
@@ -21,7 +21,7 @@ internal class LocalVariablesTracker(
     val locals: MethodVariables
 ) : MethodVisitor(ASM_API, visitor) {
     override fun visitLabel(label: Label)  {
-        super.visitLabel(label)
         locals.visitLabel(label)
+        super.visitLabel(label)
```
dmitrii-artuhov: Is there a reason why this reordering was done? In one of the old PR, we had a discussion that calls to super methods of ASM usually performed first, to essentially "show" that `visitLabel` method is being invoked and its execution "started"

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
```diff
@@ -0,0 +1,780 @@
...
+package org.jetbrains.kotlinx.lincheck.transformation
+
+import org.jetbrains.lincheck.descriptors.*
+import org.objectweb.asm.*
+import org.objectweb.asm.commons.InstructionAdapter.OBJECT_TYPE
+import kotlin.math.max
+
+/**
+ * TODO: update the documentation
+ */
```
dmitrii-artuhov: Please update

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
```diff
@@ -0,0 +1,780 @@
...
+        owner: String?,
+        name: String,
+        descriptor: String,
+        isInterface: Boolean
+    ) {
+        if (api < Opcodes.ASM5 && (opcodeAndSource and Opcodes.SOURCE_DEPRECATED) == 0) {
+            // Redirect the call to the deprecated version of this method.
+            super.visitMethodInsn(opcodeAndSource, owner, name, descriptor, isInterface)
+            return
+        }
```
dmitrii-artuhov, line 185: I suppose this code was taken from the `AnalyzerAdapter`? In our case, we never specify `api` other than `ASM9`. And here you account for the old API in a way that does not update the `locals` and `stack` fields of the class. But in other methods, you always do, so even if an old API is provided, after invocation of a method you will still have its args and receiver in your manually updated `stack`

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassFileTransformer.kt
```diff
@@ -135,7 +135,9 @@ object LincheckClassFileTransformer : ClassFileTransformer {
 
                         val index = local.index
                         val type = Type.getType(local.desc)
-                        val info = LocalVariableInfo(local.name, local.index, local.start.label to local.end.label, type)
+                        val info = LocalVariableInfo(local.name, local.index, type,
```
dmitrii-artuhov, line 142: in this the method return type, please use defined alias `LocalVariablesMap`

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
```diff
@@ -0,0 +1,780 @@
...
+        } else if (firstDescriptorChar == 'J' || firstDescriptorChar == 'D') {
+            pop(2)
+        } else {
+            pop(1)
+        }
+    }
+
+    private fun execute(
+        opcode: Int,
+        intArg: Int,
```
dmitrii-artuhov, line 385: `argIndex`?

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
```diff
@@ -0,0 +1,113 @@
...
+                }
+                is ArrayElementByNameAccess -> {
+                    with(builder) {
+                        append('[')
+                        append(location.indexAccessPath)
+                        append(']')
+                    }
+                }
+                is ArrayLengthAccess -> {
+                    builder.append(".length")
```
dmitrii-artuhov: Why isn't it `ObjectFieldAccess`?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SharedMemoryAccessTransformer.kt
```diff
@@ -234,10 +239,11 @@ internal class SharedMemoryAccessTransformer(
         val arrayElementType = getArrayElementType(opcode)
         val indexLocal = newLocal(INT_TYPE).also { storeLocal(it) }
         val arrayLocal = newLocal(getType("[$arrayElementType")).also { storeLocal(it) }
+        val ownerName = ownerNameAnalyzer?.stack?.get(1)
```
dmitrii-artuhov: Why `ownerName` is not extracted for the array store? As we discussed previously, reads will be removed from the trace. By the way, will be array elements reads removed as well? In case if they will, getting owner name for array loading will not be shown in the trace.

Assume you have a code:
```kotlin
val arr = intArrayOf(1, 2)
val local = arr[1]
arr[1] = -1
```

right now, it is printed in the trace like this:
```
  [I@455888635[0] = 1 at TRPlayground.kt:53
  [I@455888635[1] = 2 at TRPlayground.kt:53
  arr ➜ [I@455888635 at TRPlayground.kt:54
  [I@455888635[1] ➜ 2 at TRPlayground.kt:54
  arr ➜ [I@455888635 at TRPlayground.kt:55
  [I@455888635[1] = -1 at TRPlayground.kt:55
```

when reads will be removed:
```
  [I@455888635[0] = 1 at TRPlayground.kt:53
  [I@455888635[1] = 2 at TRPlayground.kt:53
  [I@455888635[1] = -1 at TRPlayground.kt:55
```
no name at all

dmitrii-artuhov: when we will remove reads from the trace, I think it makes sense to calculate the write-access owner names as well, since we will not have such pairs anymore:
```
objName ➜ Obj@XXXX
Obj@XXXX.field = value

arrName ➜ [Arr@XXXX
[Arr@XXXX[index] = value
```

dmitrii-artuhov:
In the code https://github.com/JetBrains/lincheck/pull/777/files#diff-c44d39a4e70ca4a7b63230124288a86684421139e3ff74d3f45988b5f172e0b9R107-R108
where Trace Recorder transformers are applied, we are missing the 
```kotlin
if (shouldWrapInIgnoredSection(className, methodName, desc)) {
  mv = IgnoredSectionWrapperTransformer(fileName, className, methodName, mv, mv)
  return mv
}
```

I have faced bugs when running examples on this branch with debugger

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
```diff
@@ -0,0 +1,780 @@
...
+                pop()
+            }
+
+            /* Array access instructions */
+            Opcodes.AALOAD,
+            Opcodes.IALOAD, Opcodes.LALOAD, Opcodes.BALOAD, Opcodes.CALOAD, Opcodes.SALOAD,
+            Opcodes.FALOAD, Opcodes.DALOAD -> {
+                val indexName = pop()
+                val arrayName = pop()
+                val arrayAccess = indexName?.let { ArrayElementByNameAccess(it) }
```
dmitrii-artuhov: When I see array element accesses in the trace, that still look as before: the index in the brackets is a just a number no matter how I specify it (as local variable, field, or pure number). Of course, there is an instruction that loads the index onto the stack (the read access), but the index itself is a number when array access is printed.
Is this `ArrayElementByNameAccess` class expected to provide the name of the local variable/field when it contains the element index (when reads will be removed from trace completely)? Otherwise, I see no difference in the output

dmitrii-artuhov (dismissed):
Also it would be nice to add some integration-tests from the already downloaded projects or from the new ones. Also we have no test which would show the array element access, it would be nice to add as well

eupp:
@dmitrii-artuhov thanks for the review!

Regarding:
> where Trace Recorder transformers are applied, we are missing the `IgnoredSectionWrapperTransformer`

Let's fix it in a separate small PR, since it is an orthogonal problem. 

Same about the tests. Let's add them in another PR, then I'll be happy to rebase this PR and check it on new tests. 
BTW, I think we had YT tickets for adding more integration tests for TR. Should it be done as a part of this initiative?

The rest of your comments I'll fix or answer.

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
eupp: Was there before this PR. Changed to `error("Invalid code location id $codeLocationId")`.

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
eupp: Was there before this PR. Changed to `error("Invalid code location id $codeLocationId")`.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariablesTracker.kt
eupp: Yes, I think here the order is important. 
If we call `locals.visitLabel(label)` first, it will update the set of active local variables, and all subsequent transformers in the chain will observe this updated set in their `visitLabel` overrides. 

For `OwnerNameAnalyzerAdapter` it is important, because it uses the active local variables info to update the state of locals:
```
override fun visitLabel(label: Label?) {
    super.visitLabel(label)
    val activeVars = methodVariables.getActiveVars()
    setActiveLocalVariableNames(activeVars)
}
```

Have we swapped the order of calls, then `OwnerNameAnalyzerAdapter`  would have observed the set of active local variables **before** the current label, and not at the point of the current label.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
eupp: > I suppose this code was taken from the AnalyzerAdapter?

Yeap, it was taken from `AnalyzerAdapter`. 
And I would prefer to keep it as is (unless we have some strong arguments why we need to change it), because I have not really dived deep into what it does .

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
eupp: The name was taken from `AnalyzerAdapter`. 

I think it has the following logic. `execute` is called from different visitor methods (different types of JVM instructions really), and "executes" these instructions according to analyzer semantics.
Different visitor methods in ASM `MethodVisitor` have different arguments.
The `intArg` is used to pass "integer argument" of a visitor (e.g., `opcode` in case of `visitIntInsn`), and `0` is passed in case when visitor does not have corresponding "integer argument".

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
eupp: We could have re-use `ObjectFieldAccess` indeed. 

Here, I was not sure what approach is better. 
Array length accesses are kinda special in JVM, there is even a special instruction for them:
https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.arraylength 
So I just followed the JVM here. 

However, using `ObjectFieldAccess` makes sense too. 
I do not have a strong opinion here, what do you prefer?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
eupp: > Is this ArrayElementByNameAccess class expected to provide the name of the local variable/field when it contains the element index (when reads will be removed from trace completely)? Otherwise, I see no difference in the output

Yeap, that was the intention. 

Really, I just followed the semantics and types here. When I pop `index` from the stack, I get arbitrary `OwnerName` back (because that is what we store in stack/locals of our analyzer). So I have to save it as `OwnerName` still, because that is the only information that I have statically. The real integer index will be known only at runtime (unless it is some constant). 

I think we don't see it in the trace, because in practice `indexName` is always `null` . Therefore, `(arrayAccess != null && arrayName != null)` check always fail and we fallback to `push(null)`.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/SharedMemoryAccessTransformer.kt
eupp: > Why ownerName is not extracted for the array store? 

Ohh, sorry, I think I just forgot to handle this case. Will fix.

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
dmitrii-artuhov: I think, how `length` field of array is stored in memory, is irrelevant for us. I would decrease the complexity and use `ObjectFieldAccess` instead

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
eupp: > I think we don't see it in the trace, because in practice indexName is always null . Therefore, (arrayAccess != null && arrayName != null) check always fail and we fallback to push(null).

Nevermind, I actually found couple of other bugs, and now a lot more owner names are printed.

dmitrii-artuhov:
> `where Trace Recorder transformers are applied, we are missing the IgnoredSectionWrapperTransformer`
> Let's fix it in a separate small PR, since it is an orthogonal problem.

@eupp https://github.com/JetBrains/lincheck/pull/794

integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx.collections.immutable/tests.contract.list.ImmutableListTest_empty.txt
```diff
@@ -3,642 +3,642 @@ ImmutableListTest.empty() at :0
...
                   AssertionsKt.assertEquals(false, false, "") at ComparisonDSL.kt:21
               CompareContext.propertyEquals$default(CompareContext@1, null, CollectionBehaviorsKt$$Lambda@1, 1, null) at CollectionBehaviors.kt:111
                 CompareContext@1.propertyEquals("", CollectionBehaviorsKt$$Lambda@1) at ComparisonDSL.kt:20
-                  CompareContext@1.expected ➜ EmptyList@1 at ComparisonDSL.kt:21
-                  CollectionBehaviorsKt$$Lambda@1.invoke(EmptyList@1): true at ComparisonDSL.kt:21
-                    EmptyList@1.containsAll(EmptyList@1): true at CollectionBehaviors.kt:111
-                  CompareContext@1.actual ➜ SmallPersistentVector@1 at ComparisonDSL.kt:21
-                  CollectionBehaviorsKt$$Lambda@1.invoke(SmallPersistentVector@1): true at ComparisonDSL.kt:21
-                    SmallPersistentVector@1.containsAll(SmallPersistentVector@1): true at CollectionBehaviors.kt:111
+                  this.expected ➜ EmptyList@1 at ComparisonDSL.kt:21
```
ndkoval: Should we avoid `this` as an owner name?

integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx.collections.immutable/tests.contract.list.ImmutableListTest_empty.txt
```diff
@@ -3,642 +3,642 @@ ImmutableListTest.empty() at :0
...
                         CompareContext.assertFailEquals$default(CompareContext@1, CompareContext$$Lambda@1, CompareContext$$Lambda@1, null, 4, null) at ComparisonDSL.kt:23
                           CompareContext@1.assertFailEquals(CompareContext$$Lambda@1, CompareContext$$Lambda@1, null) at ComparisonDSL.kt:30
                             Result.Companion ➜ Result.Companion@1 at ComparisonDSL.kt:31
-                            CompareContext$$Lambda@1.invoke(): threw java.util.NoSuchElementException at ComparisonDSL.kt:31
-                              CompareContext@1.expected ➜ EmptyIterator@1 at ComparisonDSL.kt:23
-                              CollectionBehaviorsKt$$Lambda@1.invoke(EmptyIterator@1): threw java.util.NoSuchElementException at ComparisonDSL.kt:23
-                                EmptyIterator@1.next(): threw java.util.NoSuchElementException at CollectionBehaviors.kt:61
+                            expected.invoke(): threw java.util.NoSuchElementException at ComparisonDSL.kt:31
+                              this$0.expected ➜ EmptyIterator@1 at ComparisonDSL.kt:23
```
ndkoval: Should we avoid `this$<N>` as an owner name?

src/jvm/test/resources/expected_logs/suspend_trace_reporting_trace_debugger.txt
```diff
@@ -12,10 +12,10 @@ The following interleaving leads to the error:
 |                                                                          | bar(): -1                                       |
 |                                                                          |   ResultKt.throwOnFailure(null)                 |
 |                                                                          |   barStarted = true                             |
-|                                                                          |   lock.lock(null) [suspendable: Continuation#1] |
```
ndkoval, line 15: It should remain `lock.lock(..)` and `lock.unlock()` (see below)

src/jvm/test/resources/expected_logs/single_suspension_point_trace_debugger.txt
```diff
@@ -20,7 +20,7 @@ The following interleaving leads to the error:
 |                                    |     counter ➜ AtomicInteger#1                               |
 |                                    |     counter.get(): 1                                        |
 |                                    |     continuation1 = Continuation#3                          |
-|                                    |     it.getOrThrow(): COROUTINE_SUSPENDED                    |
+|                                    |     continuation1.getOrThrow(): COROUTINE_SUSPENDED         |
```
ndkoval, line 23: It should be neither `it` nor `continuation1`. Let's create an issue for this bug and hope it will get fixed automatically with #782

src/jvm/test/resources/expected_logs/run_concurrent_test/livelock.txt
```diff
@@ -7,15 +7,15 @@ The following interleaving leads to the error:
 | $this_runCatching.block()                              |                                                         |                                                         |
 |   block()                                              |                                                         |                                                         |
 |     CollectionsKt.listOf(Array<Thread>#1): ArrayList#1 |                                                         |                                                         |
-|     threads.iterator(): ArrayItr#1                     |                                                         |                                                         |
+|     iterator(): ArrayItr#1                             |                                                         |                                                         |
```
ndkoval: 1. It should remain `threads.iterator`
2. We should provide a better representation for `forEach` -- let's create an issue

src/jvm/test/resources/expected_logs/run_concurrent_test/deque.txt
```diff
@@ -37,11 +37,11 @@ The following interleaving leads to the error:
 |                                                        |                                         | run()                  |
 |                                                        |                                         |   $deque.addFirst(0)   |
```
ndkoval, line 32: Please raise an issue to not show `$` in the beginning of the captured variable

src/jvm/test/resources/expected_logs/run_concurrent_test/deque.txt
```diff
@@ -37,11 +37,11 @@ The following interleaving leads to the error:
...
 |                                                        |                                         |   $deque.peekLast(): 1 |
-|                                                        |                                         |   item.intValue(): 1   |
+|                                                        |                                         |   1.intValue(): 1      |
 |                                                        |                                         |   r2 = 1               |
 |                                                        |     p.item.compareAndSet(1, null): true |                        |
 |                                                        |     unlink(Node#2)                      |                        |
-|                                                        |   item.intValue(): 1                    |                        |
+|                                                        |   1.intValue(): 1                       |                        |
```
ndkoval, line 38: Please create an issue to get rid of boxing/unboxing method calls

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
```diff
@@ -15,7 +15,7 @@ The following interleaving leads to the error:
...
 |     ArrayList#1.add(Thread#3): true                    |                                                                                                             |                                                                                                             |                                                                                                             |
-|     threads.iterator(): Itr#1                          |                                                                                                             |                                                                                                             |                                                                                                             |
+|     iterator(): Itr#1                                  |                                                                                                             |                                                                                                             |                                                                                                             |
```
ndkoval: It should remain `threads.iterator()`

src/jvm/test/resources/expected_logs/multiple_suspension_points_channels_trace_debugger.txt
```diff
@@ -53,14 +53,14 @@ The following interleaving leads to the error:
...
 |     send$suspendImpl(BufferedChannel#1, 2, Continuation#4): COROUTINE_SUSPENDED |                                                                                                    |
 |     switch (reason: coroutine is suspended)                                     |                                                                                                    |
 |                                                                                 |   ResultKt.throwOnFailure(Unit)                                                                    |
 |                                                                                 |   channel1.receive() [suspendable: Continuation#1]: 2                                              |
-|                                                                                 |   2.intValue(): 2                                                                                  |
+|                                                                                 |   intValue(): 2                                                                                    |
```
ndkoval, line 64: 1. We should not show function calls for boxing/unboxing
2. It is called not on `this`, so there should be a bug

src/jvm/test/resources/expected_logs/inline_ext_fun.txt
```diff
@@ -22,7 +22,7 @@ Detailed trace:
 |   MethodHandle.linkToStatic("SOME", MemberName#1): "SOME" at DirectMethodHandle.Holder.invokeStatic(DirectMethodHandle$Holder:0)                   |
 |   DirectMethodHandle.internalMemberName(DirectMethodHandle#2): MemberName#2 at DirectMethodHandle.Holder.invokeStatic(DirectMethodHandle$Holder:0) |
 |   MethodHandle.linkToStatic(0, "SOME", MemberName#2): 4 at DirectMethodHandle.Holder.invokeStatic(DirectMethodHandle$Holder:0)                     |
-|   argL0.invokeBasic(4, 1): 5 at DelegatingMethodHandle.Holder.reinvoke_L(DelegatingMethodHandle$Holder:0)                                          |
+|   DirectMethodHandle#3.invokeBasic(4, 1): 5 at DelegatingMethodHandle.Holder.reinvoke_L(DelegatingMethodHandle$Holder:0)                           |
```
ndkoval: Could you please recheck this change?

src/jvm/test/resources/expected_logs/concurrent_linked_deque_trace_debugger.txt
```diff
@@ -75,7 +75,7 @@ Detailed trace:
 |                                                                                                                                     |     last(): Node#1 at ConcurrentLinkedDeque.peekLast(ConcurrentLinkedDeque.java:887)                              |
 |                                                                                                                                     |       tail ➜ Node#2 at ConcurrentLinkedDeque.last(ConcurrentLinkedDeque.java:737)                                 |
 |                                                                                                                                     |       p.next ➜ Node#1 at ConcurrentLinkedDeque.last(ConcurrentLinkedDeque.java:738)                               |
-|                                                                                                                                     |       p.next ➜ null at ConcurrentLinkedDeque.last(ConcurrentLinkedDeque.java:738)                                 |
```
ndkoval, line 78: Following the source code, it should be `p.next`:

```
            for (Node<E> t = tail, p = t, q;;) {
                if ((q = p.next) != null &&
                    (q = (p = q).next) != null)
```

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
```diff
@@ -54,11 +88,14 @@ class AccessPath(val locations: List<AccessLocation>) {
...
                 is LocalVariableAccessLocation -> {
-                    builder.append(location.variableName)
+                    if (!isThisName(location.variableName)) {
+                        builder.append(location.variableName)
+                    }
                 }
                 is StaticFieldAccessLocation -> {
                     with(builder) {
-                        append(".")
+                        // append(location.className.getSimpleClassName())
```
ndkoval: In case you want to keep this code for debugging, please leave a comment

src/jvm/test/resources/expected_logs/suspend_trace_reporting_trace_debugger.txt
eupp: Here the problem is because of the `inline` functions. 

`lock` and `unlock` calls are part of the `withLock` inline functions. 
Inside this function, they are accessed via `$this$withLock_u24default` variable, which is filtered from the trace according to the "filter `this` owner name" rule. 

Once we finish https://github.com/JetBrains/lincheck/pull/782, the problem will be mitigated, as we will have the following trace:
```
lock.withLock()
  lock()
  ...
  unlock()
```  

For now, I can relax the "filter `this` owner name" rule for inline's `this`, and print it as follows:
```
$this$withLock.lock()
...
$this$withLock.unlock()
```

@ndkoval WDYT?

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
```diff
@@ -54,11 +88,14 @@ class AccessPath(val locations: List<AccessLocation>) {
         for (location in locations) {
             when (location) {
                 is LocalVariableAccessLocation -> {
-                    builder.append(location.variableName)
+                    if (!isThisName(location.variableName)) {
```
ndkoval: Should we always apply the logic with `if (!isThisName(location.variableName)) { .. }`?

src/jvm/test/resources/expected_logs/run_concurrent_test/livelock.txt
eupp: This is the same problem with `inline` functions.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
eupp: This is the same problem with inline functions.

src/jvm/test/resources/expected_logs/multiple_suspension_points_channels_trace_debugger.txt
eupp: Regarding (2): it inline functions again. It is called from inline `ensure` function, which is omitted currently.

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
eupp: Nope, just forgot to remove it.

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
eupp: > Should we always apply the logic with if (!isThisName(location.variableName)) { .. }?

What do you mean by "always" here? Do you mean "for other branches of this `when`" ?

ndkoval: Yes, for accesses to fields, for example

src/jvm/test/resources/expected_logs/multiple_suspension_points_channels_trace_debugger.txt
ndkoval: It's something added by the compiler, I suppose, and is called on an Integer object.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
ndkoval: How did it work before the change?

src/jvm/test/resources/expected_logs/suspend_trace_reporting_trace_debugger.txt
ndkoval: Let's then raise an issue

ndkoval: By the way, how did it work before your change?

common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
eupp: Hmm, yes, I think you right, will fix it.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
eupp: Before the change, we tracked the `threads` local variable state at runtime, and then, when resolving `.iterator()` method call owner name, we would have found it during local variables state lookup.

eupp: In the source code, `.iterator()` is accessed from `inline` function `forEach` via `this` --- so we don't see `threads` local variable anymore with the static owner name analyzer.

src/jvm/test/resources/expected_logs/suspend_trace_reporting_trace_debugger.txt
eupp: > By the way, how did it work before your change?

See the comment about `threads.iterator()` below.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
ndkoval: I see, thank you for the clarification. What would we ideally see in the trace then?

src/jvm/test/resources/expected_logs/run_concurrent_test/deque.txt
eupp: https://youtrack.jetbrains.com/issue/DR-291/Remove-leading-of-owner-names-in-Lincheck-traces

src/jvm/test/resources/expected_logs/run_concurrent_test/deque.txt
eupp: https://youtrack.jetbrains.com/issue/DR-292/Filter-boxing-unboxing-method-calls-from-the-trace

src/jvm/test/resources/expected_logs/single_suspension_point_trace_debugger.txt
eupp: Here the problem stands from usage of the following heuristic in the Lincheck's owner name resolution:

```
// lookup for a field name in the current stack frame `this`
shadowStackFrame.findCurrentReceiverFieldReferringTo(obj)?.let { access ->
    return if (isThisName(access.fieldName)) null else access.fieldName
}
```

If I disable it, then object representation `Continuation#3` is printed instead of `continuation1`.

However, with this heuristics disabled, we have another problem. Kotlin property accesses start to become ugly in the trace.

So we start to see

```
getJob(): JobImpl#1 at JobNode.dispose(JobSupport.kt:1351)
JobImpl#1.removeNode(ChildContinuation#1) at JobNode.dispose(JobSupport.kt:1351)
```

instead of:

```
getJob(): JobImpl#1 at JobNode.dispose(JobSupport.kt:1351)
job.removeNode(ChildContinuation#1) at JobNode.dispose(JobSupport.kt:1351)
```

for the following source code:
```kotlin
job.removeNode(this)
```

I think we instead should search for Kotlin property access patterns in the trace and prettify them. I created a tickets for this: https://youtrack.jetbrains.com/issue/DR-293/Improve-representation-of-Kotlin-property-accesses-in-the-trace.
I propose to keep the current behavior for now.

src/jvm/test/resources/expected_logs/concurrent_linked_deque_trace_debugger.txt
eupp: I think it's actually `q`, so the trace became correct:

Screenshot from 2025-08-04 23-58-33

src/jvm/test/resources/expected_logs/inline_ext_fun.txt
eupp: I think this is a code generated by the Kotlin compiler to compile string template (via invokedynamic), so it's hard to say what should be the expected trace here. In any case, names `argL0` are typically auto-generated names of fields of lambda objects storing captured values, so I guess previously it was incorrectly displayed in the trace.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
eupp: I think once we have `inline` method calls in the trace, the problem will be solved automatically. We should see the following trace after this:
```text
forEach()
    iterator(): Itr#1
    ...
```

that matches the source code.

src/jvm/test/resources/expected_logs/concurrent_linked_deque_trace_debugger.txt
ndkoval: Following the code, the read should be from `p`, but
(1) the compiler might generate a read from `q` -- the code is too non-trivial
(2) this code is hard to debug, so let's not spend time on it.

src/jvm/test/resources/expected_logs/inline_ext_fun.txt
ndkoval: We should not show all these details, let's raise an issue.

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
```diff
@@ -50,209 +50,209 @@ The following interleaving leads to the error:
...
+|                                                                                                                                                                                                                         |         deqIdx.getAndIncrement(): 0 at SemaphoreImpl.tryResumeNextFromQueue(Semaphore.kt:320)                                                                                          |
+|                                                                                                                                                                                                                         |         SemaphoreKt.access$getSEGMENT_SIZE$p(): 16 at SemaphoreImpl.tryResumeNextFromQueue(Semaphore.kt:321)                                                                           |
```
ndkoval, line 165: Please get `SemaphoreKt.access$getSEGMENT_SIZE$p` to the issue about supporting Kotlin setters/getters

src/jvm/test/resources/expected_logs/inline_ext_fun.txt
eupp: https://youtrack.jetbrains.com/issue/DR-302/Hide-invokedynamic-and-MethodHandles-implementation-details-from-the-trace

src/jvm/test/resources/expected_logs/suspend_trace_reporting.txt
eupp: Done

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
```diff
@@ -18,8 +18,8 @@ The following interleaving leads to the error:
...
-|     threads.forEach()                                  |                                                                                                             |                                                                                                             |                                                                                                             |
-|     threads.forEach()                                  |                                                                                                             |                                                                                                             |                                                                                                             |
+|     ArrayList#1.forEach()                              |                                                                                                             |                                                                                                             |                                                                                                             |
```
ndkoval, line 21: Shouldn't this one be fixed with inline functions?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
```diff
@@ -10,21 +10,53 @@
...
-import org.jetbrains.lincheck.descriptors.FieldAccessLocation
+
+internal fun findOwnerName(
+    className: String,
+    shadowStackFrame: ShadowStackFrame,
+): String? {
+    // If the class name matches the current instance's class in the shadow stack,
+    // return null since the field belongs to the current class context and doesn't
+    // need an explicit owner prefix
+    if (className == shadowStackFrame.instance?.javaClass?.name) {
```
ndkoval: Note: we might need to cache this info to reduce reflection calls

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
```diff
@@ -10,21 +10,53 @@
...
+    obj: Any?,
+    className: String,
+    codeLocationId: Int,
+    shadowStackFrame: ShadowStackFrame,
+    objectTracker: ObjectTracker,
+): String? {
+    if (obj == null) {
+        return findOwnerName(className, shadowStackFrame)
+    }
+    // do not prettify thread names
```
ndkoval, line 42: Why?

common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
```diff
@@ -18,6 +22,35 @@ fun String.adornedClassNameRepresentation(): String = this
...
+ * - `this` --- simply `this` local variable name;
+ * - `$this`, `this_$iv`--- synthetic local variable names generated to refer to `this` inside inline function;
+ * - `this$N` --- field name generated to refer to enclosing `this` object in a nested class;
+ *
+ * These names should not be used as "owner names" of the method call or field access, as they are "hidden".
+ */
+fun isThisName(name: String): Boolean =
+    (name == "this")                                    ||
+    (name == "\$this")                                  ||
+    name.matches(Regex("this\\$\\d+"))                  ||
```
ndkoval: Should we create a Regex instance each time?

common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
```diff
@@ -18,6 +22,35 @@ fun String.adornedClassNameRepresentation(): String = this
...
+    isInlineThisIVName(name)
+
+fun isInlineThisIVName(name: String): Boolean =
+    name.startsWith("this_\$iv")                        ||
+    name.contains(Regex("^\\\$this\\$.+?(\\\$iv)+$"))
```
ndkoval: Should we create a Regex instance each time?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
```diff
@@ -0,0 +1,814 @@
...
+            this.stack = null
+        }
+    }
+
+    override fun visitLdcInsn(value: Any?) {
+        super.visitLdcInsn(value)
+        if (!isInitialized) {
+            return
+        }
+        // TODO: need to investigate how to lookup constant name (if available)
```
ndkoval, line 231: Are these tickets for these TODOs?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
```diff
@@ -80,6 +89,12 @@ data class MethodVariables(val variables: LocalVariablesMap = emptyMap()) {
     fun getVarByName(name: String): Set<LocalVariableInfo> = varsByName.getOrElse(name, ::emptyList).toSet()
     fun hasVarByName(name: String): Boolean = varsByName.containsKey(name)
 
+    fun getActiveVar(index: StackSlotIndex): String? =
```
ndkoval: This function is never used

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
```diff
@@ -80,6 +89,12 @@ data class MethodVariables(val variables: LocalVariablesMap = emptyMap()) {
     fun getVarByName(name: String): Set<LocalVariableInfo> = varsByName.getOrElse(name, ::emptyList).toSet()
```
ndkoval: This function is never used

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
```diff
@@ -80,6 +89,12 @@ data class MethodVariables(val variables: LocalVariablesMap = emptyMap()) {
...
     fun hasVarByName(name: String): Boolean = varsByName.containsKey(name)
 
+    fun getActiveVar(index: StackSlotIndex): String? =
+        activeVars.firstOrNull { it.index == index }?.name
+
+    fun getActiveVars(): List<LocalVariableInfo> =
```
ndkoval: should it be renamed to `copyActiveVariables()`?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -86,6 +85,8 @@ internal class LincheckClassVisitor(
         if (isNative) {
             Logger.debug { "Skipping transformation of the native method $className.$methodName" }
             return mv
+        } else {
+            Logger.debug { "Transforming method $className.$methodName" }
```
ndkoval, line 89: Should we keep it?

integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx.collections.immutable/tests.contract.list.ImmutableListTest_empty.txt
```diff
@@ -3,643 +3,643 @@ ImmutableListTest.empty() at :0
...
+                getter.invoke(EmptyList@1): "[]" at ComparisonDSL.kt:21
+                actual ➜ SmallPersistentVector@1 at ComparisonDSL.kt:21
+                getter.invoke(SmallPersistentVector@1): "[]" at ComparisonDSL.kt:21
                 AssertionsKt.assertEquals("[]", "[]", "toString") at ComparisonDSL.kt:21
           CollectionBehaviorsKt.collectionBehavior$default(CompareContext@1, null, false, 3, null) at CollectionBehaviors.kt:10
             CollectionBehaviorsKt.collectionBehavior(CompareContext@1, "", true) at CollectionBehaviors.kt:104
               "".append(""): "" at CollectionBehaviors.kt:105
-              "".length(): 0 at CollectionBehaviors.kt:105
+              objectName.length(): 0 at CollectionBehaviors.kt:105
               "".append(""): "" at CollectionBehaviors.kt:105
```
ndkoval, line 62: Should we use pretty printing for owner names? I would argue using them for function parameters only.

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
eupp: It's fixed in a sense that we now have `ArrayList#1.forEach()` call in the trace (and for all its children we correctly omit printing `this` owner name). 

How to extract owner name of an inline function is a different question. It is not the same as for regular method calls. 

I am not familiar enough with inline functions codegen rules to understand how to do this (and if it is possible at all). I suggest to wait for @lev-serebryakov-jetbrains (or @zhelenskiy) and ask for his advice.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
eupp: Do you mean `shadowStackFrame.instance?.javaClass?.name` ?
I can just add class name as a field to `shadowStackFrame`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
eupp: It is same behavior as before. 

Essentially, because we print thread names in Lincheck trace header as `Thread #N`, and we want the thread names in the trace match their names in the header. If we start prettifying them, we lost this connection.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
eupp: I found it useful during debugging, let's keep it.

integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx.collections.immutable/tests.contract.list.ImmutableListTest_empty.txt
eupp: > Should we use pretty printing for owner names? 

What do you mean by "pretty printing" for owner names? 

We print "owner names" for "receiver objects" of field accesses and method calls. It is the same behavior as we had before in Lincheck. The only change is that now we use statically computed owner names instead of dynamic.

We can also use "owner names" to print method arguments --- but this is an orthogonal task, and it changes the behavior we had before in Lincheck.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
ndkoval: Should we show both the variable name and object names for threads, then, to keep connection with both source code and the trace output?

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
ndkoval: Got it. Let's make sure we have an issue for that and resolve this discussion.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
eupp: Removed

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
eupp: It was here before this PR, let's keep it.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
eupp: I can actually use `activeVariables` directly, so I removed this function.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/OwnerNameAnalyzer.kt
eupp: https://youtrack.jetbrains.com/issue/DR-310/Support-constants-in-OwnerNameAnalyzerAdapter

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
eupp: > Should we show both the variable name and object names for threads

Maybe, but then what is so special about thread objects in particular? I guess in this case we would like to apply this technique to all objects?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
eupp: Done

integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx.collections.immutable/tests.contract.list.ImmutableListTest_empty.txt
eupp: Created an issue for this: https://youtrack.jetbrains.com/issue/DR-311/Apply-owner-name-trace-prettifying-to-method-arguments

src/jvm/test/resources/expected_logs/run_concurrent_test/custom_threads_trace_debugger.txt
eupp: https://youtrack.jetbrains.com/issue/DR-312/Apply-owner-name-resolution-to-inline-method-owners

PR #779, opened 2025-07-28
Enable tests that were non-deterministic previously due to object numbering non-determinism

eupp:
Resolved https://github.com/JetBrains/lincheck/issues/593


src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -416,7 +412,10 @@ class ThreadPoolRunConcurrentRepresentationTest : BaseRunConcurrentRepresentatio
         //   and thus accesses to this object are not tracked,
         //   and the race on counter increment is not detected;
         // var counter = 0
-        val executorService = Executors.newFixedThreadPool(2)
+        val threadNumber = AtomicInteger(0)
+        val executorService = Executors.newFixedThreadPool(2) { runnable ->
+            Thread(runnable, "Thread-${threadNumber.incrementAndGet()}")
```
dmitrii-artuhov: Why is custom naming of threads important here?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -446,7 +445,7 @@ class ThreadPoolRunConcurrentRepresentationTest : BaseRunConcurrentRepresentatio
 class CoroutinesRunConcurrentRepresentationTest : BaseRunConcurrentRepresentationTest<Unit>(
     "run_concurrent_test/coroutines/coroutines"
 ) {
-    // TODO: check why flakiness appears in this test
+    // TODO: coroutine names are non-deterministic
```
dmitrii-artuhov: But you specify deterministic thread names and not coroutine names here. Could you please elaborate on the `todo` so that the problem is clearer? Taking a look at the issue does not help(

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
```diff
@@ -469,19 +468,25 @@ class CoroutinesRunConcurrentRepresentationTest : BaseRunConcurrentRepresentatio
...
-                r1 = channel2.receive()
-            }
-            val job2 = launch(dispatcher) {
-                channel2.send(sharedCounter++)
-                r2 = channel1.receive()
+    override fun block() {
+        val threadNumber = AtomicInteger(0)
+        val executorService = Executors.newFixedThreadPool(2) { runnable ->
+            Thread(runnable, "Thread-${threadNumber.incrementAndGet()}")
+        }
```
dmitrii-artuhov: You could add to the test utils a new class:
```kotlin
class DeterministicThreadNamesFactory(private val prefix: String) : ThreadFactory {
    private val threadNumber = AtomicInteger(0)

    override fun newThread(runnable: Runnable): Thread {
        val threadName = "$prefix-${threadCount.getAndIncrement()}"
        return Thread(runnable, threadName)
    }
}
```
and pass it in both places:

```kotlin
Executors.newFixedThreadPool(2, DeterministicThreadNamesFactory("Thread"))
```

It might be used elsewhere later

dmitrii-artuhov (changes requested):

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: Otherwise names are non-deterministic : (
Automatically generated names involve some global counter, that can change non-deterministically depending on the environment:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#defaultThreadFactory--

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: I extended the TODO comment:

```
// TODO: coroutine names are still non-deterministic.
//   Apparently, for some reason the `kotlinx.coroutines` is run in DEBUG mode,
//   and in this mode `COROUTINE_ID` is appended to each coroutine's name.
//   The `COROUTINE_ID` is global `AtomicLong` which is incremented on each new coroutine creation,
//   and thus is non-deterministic.
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/RunConcurrentRepresentationTests.kt
eupp: Done.
 
But the name `DeterministicThreadNamesFactory` was a bit unfortunate, because it will only be deterministic if it is created and used within the scope of a Lincheck test. So I decided to name it just `LincheckTestThreadFactory`.

PR #780, opened 2025-07-28
Refactor usage of `TRAppendable` so that there is no `var` fields in it

dmitrii-artuhov:


trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -53,7 +51,7 @@ sealed class TRTracePoint(
 
     val codeLocation: StackTraceElement get() = CodeLocations.stackTrace(codeLocationId)
 
-    abstract fun toText(verbose: Boolean): String
+    abstract fun toText(appendable: TRAppendable, verbose: Boolean)
```
eupp: Could we make `verbose` a parameter of `DefaultTRTextAppendable`, instead of passing it to every `toText()` invocation?

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -53,7 +51,7 @@ sealed class TRTracePoint(
...
 }
```
eupp, line 62: Also, let's add `toString(): String` that creates default appendable inside and delegates to `toText()`.

eupp (approved):

PR #781, opened 2025-07-29
Add printing of enhanced trace via eagerly traversing points with `LazyTraceReader`

dmitrii-artuhov:
Adds a function which print the whole trace to text but via `LazyTraceLoader` in order to apply all node compressions by `TracePostprocessor`

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/Printing.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/Printing.kt
@@ -10,10 +10,44 @@
 
 package org.jetbrains.lincheck.trace
 
+import java.io.File
 import java.io.OutputStream
 import java.io.PrintStream
 
 private const val OUTPUT_BUFFER_SIZE: Int = 16*1024*1024
+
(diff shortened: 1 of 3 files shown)


trace/src/main/org/jetbrains/lincheck/trace/Printing.kt
```diff
@@ -10,11 +10,45 @@
...
+    if (node is TRMethodCallTracePoint && node.events.isNotEmpty()) {
+        reader.loadChildrenRange(node, 0, node.events.size)
+        node.events.forEach { event ->
+            if (event != null) {
+                lazyPrintTRPoint(output, reader, event, depth + 1, verbose)
+            }
+        }
+    }
+}
+
```
lev-serebryakov-jetbrains, line 52: Add "forget" after printing, or it will load whole trace to memory in process. Also, there is `loadAllChildren()` for such cases.

dmitrii-artuhov:
@lev-serebryakov-jetbrains this feature is implemented as part of the bigger PR with owner names. See https://github.com/JetBrains/lincheck/pull/808 instead

PR #782, opened 2025-07-30
Lev serebryakov/inline methods revamp

lev-serebryakov-jetbrains:
It is continuation of eupp/inline-non-local-return

Inline methods support is completely rewritten. New features:

1. Support for non-local returns, any level deep.
2. Support for throws and exceptions.
3. Support for lambda arguments, with non-local returns and exceptions.
4. Support for nested inline methods which begins ad/or ends at same label.
5. Better detection of inline `this`.
6. Better detection of owners and their classes, especially for nested inline methods.
7. Support for «primitive» owners (like inline extension function for type `Int`).
8. Detection of call/return imbalances and methods mismatches, both statically (at transformation time) and dynamically (at execution/trace recording time) with extensive diagnostics*.
9. A lot of tests for all above.

* — Dynamic checks are implemented only in Trace Recorder mode for now.


lev-serebryakov-jetbrains:
Ooops., there is problem with coroutines `synchronized`

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -128,12 +131,12 @@ internal class LincheckClassVisitor(
             // Inline method call transformer relies on the original variables' indices,
             // so it should go before (in FIFO order) all transformers which can create local variables.
             // All visitors created AFTER InlineMethodCallTransformer must use a non-remapping Generator adapter.
```
eupp: Looks like this comment is not relevant anymore.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -128,12 +131,12 @@ internal class LincheckClassVisitor(
...
-            // mv = InlineMethodCallTransformer(fileName, className, methodName, desc, adapter, mv, locals)
+            mv = InlineMethodCallTransformer(fileName, className, methodName, desc, adapter, mv, locals, labels)
```
eupp, line 135: Can we now move `InlineMethodCallTransformer` near regular `MethodCallTransformer` ?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
```diff
@@ -230,12 +233,12 @@ internal class LincheckClassVisitor(
         // Inline method call transformer relies on the original variables' indices,
         // so it should go before (in FIFO order) all transformers which can create local variables.
         // All visitors created AFTER InlineMethodCallTransformer must use a non-remapping Generator adapter too.
-        // mv = InlineMethodCallTransformer(fileName, className, methodName, desc, adapter, mv, locals)
+        mv = InlineMethodCallTransformer(fileName, className, methodName, desc, adapter, mv, locals, labels)
```
eupp, line 233: Same comments here.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassFileTransformer.kt
```diff
@@ -93,15 +93,19 @@ object LincheckClassFileTransformer : ClassFileTransformer {
         val classNode = ClassNode()
         reader.accept(classNode, ClassReader.EXPAND_FRAMES)
 
-        val methods = mapMethodsToLabels(classNode)
-        val methodVariables = methods.mapValues { MethodVariables(it.value) }
+        val methodVariables = getMethodsLocalVariables(classNode)
+
+        // Don't use class/method visitors on classNode to collect labels, as
+        // MethodNode reset all labels on a re-visit (WHY?!).
+        // Only one visit is possible to have labels stable.
```
eupp, line 98: Same comment applies to `getMethodsLocalVariables` I guess, so better move it above.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassFileTransformer.kt
```diff
@@ -93,15 +93,19 @@ object LincheckClassFileTransformer : ClassFileTransformer {
...
+        val methodLabels = getMethodsLabels(classNode)
 
         val writer = SafeClassWriter(reader, loader, ClassWriter.COMPUTE_FRAMES)
-        val visitor = LincheckClassVisitor(writer, instrumentationMode, methodVariables)
+        val visitor = LincheckClassVisitor(writer, instrumentationMode, methodVariables, methodLabels)
         try {
             classNode.accept(visitor)
             writer.toByteArray().also {
-                if (dumpTransformedSources) {
+                if (true) {
```
eupp: Please revert.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LabelsTracker.kt
```diff
@@ -14,14 +14,16 @@ import org.objectweb.asm.Label
 import org.objectweb.asm.MethodVisitor
 
 /**
- * [LocalVariablesTracker] tracks active regions of variables in the method.
+ * [LabelsTracker] tracks active regions of variables in the method.
```
eupp: Should the comment be updated?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -155,7 +156,13 @@ object LincheckJavaAgent {
...
                 // for some reason, without `isNotEmpty()` check this code can throw NPE on JVM 8
                 if (classes.isNotEmpty()) {
-                    instrumentation.retransformClasses(*classes.toTypedArray())
+                    classes.forEach {
+                        try {
+                            instrumentation.retransformClasses(it)
+                        } catch (t: Throwable) {
+                            Logger.error { "Failed to retransforming class ${it.name}: ${t.message}" }
+                        }
+                    }
```
eupp, line 171: I am a bit scared there could be some subtle difference between:
```kotlin
instrumentation.retransformClasses(*classes.toTypedArray())
```
and
```kotlin
classes.forEach {
    instrumentation.retransformClasses(it)
}
```

Especially, wrt. performance. 

Can we keep the first version? Logging can be kept on exception.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -184,15 +196,15 @@ class TraceCollectingEventTracker(
             parameters = emptyList()
         )
         strategy.tracePointCreated(null, tracePoint)
-        threadData.pushStackFrame(tracePoint, Thread.currentThread())
+        threadData.pushStackFrame(tracePoint, Thread.currentThread(), false)
```
eupp:
```suggestion
threadData.pushStackFrame(tracePoint, Thread.currentThread(), isInline = false)
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -423,7 +435,7 @@ class TraceCollectingEventTracker(
             parameters = params.map { TRObjectOrNull(it) }
         )
         strategy.tracePointCreated(threadData.currentMethodCallTracePoint(), tracePoint)
-        threadData.pushStackFrame(tracePoint, receiver)
+        threadData.pushStackFrame(tracePoint, receiver, false)
```
eupp:
```suggestion
threadData.pushStackFrame(tracePoint, receiver, isInline = false)
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -485,16 +516,35 @@ class TraceCollectingEventTracker(
             parameters = emptyList()
         )
         strategy.tracePointCreated(threadData.currentMethodCallTracePoint(), tracePoint)
-        threadData.pushStackFrame(tracePoint, owner)
+        threadData.pushStackFrame(tracePoint, owner, true)
```
eupp:
```suggestion
threadData.pushStackFrame(tracePoint, owner, isInline = true)
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -547,7 +595,7 @@ class TraceCollectingEventTracker(
         )
         strategy.registerCurrentThread(threadData.threadId)
         strategy.tracePointCreated(null,tracePoint)
-        threadData.pushStackFrame(tracePoint, null)
+        threadData.pushStackFrame(tracePoint, null, false)
```
eupp: threadData.pushStackFrame(tracePoint, null, isInline = false)

trace/src/test/org/jetbrains/lincheck/trace/util/TRPlayground.kt
```diff
@@ -15,13 +15,14 @@ import org.jetbrains.lincheck.trace.TRMethodCallTracePoint
...
+import kotlin.random.Random
 
 fun main(args: Array<String>) {
     // Choose one!
 
-    // eagerLoadAndPrintTrace("<path to binary file>")
+    eagerLoadAndPrintTrace("C:\\home\\lev\\dada\\CachingCacheStorageTest_testFindAllLoadsDataFromDelegate487622695707007237.bin")
 
-    // lazyLoadAndPrintTrace("<path to binary file>")
+    // lazyLoadAndPrintTrace("C:\\home\\lev\\dada\\CachingCacheStorageTest_testFindAllLoadsDataFromDelegate487622695707007237.bin")
```
eupp: Please rollback

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/InlineMethodCallTransformer.kt
```diff
@@ -30,133 +36,340 @@ internal class InlineMethodCallTransformer(
...
-            locals.hasVarByName("\$completion") ||
-            locals.hasVarByName("\$continuation") ||
-            locals.hasVarByName("\$result")
-        )
+                methodType.argumentTypes.lastOrNull()?.className == contType &&
+                (
+                        locals.hasVarByName("\$completion") ||
+                                locals.hasVarByName("\$continuation") ||
+                                locals.hasVarByName("\$result")
+                        )
```
eupp: Please revert.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/InlineMethodCallTransformer.kt
```diff
@@ -30,133 +36,340 @@ internal class InlineMethodCallTransformer(
...
+                // Duplicate two top category-1 values
+                adapter.dup2()
+                adapter.invokeIfTrueAndInAnalyzedCode(opcode) {
+                    // Stop when we jump inside the inline method, comparison is strict because
+                    // "normal" exit code will be generated BEFORE this label (as method epilogue)
+                    exitFromInlineMethods { labelSorter.compare(it.endLabel, label) > 0 }
+                }
+            }
+
+            JSR -> Unit
```
eupp:
```suggestion
JSR -> {}
```

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/InlineMethodCallTransformer.kt
```diff
@@ -30,133 +36,340 @@ internal class InlineMethodCallTransformer(
...
-        // Create a fake method descriptor
-        val methodId = getPseudoMethodId(null, startLabel, inlineMethodName)
+    private fun processInlineMethodReturn(methodId: Int)  = adapter.run {
         push(methodId)
         invokeStatic(Injections::onInlineMethodCallReturn)
     }
 
-    private fun getPseudoMethodId(possibleClassName: String?, startLabel: Label, inlineMethodName: String): Int =
-        TRACE_CONTEXT.getOrCreateMethodId(possibleClassName ?: "$className\$$methodName\$$startLabel\$inlineCall", inlineMethodName, "()V")
+    private fun getPseudoMethodId(possibleClassName: String, inlineMethodName: String): Int =
```
eupp: Should we rename it to `getInlineMethodId` ?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/MethodLabels.kt
```diff
@@ -0,0 +1,54 @@
...
+
+package org.jetbrains.kotlinx.lincheck.transformation
+
+import org.objectweb.asm.Label
+import org.objectweb.asm.MethodVisitor
+
+class MethodLabels internal constructor(
+    labels: Map<Label, Int>
+): Comparator<Label> {
+    private class LabelInfo(val index: Int, var seen: Boolean)
```
eupp, line 25: Let's add a comment that `index` is a serial number of the label in the order of method instructions traversal, and that the comparator compares labels by their order in traversal.

src/jvm/test/resources/expected_logs/inlines/diversity.txt
```diff
@@ -0,0 +1,28 @@
...
+| ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| operation()                                                                                                                                             |
+|   escape = "START" at InlineDiversityRepresentationTest.operation(InlineDiversityRepresentationTest.kt:25)                                              |
+|   thisClassInlineFunc() at InlineDiversityRepresentationTest.operation(InlineDiversityRepresentationTest.kt:26)                                         |
+|     isc.otherClassInlineFunc() at InlineDiversityRepresentationTest.operation(InlineDiversityRepresentationTest.kt:46)                                  |
+|       outsideInlineFunc() at InlineDiversityRepresentationTest.operation(InlineDiversityRepresentationTest.kt:47)                                       |
+|         Lambda argument to thisClassInlineFunc() at InlineDiversityRepresentationTest.operation(InlineDiversityRepresentationTest.kt:48)                |
```
eupp, line 24: Let's add an issue to improve inline lambda argument representations, and let's link it to https://github.com/JetBrains/lincheck/issues/519

IMO, currently it looks un-intuitive. 
I would rather expect something like this:
```text
thisClassInlineFunc(..., Lambda#1)
  isc.otherClassInlineFunc()
    outsideInlineFunc()
      Lambda#1()
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/inlines/InlineLambdaConditionalNotTakenReturnTest.kt
```diff
@@ -0,0 +1,49 @@
...
+        // but due to https://github.com/JetBrains/lincheck/issues/500
+        // we cannot set trace-debugger & JDK-8 specific expected output file
+        assumeFalse(isInTraceDebuggerMode || isJdk8)
+    }
+
+    override fun operation() {
+        escape = "START"
+        run {
+            escape = "IN LAMBDA"
+            if (isFalse) return@run
```
eupp, line 35: Shouldn't it be `if (!isFalse) return@run` ?
And then `UNREACHABLE` should not appear in the trace.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/inlines/InlineLambdaConditionalNotTakenThrowTest.kt
```diff
@@ -0,0 +1,53 @@
...
+        // we cannot set trace-debugger & JDK-8 specific expected output file
+        assumeFalse(isInTraceDebuggerMode || isJdk8)
+    }
+
+    override fun operation() {
+        escape = "START"
+        try {
+            run {
+                escape = "IN LAMBDA"
+                if (isFalse) throw RuntimeException()
```
eupp, line 36: Same here

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/inlines/InlineLambdaReturnTest.kt
```diff
@@ -0,0 +1,46 @@
...
+ * Lincheck
+ *
+ * Copyright (C) 2019 - 2025 JetBrains s.r.o.
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.representation.inlines
```
eupp, line 11: Overall, impressive list of tests, thanks!

eupp (changes requested):

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/inlines/InlineLambdaConditionalNotTakenReturnTest.kt
lev-serebryakov-jetbrains: It is NOT TAKEN test, so `"UNREACHABLE"` should be changed!

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/representation/inlines/InlineLambdaConditionalNotTakenThrowTest.kt
lev-serebryakov-jetbrains: Same answer :)

src/jvm/test/resources/expected_logs/inlines/diversity.txt
lev-serebryakov-jetbrains: It is very hard to add `Lambda#1` to arguments of inline method. It can be done in DR-278 and DR-279 where I plan to make pre-analyzis on variables to detect "tear apart" calls before visiting labels. There it could be transformed in such nice representation.

lev-serebryakov-jetbrains: I've added this as comment to DR-279.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
lev-serebryakov-jetbrains: Problem is, if one class throw exception no classes are retransformed at all... Even if it is latest one in the list.

I understand your concern, but there is no other place to insert this catch on per-class basis.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/transformers/InlineMethodCallTransformer.kt
lev-serebryakov-jetbrains: How does this leak here!?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassVisitor.kt
lev-serebryakov-jetbrains: Hmm... I better don't for now.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: > I understand your concern, but there is no other place to insert this catch on per-class basis.

I think there actually is a place.
It's `LincheckClassFileTransformer::transformImpl` --- if there is some error during bytecode transformation, you should be able to intercept the exception there. Moreover, it gives you more flexibility on how to handle the error. For instance, instead on not-instrumenting the class at all, you can try to instrument it second time but with inline methods transformer disabled.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckClassFileTransformer.kt
lev-serebryakov-jetbrains: getMEthodLocalVariables() doesn't visit anything, they go directly to data structures. But good point

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
lev-serebryakov-jetbrains: Nope. Error is during verification by JDK native code. `transformImpl` returns byte array, it goes back to agent API, agent API calls class re-definition and, BLAM, exception.

eupp:
@ndkoval need to consult with you on what to do with inline lambda arguments representation in the trace. 

Currently it looks like follows:
```
threads.forEach()
    ...
    Lambda argument to forEach() // <- new behavior --- denotes a call to a lambda passed to inline forEach
        x = 42
    ...
```

It affects several Lincheck representation tests, and looks like it is going to pop-up in Kotlin traces a lot. 
IMO, it currently looks a bit unpolished in the trace. 

What should we do about it?
- Leave it as is for now.
- Disable printing lambda arguments for now (maybe for Lincheck only?)
- Try to apply some quick/simple approach to improve representation, e.g. display as:

```
threads.forEach()
    ...
    forEach$Lambda() // <- new behavior --- denotes a call to a lambda passed to inline forEach
        x = 42
    ...
```

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
eupp: BTW, ASM provides some utilities for bytecode verifcation:
- [CheckClassAdapter](https://asm.ow2.io/javadoc/org/objectweb/asm/util/CheckClassAdapter.html)
- [SimpleVerifier](https://asm.ow2.io/javadoc/org/objectweb/asm/tree/analysis/SimpleVerifier.html)

As far as I understand, those are not the same as JVM internal verification. But we might still consider adding them to our transformers chain --- perhaps they will be able to detect most common sources of `VerifyError`.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
```diff
@@ -26,7 +28,27 @@ internal fun List<LocalVariableInfo>.isUniqueVariable(): Boolean {
...
+            if (isInlineCallMarker) {
+                return name.removePrefix(INLINE_FUNC_PREFIX)
+            }
+            if (isInlineLambdaMarker) {
+                val match = INLINE_LAMBDA_PATTERN.matchEntire(name)
+                if (match != null) {
+                    val callerInlineName = match.groups["inlineName"]?.value ?: "<unknown inline>"
+                    return "Lambda argument to $callerInlineName"
+                } else {
+                    return name.removePrefix(INLINE_LAMBDA_PREFIX)
```
harrytmthy, line 44: NIT: Would it be more consistent to always return `"Lambda argument to <unknown inline>"` instead?

```kotlin
if (isInlineLambdaMarker) {
    val match = INLINE_LAMBDA_PATTERN.matchEntire(name)
    val callerInlineName = match?.groups?.get("inlineName")?.value ?: "<unknown inline>"
    return "Lambda argument to $callerInlineName"
}
```

This would align the fallback with the valid case more closely and avoid raw fallback strings that look like unrelated identifiers 🙂

lev-serebryakov-jetbrains:
> forEach$Lambda() // <- new behavior --- denotes a call to a lambda passed to inline forEach

It is trivial to implement right now.

ndkoval:
> need to consult with you on what to do with inline lambda arguments representation in the trace.

Let's do something that looks not ugly and can be done in <1 hour, and raise an issue to plan & solve it properly.

@eupp @lev-serebryakov-jetbrains

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -138,7 +139,13 @@ object LincheckJavaAgent {
...
                 // Re-transform the already loaded classes.
                 // New classes will be transformed automatically.
-                instrumentation.retransformClasses(*getLoadedClassesToInstrument().toTypedArray())
+                getLoadedClassesToInstrument().forEach {
+                    try {
+                        instrumentation.retransformClasses(it)
+                    } catch (t: Throwable) {
+                        Logger.error { "Failed to retransform class ${it.name}: ${t.message}" }
+                    }
+                }
```
eupp, line 148: Now there are several places with this duplicated loop. Let's extract it into a new method `LincheckJavaAgent::retransformClasses(classes: List<Class<*>>)`.

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LocalVariables.kt
eupp: @lev-serebryakov-jetbrains WDYT?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
lev-serebryakov-jetbrains: Ooops, github loaded this comment AFTER I've squish & merge. Sorry.

I hope, we'll revert these loops back to "internal" ones soon.

PR #784, opened 2025-07-31
Various fixes for Lincheck's owner names resolution

eupp:
This PR is made in preparation for integrating new ASM-based owner name resolution in Lincheck: https://github.com/JetBrains/lincheck/pull/777

This PR:
- Unifies owner name resolution rules for different atomics APIs.
- Isolates atomics owner name resolution and regular owner name resolution (will simplify migration in https://github.com/JetBrains/lincheck/pull/777)
- Fixes various small bugs and problems in the owner names resolution (see the changed representation tests' traces for examples).

 Please note that the new `AccessLocation` class hierarchy also was taken from https://github.com/JetBrains/lincheck/pull/777

### DIFF
--- a/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
+++ b/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
@@ -107,15 +107,6 @@ internal fun findAtomicOwnerName(
     var ownerName: String? = null
     var params: List<Any?>? = null
 
-    var arrayAccess = ""
-    if (apiKind == AtomicApiKind.ATOMIC_ARRAY) {
-        val info = atomicMethodDescriptor.getAtomicArrayAccessInfo(atomic, arguments)
-        arrayAccess = "[${(info.location as ArrayElementByIndexAccessLocation).index}]"
-        params = info.arguments
-    } else if (apiKind == AtomicApiKind.ATOMIC_OBJECT) {
-        params = arguments.asList()
-    }
-
     fun getOwnerName(owner: Any?, className: String?, location: AccessLocation): String {
         val owner = findOwnerName(owner, className, shadowStackFrame, objectTracker, constants)
         return when (location) {
@@ -148,6 +139,16 @@ internal fun findAtomicOwnerName(
             ?: shadowStackFrame.findLocalVariableReferringTo(atomic)?.toString()
             // then try to search in local variables' fields
             ?: shadowStackFrame.findLocalVariableFieldReferringTo(atomic)?.toString()
+    }
+
+    var arrayAccess = "" // will contain `[i]` string denoting the accessed element index
+                         // in case we are dealing with atomic arrays API
+    if (apiKind == AtomicApiKind.ATOMIC_ARRAY) {
+        val info = atomicMethodDescriptor.getAtomicArrayAccessInfo(atomic, arguments)
+        arrayAccess = "[${(info.location as ArrayElementByIndexAccessLocation).index}]"
+        params = info.arguments
+    } else if (apiKind == AtomicApiKind.ATOMIC_OBJECT) {
+        params = arguments.asList()
     }
 
     if (apiKind == AtomicApiKind.ATOMIC_FIELD_UPDATER ||


src/jvm/test/resources/expected_logs/coroutine_cancellation.txt
```diff
@@ -53,13 +53,13 @@ Detailed trace:
...
-|                                                                                                                      |   _state.get(): InvokeOnCancel#1 at CancellableContinuationImpl.cancel(CancellableContinuationImpl.kt:0)                                                                                                            |
-|                                                                                                                      |   _state.compareAndSet(InvokeOnCancel#1, CancelledContinuation#1): true at CancellableContinuationImpl.cancel(CancellableContinuationImpl.kt:209)                                                                   |
+|                                                                                                                      |   cancellable$iv._state.get(): InvokeOnCancel#1 at CancellableContinuationImpl.cancel(CancellableContinuationImpl.kt:0)                                                                                             |
```
ndkoval, line 56: @dmitrii-artuhov could you please fix this `ancellable$iv` in a separate PR?

src/jvm/test/resources/expected_logs/switch_as_first_method_event.txt
```diff
@@ -6,43 +6,43 @@
...
-|                     |   result: 4                                                                                  |
-| ------------------------------------------------------------------------------------------------------------------ |
+| ------------------------------------------------------------------------------------- |
+|      Thread 1       |                            Thread 2                             |
+| ------------------------------------------------------------------------------------- |
+|                     | incTwiceAndGet(): 4                                             |
+|                     |   incAndGet(): 1                                                |
+|                     |   incAndGet(): 4                                                |
+|                     |     incAndGetImpl(): 4                                          |
+|                     |       getCounter$volatile$FU(): AtomicIntegerFieldUpdaterImpl#1 |
```
ndkoval, line 16: @dmitrii-artuhov, please take a look at this line -- it seems redundant

src/jvm/test/resources/expected_logs/suspended_not_awoke_trace_debugger.txt
```diff
@@ -6,21 +6,21 @@
...
+| ----------------------------------------------------------------------------------------------------- |
+|     Thread 1      |                                     Thread 2                                      |
+| ----------------------------------------------------------------------------------------------------- |
+| trigger()         |                                                                                   |
+|   allowed = true  |                                                                                   |
+|   switch          |                                                                                   |
+|                   | operation(): Suspended                                                            |
+|                   |   allowed ➜ true                                                                  |
+|                   |   neverWakeUpChannel.receive() [suspendable: Continuation#1]: COROUTINE_SUSPENDED |
+|                   |     receive$suspendImpl(BufferedChannel#1, Continuation#1): COROUTINE_SUSPENDED   |
```
ndkoval, line 18: @dmitrii-artuhov, please take a look at this intermediate method call -- we probably need to omit it

ndkoval (approved):
Approved + left a few comments on trace polishing (they should not be a part of this PR).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
```diff
@@ -0,0 +1,182 @@
...
+        params = info.arguments
+    } else if (apiKind == AtomicApiKind.ATOMIC_OBJECT) {
+        params = arguments.asList()
+    }
+
+    fun getOwnerName(owner: Any?, className: String?, location: AccessLocation): String {
+        val owner = findOwnerName(owner, className, shadowStackFrame, objectTracker, constants)
+        return when (location) {
+            is ArrayElementByIndexAccessLocation -> {
+                (owner ?: "this") + "[${location.index}]"
```
dmitrii-artuhov: for lines:
> 113: arrayAccess = "[${(info.location as ArrayElementByIndexAccessLocation).index}]"

and

> 123: (owner ?: "this") + "[${location.index}]"

Will the result contain double brackets in the representation, since you return from this method a concatination `ownerName + arrayAccess`? Why is `arrayAccess` needed in general?
Does not seem very obvious to me. Could you please check that there is no mistake and add a clarifying comment here

dmitrii-artuhov (approved):
Left a single comment, apart from that lgtm

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt
eupp: `arrayAccess` is added in case when we are processing atomic array API method call. It will denote the element index of the accesses array element. 

In `getOwnerName` we have `location` that could be `ArrayElementByIndexAccessLocation` only in cases of AFU, VarHandle or Unsafe APIs (disjoint with atomic arrays API case). In case of `apiKind == AtomicApiKind.ATOMIC_OBJECT || apiKind == AtomicApiKind.ATOMIC_ARRAY` it can only be `FieldAccessLocation`. 

I re-arranged the code a bit, and added a comment that should help to clarify the meaning.

src/jvm/test/resources/expected_logs/switch_as_first_method_event.txt
dmitrii-artuhov: I think, it is not: the method is generated by the atomicfu library, if I remove the `$volatile$fu` suffix, I will get the name `[get/set]FieldName`, here it will be `getCounter` which is completely unexpected for the user, since he did not declare anything named like this. Or in case he did, we would show the name of the method which wasn't called in fact, because method from atomicfu was intended

src/jvm/test/resources/expected_logs/suspended_not_awoke_trace_debugger.txt
dmitrii-artuhov: It appeared there because we can insert `<switch>: suspended coroutine` trace points right after the `methodName$suspendImpl` call. The compression only happens when `methodName$suspendImpl` is a single child in the trace tree. In this situation it is not the case.

Here is the fix, but the resulting trace does not look very pretty imho, maybe we shouldn't extend the compression to the case when there is a `<switch>` inserted https://github.com/JetBrains/lincheck/pull/788

src/jvm/test/resources/expected_logs/coroutine_cancellation.txt
dmitrii-artuhov: fixed here https://github.com/JetBrains/lincheck/pull/787

src/jvm/test/resources/expected_logs/switch_as_first_method_event.txt
ndkoval: Do we need to show AtomicFieldUpdater getters in the trace at all?

PR #796, opened 2025-08-04
Call `ensureObjectIsTransformed` for all fields recursively, even when the analyzing object class should not be retransformed

ndkoval:


src/jvm/main/org/jetbrains/lincheck/Lincheck.kt
```diff
@@ -121,3 +123,4 @@ internal class NoExceptionVerifier(@Suppress("UNUSED_PARAMETER") sequentialSpeci
         results.parallelResults[0][0] !is ExceptionResult
 }
 
+internal val LINCHECK_TEST_LOCK = ReentrantLock()
```
eupp: Why do we need a global lock for Lincheck tests now?
Let's add a comment to explain the reasoning.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
```diff
@@ -9,20 +9,14 @@
  */
 package org.jetbrains.kotlinx.lincheck
 
-import org.jetbrains.lincheck.LincheckAssertionError
+import kotlinx.atomicfu.locks.*
```
eupp: Are we sure we want ot use atomicfu here?
BTW, I suppose this import was added by mistake?

jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/transformation/LincheckJavaAgent.kt
```diff
@@ -162,7 +162,12 @@ object LincheckJavaAgent {
 
             // In the model checking mode, Lincheck processes classes lazily, only when they are used.
             instrumentationMode == MODEL_CHECKING || instrumentationMode == TRACE_RECORDING -> {
-                check(instrumentedClasses.isEmpty())
```
eupp, line 165: The purpose of this check was to ensure that previous test run cleaned up properly and unrolled the instrumentation of all classes. Do we want to get rid of this check?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/util/TestUtils.kt
```diff
@@ -24,6 +29,13 @@ import java.io.File
 import java.util.concurrent.ThreadFactory
 import java.util.concurrent.atomic.AtomicInteger
 
+internal inline fun withLincheckJavaAgentAndLincheckTestLock(
+    instrumentationMode: InstrumentationMode,
+    block: () -> Unit
+): Unit = LINCHECK_TEST_LOCK.withLock {
+    withLincheckJavaAgent(instrumentationMode, block)
+}
```
eupp: BTW, we can move it to `Lincheck.kt`, rename to something like `lincheckTest` or `withLincheckTestContext` and use it both in tests and Lincheck itself.

src/jvm/main/org/jetbrains/kotlinx/lincheck/LinChecker.kt
ndkoval: This is a mistake, fixed

eupp (approved):

PR #798, opened 2025-08-08
Support character escaping for args passed to javaagent

bbrockbernd:
Kotlin function names can contain spaces commas quotes etc which can mess up argument passing.

Parsing is updated to unescape everything that is escaped and properly split the string into args.

It is backwards compatible aka without escaping logic it should work normally.

### DIFF
--- a/jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/trace/agent/TraceAgentParameters.kt
+++ b/jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/trace/agent/TraceAgentParameters.kt
@@ -51,6 +51,17 @@ object TraceAgentParameters {
     }
 }
 
+/**
+ * Splits a given string of arguments into a list, using a comma as a delimiter.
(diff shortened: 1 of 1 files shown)


jvm-agent/src/main/org/jetbrains/kotlinx/lincheck/trace/agent/TraceAgentParameters.kt
```diff
@@ -30,8 +30,7 @@ object TraceAgentParameters {
         if (args == null) {
             error("Please provide class and method names as arguments")
         }
-
-        val actualArguments = args.split(",")
+        val actualArguments = splitArgs(args)
```
eupp, line 33: Could you please add somewhere a documentation on what is expected format of arguments' input, and how escaping works in this format.

ndkoval:
> It is backwards compatible aka without escaping logic it should work normally.

@bbrockbernd, please note that it's not necessary to preserve backwards compatibility at the current project stage.

eupp (approved):

PR #804, opened 2025-08-14
Fix stack overflow error on object graph traversal in `ensureObjectIsTransformed`

eupp:
Avoid recursive graph traversal in `ensureObjectIsTransformed` and use queue-based DFS instead. 
Re-use the implementation from `ObjectGraph.kt` for this purpose. 

Also had to move `ObjectGraph.kt` and few other things to `common` module in the process.

### DIFF
--- a/common/src/main/org/jetbrains/lincheck/util/AtomicMethods.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/AtomicMethods.kt
@@ -428,10 +428,10 @@ private val varHandleInstanceFieldExtractors = ConcurrentHashMap<Class<*>, VarHa
 
 
 internal fun isAtomic(receiver: Any?) =
-    isAtomicJava(receiver) ||
+    isJavaAtomic(receiver) ||
     isAtomicFU(receiver)
 
-internal fun isAtomicJava(receiver: Any?) =
+internal fun isJavaAtomic(receiver: Any?) =
     // java.util.concurrent
     receiver is AtomicReference<*> ||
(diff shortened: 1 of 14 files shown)


common/src/main/org/jetbrains/lincheck/util/AtomicMethods.kt
```diff
@@ -481,16 +484,25 @@ internal fun isAtomicArrayJava(receiver: Any?) =
 
 internal fun isAtomicFUArray(receiver: Any?) =
     // kotlinx.atomicfu
-    receiver is kotlinx.atomicfu.AtomicArray<*> ||
-    receiver is kotlinx.atomicfu.AtomicBooleanArray ||
-    receiver is kotlinx.atomicfu.AtomicIntArray ||
-    receiver is kotlinx.atomicfu.AtomicLongArray
+    receiver?.javaClass?.name?.let { isAtomicFUArrayClass(it) } ?: false
```
ndkoval, line 487: Should we avoid the comments below add modify the code to `receiver.isAtomicFUArray()`?

eupp: The comment is just to clarify why reflection is used instead of `is` checks. 

> modify the code to receiver.isAtomicFUArray()?

Not understood this part, what do you meant?

ndkoval: Let's create an extension function to not repeat the logic

PR #805, opened 2025-08-15
Minimize the static field traversals during lazy re-transformation 

eupp:
Traversing and transforming all static field (and reachable objects) in each call to `ensureClassHierarchyIsTransformed` is not necessary. 

We need to transform only the value read from a static field.


ndkoval:
@eupp please leave a comment about the performance impact

PR #806, opened 2025-08-22
Add more methods to inline calls instrumentation blocklists...

lev-serebryakov-jetbrains:
to support run IDEA tests.


ndkoval:
@lev-serebryakov-jetbrains, please make sure the tests are green

bbrockbernd (approved):

PR #810, opened 2025-09-01
Fix bug with missing thread forks in Trace Recorder

dmitrii-artuhov:


common/src/main/org/jetbrains/lincheck/descriptors/Descriptors.kt
```diff
@@ -33,7 +33,7 @@ data class MethodDescriptor(
 ) {
     var isIntrinsic: Boolean = false
 
-    val classDescriptor: ClassDescriptor = context.getClassDescriptor(classId)
+    val classDescriptor: ClassDescriptor get() = context.getClassDescriptor(classId)
```
dmitrii-artuhov, line 36: When multiple threads are run with trace recorder, then the context can be fully built only when all threads' index streams are read. Before that we cannot access the 'ClassDescriptor' objects from 'context', because they could be saved by another thread and, thus, not yet added to the index (there will be just nulls).

PR #811, opened 2025-09-01
Disable local variables tracking in lincheck

eupp:
- Remove all usages of local vars in owner name resolution in the Lincheck
- Disable local variables tracking bytecode transformer in Lincheck

Closes https://youtrack.jetbrains.com/issue/DR-360/Cut-out-local-variables-usage-for-atomics-in-findOwnerNames

Disabling local vars speed ups Lincheck tests execution, for instance:

- Locally on my machine, Lincheck integration tests: develop - 20m, this branch - 17m (~15% speed-up).
- On CI, kotlinx-coroutines tests: develop - 34m, this branch - 29m (~15% speed-up).
- Overall, on our different CI configurations we can observe speed-up around 10-15%.

Some representation tests has changed, because atomics owner name resolution still used local vars info to resolve owner name of field accessed through atomics APIs. Currently, it is not possible to restore owner name in these cases, because we need to lookup not the owner name of the receiver object (e.g., VarHandle instance) but owner name of one of the arguments of the method call. It will be possible to improve atomics owner name resolution after we implement https://youtrack.jetbrains.com/issue/DR-311/Apply-owner-name-trace-prettifying-to-method-arguments.  

- 
-


eupp:
@ndkoval 

I restored owner name behavior for atomics by implementing supporting method argument names in `OwnerNameAnalyzerAdapter`. 

Now there is only a single change in traces in `obstruction_freedom_violation_events_cut.txt` file. 

Instead of 
```
curTail.next.compareAndSet(null, Node#3): false
```
we now have:
```
AtomicReference#3.compareAndSet(null, Node#3): false
```

Here the problem is because `curTail.next.compareAndSet(..)` is actually `curTail.getNext().compareAndSet(...)` (Kotlin getter), and for method calls (i.e., `getNext()`) we do not treat them as "owner names". It is possible in theory to support method call chains inside owner names in `OwnerNameAnalyzer`, but it is a different task altogether. 

   
I also removed after read/write local var injections from `ManagedStrategy` and constants (static fields values) tracking. 

I have not yet removed `localVariables` from `ShadowStackFrame`, because apparently they are also used in Trace Recorder. Perhaps we can remove their usages from trace recorder as well, but I think we can postpone it and then re-visit the whole `ShadowStack` code (because currently there is a lot of duplicate logic around it in Lincheck & Trace Recorder).

common/src/main/org/jetbrains/lincheck/util/AtomicMethods.kt
```diff
@@ -101,6 +101,15 @@ internal fun getAtomicMethodDescriptor(obj: Any?, methodName: String): AtomicMet
     }
 }
```
ndkoval, line 103: @jetbrains-junie add a short comment describing the function contract

ghost:
Looks like you have spent the balance for Junie! No worries, we will help you – please contact us on our [Discord](https://jb.gg/junie/github) to get more balance.

ndkoval (approved):

PR #812, opened 2025-09-02
Implement (de)serialization of thread names

dmitrii-artuhov:


lev-serebryakov-jetbrains (changes requested):
Bump format version!

lev-serebryakov-jetbrains (approved):
LGTM now!

PR #819, opened 2025-09-09
Add new format of agent options: key=value pairs.

lev-serebryakov-jetbrains:


jvm-agent/build.gradle.kts
```diff
@@ -31,9 +35,26 @@ sourceSets {
...
+            javaToolchains.launcherFor {
+                val jdkToolchainVersion: String by project
+                val testInTraceDebuggerMode: String by project
+                val jdkVersion = jdkToolchainVersion.toInt()
+                // https://github.com/JetBrains/lincheck/issues/500
+                val jreVersion = if (testInTraceDebuggerMode.toBoolean() && jdkVersion == 8) 17 else jdkVersion
+                languageVersion.set(JavaLanguageVersion.of(jreVersion))
+            }
+        )
+    }
```
eupp: It looks like this was code was copied from top-level `build.gradle.kts`. 
Could you please instead extract into a method and put into `buildSrc/main/kotlin/Testing.kt`?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -23,24 +29,52 @@ object TraceAgentParameters {
...
-        val actualArguments = splitArgs(args)
-        classUnderTraceDebugging = actualArguments.getOrNull(0) ?: error("Class name was not provided")
-        methodUnderTraceDebugging = actualArguments.getOrNull(1) ?: error("Method name was not provided")
-        traceDumpFilePath = actualArguments.getOrNull(2)
-        if (actualArguments.size > 3) {
-            restOfArgs.addAll(actualArguments.subList(3, actualArguments.size))
+        // Try to parse new-style
+        val kvArguments = parseKVArgs(args)
+        if (kvArguments == null) {
+            Logger.warn { "Looks like old-style arguments found, consider migrate to key-value arguments" }
```
eupp, line 42: Do we need to support old-style? 
Can we just update all usages on plugin side and migrate in one step?

eupp (changes requested):

jvm-agent/build.gradle.kts
lev-serebryakov-jetbrains: Yep, without it tests fails with class version mismatch.

I'll extract it.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
lev-serebryakov-jetbrains: I think, it is better to make it as two-step process, as we can have external users. I'm always affraid of external users and prefer make deprecation-removal cycle.

eupp (approved):

PR #820, opened 2025-09-09
Add packed (single-file) trace format.

lev-serebryakov-jetbrains:
Add ability to put trace data and trace index into one ZIP after collection and open such ZIPs both with lazy and eager loaders.

Add some human-readable meta-info into into this archive too.

Old format can be loaded too, code try to detect is it ZIP or "raw" data. If ZIP is detected, metainformation is loaded too.

No changes to trace format or index themselves.

### DIFF
--- a/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
+++ b/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
@@ -13,6 +13,7 @@ package org.jetbrains.lincheck.trace.recorder
 import org.jetbrains.lincheck.analysis.ShadowStackFrame
 import org.jetbrains.lincheck.trace.TRACE_CONTEXT
 import org.jetbrains.lincheck.jvm.agent.LincheckJavaAgent
+import org.jetbrains.lincheck.jvm.agent.TraceAgentParameters
 import org.jetbrains.lincheck.trace.*
 import org.jetbrains.lincheck.util.*
 import sun.nio.ch.lincheck.*
@@ -147,12 +148,12 @@ class TraceCollectingEventTracker(
     private val methodName: String,
     private val traceDumpPath: String?,
     private val mode: TraceCollectorMode,
-    private val combine: Boolean
+    private val packTrace: Boolean
 ) : EventTracker {
     // We don't want to re-create this object each time we need it
     private val analysisProfile: AnalysisProfile = AnalysisProfile(false)
 
-    private val metaInfo = TraceMetaInfo.start(className, methodName)
+    private val metaInfo = TraceMetaInfo.start(TraceAgentParameters.rawArgs, className, methodName)
 
     // [ThreadDescriptor.eventTrackerData] is weak ref, so store it here too, but use
     // only at the end
@@ -638,7 +639,7 @@ class TraceCollectingEventTracker(
         startTime = System.currentTimeMillis()
 
         if (mode == TraceCollectorMode.BINARY_STREAM) {
-            if (combine) {
+            if (packTrace) {
                 packRecordedTrace(traceDumpPath!!, metaInfo)
             }
             return
@@ -669,7 +670,7 @@ class TraceCollectingEventTracker(
             when (mode) {
                 TraceCollectorMode.BINARY_DUMP -> {
                     saveRecorderTrace(traceDumpPath!!, TRACE_CONTEXT, roots)
-                    if (combine) {
+                    if (packTrace) {
                         packRecordedTrace(traceDumpPath, metaInfo)
                     }
                 }
--- a/trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
@@ -226,11 +226,28 @@ internal class CodeLocationsContext {
  */
 class LazyTraceReader private constructor(
     private val traceFileName: String,
-    private val input: TwoStreamProvider,
+    private val input: TraceDataProvider,
     private val postprocessor: TracePostprocessor
 ) : Closeable {
 
-    private class TwoStreamProvider(val baseFileName: String): AutoCloseable {
+    /**
+     * This class try to detect is provided path points to packed (ZIP) trace or
+     * raw data file.
+     *
+     * If provided file is packed trace, it unpacks data file as it needs to be seekable
+     * and packed file is not seekable, unfortunately.
+     *
+     * In any case it provides three data elements:
+     *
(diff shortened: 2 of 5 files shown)


trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -32,10 +36,201 @@ internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
...
+        private fun Appendable.appendMap(header: String, map: Map<String, String>) {
+            appendLine(header)
+            map.keys.sorted().forEach {
+                appendLine(" ${it.escapeKey()}=${map[it]?.escapeValue()}")
+            }
+        }
+
+        private fun String.escapeKey(): String =
+            replace("\\", "\\\\")
+            .replace("=", "\\=")
```
eupp:
```suggestion
        private fun String.escapeKey(): String = this
            .replace("\\", "\\\\")
            .replace("=", "\\=")
```

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -32,10 +36,201 @@ internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
...
+
+        private fun String.escapeValue(): String =
+            replace("\\", "\\\\")
+            .replace("\r", "\\r")
+            .replace("\n", "\\n")
+            .replace("\t", "\\t")
+    }
```
eupp:
```suggestion
        private fun String.escapeValue(): String = this
            .replace("\\", "\\\\")
            .replace("\r", "\\r")
            .replace("\n", "\\n")
            .replace("\t", "\\t")
    }
```

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -32,10 +36,201 @@ internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
...
+internal const val PACKED_META_ITEM_NAME = "info.txt"
 
 internal const val BLOCK_HEADER_SIZE: Int = Byte.SIZE_BYTES + Int.SIZE_BYTES
 internal const val BLOCK_FOOTER_SIZE: Int = Byte.SIZE_BYTES
 
+/**
+ * Information about conditions in which trace was collected.
+ *
+ *  - [className] — Name of traced class.
+ *  - [methodName] — Name of traced method.
```
eupp, line 53: Let's instead save parameters passed to `javaagent`. 
Would make it more future-proof in case we decide to change our tracing/instrumentation strategy (for instance, with the upcoming class filters feature).

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -32,10 +36,201 @@ internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
...
+ *  - [startTime] — start time of trace collection, as returned by [System.currentTimeMillis]
+ *  - [endTime] — end time of trace collection, as returned by [System.currentTimeMillis]
+ *  - [systemProperties] — state of system properties ([System.getProperties]) at the beginning of trace collection.
+ *  - [environment] — state of system environment ([System.getenv]) at the beginning of trace collection.
```
eupp, line 57: Can we also save the command that was used to run the jvm process? 
Thus we will have the full information to reproduce the run: command, system properties, environment variables.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -146,11 +146,14 @@ class TraceCollectingEventTracker(
     private val className: String,
     private val methodName: String,
     private val traceDumpPath: String?,
-    private val mode: TraceCollectorMode
+    private val mode: TraceCollectorMode,
+    private val combine: Boolean
```
eupp: Maybe rename `combine` to something like `pack` or `packTrace` ? BTW, `pack` is used everywhere else.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -222,20 +224,79 @@ internal class CodeLocationsContext {
...
-class LazyTraceReader(
-    private val dataFileName: String,
-    private val index: DataInputStream?,
+class LazyTraceReader private constructor(
+    private val traceFileName: String,
+    private val input: TwoStreamProvider,
     private val postprocessor: TracePostprocessor
 ) : Closeable {
 
+    private class TwoStreamProvider(val baseFileName: String): AutoCloseable {
```
eupp: Why is it called `TwoStreamProvider` ? 
The purpose of this class is not obvious from its name? 
Is it used to read and unpack zip file? 
Maybe then `ZipStreamProvider` or `ZipProvider` or something like this?

eupp (changes requested):

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
lev-serebryakov-jetbrains: I'll add them too! This is also human-readable format, and for human being it is simpler to read "parsed" version.

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
lev-serebryakov-jetbrains: I'll investigate is it possible to extract it somehow.,

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: It provides name of data (which cannot be read directly from ZIP, as we need to seek over it and ZIP stream is not seekable) and stream for index (maybe directly from ZIP, as it can be very basic stream, maybe from file). It provides them both for ZIP and "raw" traces, hide difference between two.

eupp: > It provides them both for ZIP and "raw" traces, hide difference between two.

Let's still rename it somehow. `TwoStreamProvider` is not obvious, what are these "two streams"? Maybe something like `TraceDataProvider`?

lev-serebryakov-jetbrains: Yep, I'll think about better name. Especially now, when it is read meta too (it was not in first attempt!)

eupp (approved):

PR #821, opened 2025-09-09
Rework bytecode transformation chain management 

eupp:
- Unify bytecode transformation chain management code between different instrumentation modes.
- Add ability to customize fine-grained transformation filters per instrumented class/method.


lev-serebryakov-jetbrains (approved):
LGTM.

PR #822, opened 2025-09-09
Implement bytecode class transformation filters for Trace Recorder

eupp:
Adds two new options to trace recorder jvm agent: `include` and `exclude` that can be used to customize include/exclude list of (fully-qualified) class name patterns. Class name patterns support wildcards `*`. 

Based on #821

### DIFF
--- a/integration-test/trace-debugger/src/main/org/jetbrains/trace/debugger/test/AbstractTraceDebuggerIntegrationTest.kt
+++ b/integration-test/trace-debugger/src/main/org/jetbrains/trace/debugger/test/AbstractTraceDebuggerIntegrationTest.kt
@@ -19,7 +19,7 @@ abstract class AbstractTraceDebuggerIntegrationTest : AbstractTraceIntegrationTe
         testClassName: String,
         testMethodName: String,
         extraJvmArgs: List<String>,
-        extraAgentArgs: List<String>,
+        extraAgentArgs: Map<String, String>,
         gradleCommands: List<String>,
         checkRepresentation: Boolean,
         testNameSuffix: String?,


integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -46,17 +46,18 @@ abstract class AbstractTraceIntegrationTest {
                         val options = task as JavaForkOptions
                         val jvmArgs = options.jvmArgs?.toMutableList() ?: mutableListOf()
                         jvmArgs.addAll(listOf(${extraJvmArgs.joinToString(", ") { "\"$it\"" }}))
-                        jvmArgs.add("-javaagent:${pathToFatJar.absolutePath.escape()}=$testClassName,$testMethodName,$pathToOutput${if (extraAgentArgs.isNotEmpty()) ",${extraAgentArgs.joinToString(",")}" else ""}")
+                        jvmArgs.add("-javaagent:${pathToFatJar.absolutePath.escape()}=class=$testClassName,method=$testMethodName,output=$pathToOutput${if (extraAgentArgs.isNotEmpty()) ",${extraAgentArgs.joinToString(",")}" else ""}")
```
lev-serebryakov-jetbrains: Should we change `extraAgentArgs` to `map`? Now it looks like mix of new-style (kv) and old-style (position) arguments.

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/AbstractTraceRecorderIntegrationTest.kt
```diff
@@ -22,20 +22,23 @@ abstract class AbstractTraceRecorderIntegrationTest : AbstractTraceIntegrationTe
         extraAgentArgs: List<String>,
```
lev-serebryakov-jetbrains: Make it map here and don't use list with `name=value` items looks more natural

PR #824, opened 2025-09-12
Add support for all-threads tracking

dmitrii-artuhov:
TODO:
- [x] collect stack trace of the thread when it is first registered via `beforeExistingThreadTracking` call and append them to the stored call stack 
- [x] add markers that would allow to highlight missing prefix (thread not tracked from its start)
- [x] and suffix (thread keeps working after finish of Main) of thread trace points on the plugin side
- [x] implement plugin highlighting 
- [x] tests (3 integration tests from coroutines)

PR in the plugin: https://jetbrains.team/p/concurrency-tools/reviews/106/files

### DIFF
--- a/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
@@ -18,6 +18,7 @@ import java.lang.invoke.CallSite;
  */
 public interface EventTracker {
 
+    void beforeExistingThreadTracking(Thread thread, ThreadDescriptor descriptor);
     void beforeThreadFork(Thread thread, ThreadDescriptor descriptor);
     void beforeThreadStart();
     void afterThreadFinish();
--- a/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
@@ -18,7 +18,7 @@ import java.lang.invoke.CallSite;
  */
 public interface EventTracker {
 
-    void beforeExistingThreadTracking(Thread thread, ThreadDescriptor descriptor);
+    void registerRunningThread(Thread thread, ThreadDescriptor descriptor);
     void beforeThreadFork(Thread thread, ThreadDescriptor descriptor);
     void beforeThreadStart();
     void afterThreadFinish();
--- a/bootstrap/src/sun/nio/ch/lincheck/Injections.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/Injections.java
@@ -52,7 +52,6 @@ public class Injections {
             ThreadDescriptor.setThreadDescriptor(t, descriptor);
         }
         descriptor.setEventTracker(allThreadsEventTracker);
-        descriptor.setTrackedFromStart(false);
         ThreadDescriptor.setCurrentThreadDescriptor(descriptor);
         allThreadsEventTracker.beforeExistingThreadTracking(t, descriptor);
         return descriptor;
--- a/bootstrap/src/sun/nio/ch/lincheck/Injections.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/Injections.java
@@ -30,7 +30,7 @@ public class Injections {
     public static Object lastSuspendedCancellableContinuationDuringVerification = null;
 
     /**
-     * In case if {@code globalEventTracker != null} is {@code false}, returns the current thread descriptor or {@code null}.
+     * In case if {@code globalEventTracker == null}, returns the current thread descriptor or {@code null}.
      * </br></br>
      * Otherwise, creates a new thread descriptor for the current thread. This thread is considered
      * as the one not tracked from the start, so it is registered in the {@code globalEventTracker}
--- a/bootstrap/src/sun/nio/ch/lincheck/Injections.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/Injections.java
@@ -22,8 +22,41 @@ public class Injections {
     // Special object to represent void method call result.
     public static final Object VOID_RESULT = new Object();
 
+    // Agent passes the event tracker if it decides to start recording trace points
+    // in already running threads.
+    public static EventTracker allThreadsEventTracker = null;
+
+    // Flag which enables tracking of events in all threads.
+    public static volatile boolean allThreadsTracked = false;
+
     // Used in the verification phase to store a suspended continuation.
     public static Object lastSuspendedCancellableContinuationDuringVerification = null;
+
+    /**
+     * In case if {@code allThreadsTracked} is {@code false}, returns the current thread descriptor or {@code null}.
+     * </br></br>
+     * Otherwise, creates a new thread descriptor for the current thread. This thread is considered
+     * as the one not tracked from the start, so it is registered in the {@code allThreadsEventTracker}
+     * by invoking {@code EventTracker.beforeExistingThreadTracking}.
+     * Eventually, the method returns the created thread descriptor.
+     */
+    private static ThreadDescriptor getOrCreateCurrentThreadDescriptor() {
+        ThreadDescriptor descriptor = ThreadDescriptor.getCurrentThreadDescriptor();
+        if (descriptor != null || !allThreadsTracked) {
+            return descriptor;
+        }
+        Thread t = Thread.currentThread();
+        descriptor = ThreadDescriptor.getThreadDescriptor(t);
+        if (descriptor == null) {
+            descriptor = new ThreadDescriptor(t);
+            ThreadDescriptor.setThreadDescriptor(t, descriptor);
+        }
+        descriptor.setEventTracker(allThreadsEventTracker);
+        descriptor.setTrackedFromStart(false);
+        ThreadDescriptor.setCurrentThreadDescriptor(descriptor);
+        allThreadsEventTracker.beforeExistingThreadTracking(t, descriptor);
+        return descriptor;
+    }
 
     public static EventTracker getEventTracker() {
         ThreadDescriptor descriptor = ThreadDescriptor.getCurrentThreadDescriptor();
@@ -41,6 +74,17 @@ public class Injections {
             // We are in the verification phase.
             lastSuspendedCancellableContinuationDuringVerification = cont;
         }
+    }
+
+    /**
+     * Enables tracking of all threads. Expects {@code eventTracker} to be the tracker
+     * which will be responsible for registering existing threads when they generate some trace point.
+     */
+    public static void enableAllThreadsTracking(EventTracker eventTracker) {
+        allThreadsEventTracker = eventTracker;
+        // Since `allThreadsTracked` is volatile, the order of assigning these variables is important:
+        // `allThreadsTracked` must be assigned afterward to allow for proper visibility of the `allThreadsEventTracker`.
+        allThreadsTracked = true;
     }
 
     /**
@@ -94,7 +138,11 @@ public class Injections {
      * @return true if the current thread is inside an ignored section, false otherwise.
      */
     public static boolean inAnalyzedCode() {
-        ThreadDescriptor descriptor = ThreadDescriptor.getCurrentThreadDescriptor();
+        // The tracking points for every existing thread are, basically, any of the events:
+        // field read/write, method call, etc. So to avoid inserting the call to each of the
+        // corresponding injection methods, `getOrCreateCurrentThreadDescriptor` is only called here:
+        // in the check which is invoked before each of the injection methods.
+        ThreadDescriptor descriptor = getOrCreateCurrentThreadDescriptor();
         if (descriptor == null) return false;
         return descriptor.inAnalyzedCode();
     }
--- a/bootstrap/src/sun/nio/ch/lincheck/Injections.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/Injections.java
@@ -23,37 +23,37 @@ public class Injections {
     public static final Object VOID_RESULT = new Object();
 
     // Agent passes the event tracker if it decides to start recording trace points
-    // in already running threads.
-    public static EventTracker allThreadsEventTracker = null;
-
-    // Flag which enables tracking of events in all threads.
-    public static volatile boolean allThreadsTracked = false;
+    // in already running threads. Field is non-null in case if all-threads tracking is enabled.
+    public static volatile EventTracker globalEventTracker = null;
 
     // Used in the verification phase to store a suspended continuation.
     public static Object lastSuspendedCancellableContinuationDuringVerification = null;
 
     /**
-     * In case if {@code allThreadsTracked} is {@code false}, returns the current thread descriptor or {@code null}.
+     * In case if {@code globalEventTracker != null} is {@code false}, returns the current thread descriptor or {@code null}.
      * </br></br>
      * Otherwise, creates a new thread descriptor for the current thread. This thread is considered
-     * as the one not tracked from the start, so it is registered in the {@code allThreadsEventTracker}
-     * by invoking {@code EventTracker.beforeExistingThreadTracking}.
-     * Eventually, the method returns the created thread descriptor.
+     * as the one not tracked from the start, so it is registered in the {@code globalEventTracker}
+     * by invoking {@code EventTracker.registerRunningThread}.
+     * Afterwards, the method returns the created thread descriptor.
      */
     private static ThreadDescriptor getOrCreateCurrentThreadDescriptor() {
         ThreadDescriptor descriptor = ThreadDescriptor.getCurrentThreadDescriptor();
-        if (descriptor != null || !allThreadsTracked) {
+        if (descriptor != null || globalEventTracker == null) {
             return descriptor;
         }
+        // handle the case when all threads tracking was requested,
+        // and we need to self-register current running thread and create new descriptor for it,
+        // basically, here we have `globalEventTracker != null`
         Thread t = Thread.currentThread();
         descriptor = ThreadDescriptor.getThreadDescriptor(t);
         if (descriptor == null) {
             descriptor = new ThreadDescriptor(t);
             ThreadDescriptor.setThreadDescriptor(t, descriptor);
         }
-        descriptor.setEventTracker(allThreadsEventTracker);
+        descriptor.setEventTracker(globalEventTracker);
         ThreadDescriptor.setCurrentThreadDescriptor(descriptor);
-        allThreadsEventTracker.beforeExistingThreadTracking(t, descriptor);
+        globalEventTracker.registerRunningThread(t, descriptor);
         return descriptor;
     }
 
@@ -79,11 +79,8 @@ public class Injections {
      * Enables tracking of all threads. Expects {@code eventTracker} to be the tracker
      * which will be responsible for registering existing threads when they generate some trace point.
      */
-    public static void enableAllThreadsTracking(EventTracker eventTracker) {
-        allThreadsEventTracker = eventTracker;
-        // Since `allThreadsTracked` is volatile, the order of assigning these variables is important:
-        // `allThreadsTracked` must be assigned afterward to allow for proper visibility of the `allThreadsEventTracker`.
-        allThreadsTracked = true;
+    public static void enableGlobalThreadsTracking(EventTracker eventTracker) {
+        globalEventTracker = eventTracker;
     }
 
     /**
--- a/bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
@@ -83,11 +83,6 @@ public class ThreadDescriptor {
      */
     private int ignoredSectionDepth = 0;
 
-    /**
-     * This flag indicates whether the thread is tracked from the start by the attached agent.
-     */
-    private boolean isTrackedFromStart = true;
-
     public ThreadDescriptor(Thread thread) {
         if (thread == null) {
             throw new IllegalArgumentException("Thread must not be null");
@@ -113,14 +108,6 @@ public class ThreadDescriptor {
 
     public void setEventTrackerData(Object eventTrackerData) {
         this.eventTrackerData = new WeakReference<>(eventTrackerData);
-    }
-
-    public boolean isTrackedFromStart() {
-        return isTrackedFromStart;
-    }
-
-    public void setTrackedFromStart(boolean trackedFromStart) {
-        isTrackedFromStart = trackedFromStart;
     }
 
     public void setAsRootDescriptor() {
--- a/bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
@@ -60,6 +60,9 @@ public class ThreadDescriptor {
 
     /**
      * This flag indicates whether the Lincheck analysis is currently enabled in this thread.
+     * <br><br>
+     * Note: the field is volatile because it could be modified from different threads:
+     * owner of this {@code ThreadDescriptor} and the Main thread of the test.
      */
     private volatile boolean isAnalysisEnabled = false;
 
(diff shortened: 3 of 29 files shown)


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -155,6 +157,14 @@ class TraceCollectingEventTracker(
     // only at the end
     private val threads = ConcurrentHashMap<Thread, ThreadData>()
 
+    // For proper completion of threads which are not tracked from the start of the agent,
+    // of of those threads which are not joined by the Main thread,
+    // we need to perform operations in them under locks.
+    // Note: the only place where there is a contention on the locks is
+    // when Main thread finishes and decides to "finish" all other running threads.
+    // By "finish" here we imply that it will dump their recorded data.
+    private val locks = ConcurrentHashMap<Thread, ReentrantLock>()
```
dmitrii-artuhov: This should be rewritten to thread-local variable + registering the link to the lock instance in some list. So that each read of the lock does not read from shared-memory and the list is required for synchronization when the Main Thread is completed and wants to disable tracking in all still-running threads

dmitrii-artuhov: We decided to implement a fine-grained volatile-flags-based solution instead

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -18,6 +18,7 @@
  */
 public interface EventTracker {
 
+    void beforeExistingThreadTracking(Thread thread, ThreadDescriptor descriptor);
```
eupp: Maybe `registerRunningThread` ?

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -22,9 +22,41 @@ public class Injections {
     // Special object to represent void method call result.
     public static final Object VOID_RESULT = new Object();
 
+    // Agent passes the event tracker if it decides to start recording trace points
+    // in already running threads.
+    public static EventTracker allThreadsEventTracker = null;
```
eupp: Let's rename to `globalEventTracker`.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -22,9 +22,41 @@ public class Injections {
...
+
+    // Flag which enables tracking of events in all threads.
+    public static volatile boolean allThreadsTracked = false;
```
eupp: Do we need both `allThreadsEventTracker` and `allThreadsTracked` flag? 

I think `allThreadsEventTracker` can serve as a flag itself: if `allThreadsEventTracker == null` then all threads tracking is disabled. 

You'll need to make `allThreadsEventTracker` volatile still.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -22,9 +22,41 @@ public class Injections {
...
     // Used in the verification phase to store a suspended continuation.
     public static Object lastSuspendedCancellableContinuationDuringVerification = null;
 
+    /**
+     * In case if {@code allThreadsTracked} is {@code false}, returns the current thread descriptor or {@code null}.
+     * </br></br>
+     * Otherwise, creates a new thread descriptor for the current thread. This thread is considered
+     * as the one not tracked from the start, so it is registered in the {@code allThreadsEventTracker}
+     * by invoking {@code EventTracker.beforeExistingThreadTracking}.
+     * Eventually, the method returns the created thread descriptor.
```
eupp:
```suggestion
     * Afterwards, the method returns the created thread descriptor.
```

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -22,9 +22,41 @@ public class Injections {
...
+     */
+    private static ThreadDescriptor getOrCreateCurrentThreadDescriptor() {
+        ThreadDescriptor descriptor = ThreadDescriptor.getCurrentThreadDescriptor();
+        if (descriptor != null || !allThreadsTracked) {
+            return descriptor;
+        }
+        Thread t = Thread.currentThread();
```
eupp: Could you please make "create the new descriptor" code branch more explicit,  either by:

- adding a comment:
```kotlin
if (descriptor != null || !allThreadsTracked) {
    return descriptor;
}
// handle the case when all threads tracking was requested 
// and we need to self-register current running thread and create new descriptor for it
Thread t = Thread.currentThread();
```

- or by enclosing it into a separate `if` branch explicitly:
```kotlin
if (descriptor != null) {
    return descriptor;
}
if (allThreadsTracked) {
    ...
}
``` 

- or both.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -43,6 +75,17 @@ public static void storeCancellableContinuation(Object cont) {
         }
     }
 
+    /**
+     * Enables tracking of all threads. Expects {@code eventTracker} to be the tracker
+     * which will be responsible for registering existing threads when they generate some trace point.
+     */
+    public static void enableAllThreadsTracking(EventTracker eventTracker) {
```
eupp: If we will rename `allThreadsEventTracker` to `globalEventTracker`, consider also renaming this method to `enableGlobalThreadsTracking`.

bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
```diff
@@ -61,7 +61,20 @@ public class ThreadDescriptor {
     /**
      * This flag indicates whether the Lincheck analysis is currently enabled in this thread.
      */
-    private boolean isAnalysisEnabled = false;
+    private volatile boolean isAnalysisEnabled = false;
```
eupp, line 67: Please update the documentation to explain why this field is `volatile`.

bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
```diff
@@ -61,7 +61,20 @@ public class ThreadDescriptor {
...
+     * <br><br>
+     * This flag is used together with {@code isAnalysisEnabled}, they allow for proper synchronization in
+     * case if Trace Recorder is used with {@code allThreadsTracked} flag enabled.
+     * <br>
+     * When the Main thread wants to dump the trace, it needs to logically "finish" all currently running threads
+     * so that they stop writing their trace points. For that Main marks some thread's {@code isAnalysisEnabled} flag as false,
+     * and then waits until that thread's {@code isInsideInjectedCode} flag is also false,
+     * afterward Main thread dumps the remaining tracepoints of thread that it waited for.
+     */
+    private volatile boolean isInsideInjectedCode = false;
```
eupp, line 83: Please also mention in the doc why this field is volatile.

common/src/main/org/jetbrains/lincheck/util/AnalysisSections.kt
```diff
@@ -156,6 +156,42 @@ inline fun <R> runInsideIgnoredSection(block: () -> R): R {
...
+ * @param block the code to execute within the injected code.
+ * @return result of the [block] invocation.
+ */
+inline fun <R> runInsideInjectedCode(block: () -> R): R? {
+    val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return block()
+
+    desc.enterInjectedCode()
+    if (!desc.isAnalysisEnabled) {
+        desc.leaveInjectedCode()
+        return null
```
eupp, line 195: Please add a comment to explain why double-check reading is used here (and synchronization scheme in general).

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -173,53 +183,97 @@ class TraceCollectingEventTracker(
...
+            strategy.registerCurrentThread(threadData.threadId)
+            for (frame in thread.stackTrace.reversed()) {
+                if (frame.className == "sun.nio.ch.lincheck.Injections") break
+                if (
+                    !frame.isLincheckInternals &&
+                    !frame.isNativeMethod &&
+                    !analysisProfile.shouldBeHidden(frame.className, frame.methodName)
+                ) {
+                    // TODO: should code locations be computed from the frame?
+                    appendMethodCall(null, frame.className, frame.methodName, "()V")
```
eupp: Is it a good idea to pass `()V` as a method descriptor? 

The called method could have different signature.
And when we will register it via `TRACE_CONTEXT.getOrCreateMethodId` it can collide with some existing method that actually have `()V` signature. 

Better to provide special `UNKNOWN_METHOD_SIGNATURE` for now. 
We can revisit this problem later --- I think it is possible to do something better via reflection and/or by maintaining line number <-> method descriptor mapping, but it is a separate task.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -173,53 +183,97 @@ class TraceCollectingEventTracker(
...
 
-    override fun beforeThreadFork(thread: Thread, descriptor: ThreadDescriptor) = runInsideIgnoredSection {
+    override fun beforeExistingThreadTracking(thread: Thread, descriptor: ThreadDescriptor): Unit = runInsideIgnoredSection {
+        val threadData = threads.computeIfAbsent(thread) {
+            val threadData = ThreadData(threads.size)
+            ThreadDescriptor.getThreadDescriptor(thread).eventTrackerData = threadData
+            threadData
+        }
+        val thread = Thread.currentThread()
+        descriptor.enableAnalysis()
```
eupp, line 191: Why not enabling the analysis at the end of the method?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -173,53 +183,97 @@ class TraceCollectingEventTracker(
...
+                obj = obj,
+                parameters = params
+            )
+            val parentTracePoint = if (threadData.getStack().isEmpty()) null
+                                   else threadData.currentMethodCallTracePoint()
+            strategy.tracePointCreated(parentTracePoint, methodCall)
+            threadData.pushStackFrame(methodCall, thread, isInline = false)
+        }
+
+        runInsideInjectedCode {
```
eupp, line 210: Why the whole method is not enclosed in `runInsideInjectedCode` ? 
Could you please add a comment?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -173,53 +183,97 @@ class TraceCollectingEventTracker(
...
-            methodId = TRACE_CONTEXT.getOrCreateMethodId("Thread", "run", "()V"),
-            obj = TRObject(Thread.currentThread()),
-            parameters = emptyList()
-        )
-        strategy.tracePointCreated(null, tracePoint)
-        threadData.pushStackFrame(tracePoint, Thread.currentThread(), isInline = false)
+        val thread = Thread.currentThread()
         threadDescriptor.enableAnalysis()
+
+        runInsideInjectedCode {
```
eupp, line 243: Why the whole method is not enclosed in `runInsideInjectedCode` ? 
Could you please add a comment?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -314,7 +368,7 @@ class TraceCollectingEventTracker(
         array: Any,
         index: Int,
         codeLocation: Int
-    ) {}
+    ) = Unit
```
eupp, line 373: Why this change?) 
IMO, empty method body `{ }` is cleaner/simpler.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -396,26 +455,25 @@ class TraceCollectingEventTracker(
...
+        val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: return
 
+            // TODO: make in optional
+//        val tracePoint = TRReadLocalVariableTracePoint(
+//            threadId = threadData.threadId,
+//            codeLocationId = codeLocation,
+//            localVariableId = variableId,
+//            value = TRObjectOrNull(value)
+//        )
+//        strategy.tracePointCreated(threadData.currentMethodCallTracePoint(), tracePoint)
```
eupp: please fix indentation here

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -616,13 +681,48 @@ class TraceCollectingEventTracker(
         startTime = System.currentTimeMillis()
     }
 
+    fun finishRunningThread(thread: Thread) {
```
eupp: Maybe `completeRunningThread` ?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -616,13 +681,48 @@ class TraceCollectingEventTracker(
...
+        //  2. we should write the ends of footers of each still-open MethodCallTracePoint's
+        //  3. - mark them some-how that their body wasn't finished when main finished
+        //  4. collect the prefix of the existing trace and create trace points for that
+        //  5. - mark existing threads that their prefixes are missing
+
+        // Finish existing threads, except for Main
+        val mainThread = Thread.currentThread()
+        threads.forEach { (thread, _) ->
+            if (thread != mainThread) {
+                finishRunningThread(thread)
```
eupp: Probably better to first call `disableAnalysis()` for all threads, and then wait for their completion.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -616,13 +681,48 @@ class TraceCollectingEventTracker(
...
+    }
+
     fun finishAndDumpTrace() {
+        // TODO:
+        //  1. disable analysis in all still-running threads
+        //     a. finish threads via CAS, so `afterThreadFinish` and `finishRunningThread` are not called simultaneously
+        //  2. we should write the ends of footers of each still-open MethodCallTracePoint's
+        //  3. - mark them some-how that their body wasn't finished when main finished
+        //  4. collect the prefix of the existing trace and create trace points for that
+        //  5. - mark existing threads that their prefixes are missing
```
eupp: Is this TODO still relevant?

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -192,6 +194,39 @@ class TRMethodCallTracePoint(
     }
 }
 
+class TRIncompleteMethodCallTracePoint(
```
eupp: Do we really need a new class? Why can't we just add a boolean flag to  `TRMethodCallTracePoint` ?

dmitrii-artuhov: The motivation for a new class was not to increase the serialization size of method calls. Most of the methods (like 99%) will have this flag always set to `false`. And only a couple (0-10) very first methods of each thread will have it as `true`. This would either increase the complexity of parsing (so that `false` values are not written to the trace somehow) or incerase the trace size by one byte per each method call

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: No, removed

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: indentation for the commented-out code is correct. Fixed the indentation for the TODO above

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: All other methods in this class, which have "empty" body, implement it via `= Unit` instead of ` {}`

bootstrap/src/sun/nio/ch/lincheck/Injections.java
dmitrii-artuhov: Agree

bootstrap/src/sun/nio/ch/lincheck/Injections.java
dmitrii-artuhov: I would avoid extra read of volatile, so I just added a comment

bootstrap/src/sun/nio/ch/lincheck/Injections.java
dmitrii-artuhov: Will apply locally

bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
dmitrii-artuhov: Added: 
```txt
...
Note: the field is volatile because it could be modified from different threads:
owner of this {@code ThreadDescriptor} and the Main thread of the test.
```

bootstrap/src/sun/nio/ch/lincheck/ThreadDescriptor.java
dmitrii-artuhov: Added the same as for the `isAnalysisEnabled` flag

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: `runInsideInjectedCode` internally checks that `isAnalysisEnabled == true` (when performing a double-checking), so the `descriptor.enableAnalysis()` should be called before the `runInsideInjectedCode`.

And that is also the reason why `runInsideInjectedCode` does not wrap the whole method: analysis of this thread has not been enabled yet, so `runInsideInjectedCode` would just exit without even invoking the body of this method

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: added

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: It must be enabled before calling `runInsideInjectedCode`, otherwise, `runInsideInjectedCode` would exit on the short-path without calling its lambda, because analysis is not enabled

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
eupp: IMO, could be a good idea to add general compactified `flags: Int` field to ` TRMethodCallTracePoint`, and then use one bit for `incomplete` flag. We might want to add more flags in the future.  

@lev-serebryakov-jetbrains WDYT?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: Adding a special `UNKNOWN_METHOD_SIGNATURE` does not seem very easy: it either requires to still be some correct asm method descriptor, or to be a subclass of `MethodSignature`, which would still need to be some correct parameters types in order for it to be correctly saved in trace and parsed on the plugin side. It does not look very clean. I think adding `UNKNOWN_METHOD_SIGNATURE` would result in the same outcome: overlap with a potential existing method stored in context.

On the other hand, I don't know why such overlap might be a problem: it will only lead to the fact that method signatures of these methods will be stored as a single record in the binary trace

eupp: > I think adding UNKNOWN_METHOD_SIGNATURE would result in the same outcome: overlap with a potential existing method stored in context.

`TraceContext` uses map with keys being `MethodDescriptor`-s. If we use `MethodDescriptor` with `()V` signature for unknown method call signature there could be a collision with actual method descriptor. If we would use special signature object that cannot belong to valid method signature, there would be no conflict. 

> On the other hand, I don't know why such overlap might be a problem

IMO, it can lead to very hard-to-debug bugs. Method descriptors can be registered in non-deterministic order, and thus this "unknown" descriptor can collide with actual method descriptor registration.  

> it either requires to still be some correct asm method descriptor, or to be a subclass of MethodSignature

`TraceContext` should allow registering method by passing `MethodSignature` instead of ASM string descriptor. IMO, it should not allow to pass ASM descriptor at all, it is different abstraction layer. 

> it either requires to still be some correct asm method descriptor, or to be a subclass of MethodSignature

As for "subclass of MethodSignature" --- there are two options. 

1. We can make `methodType: Types.MethodType?` parameter optional for `MethodSignature`, where `null` indicates that method type is unknown.

2. Or we can encode "invalid" method signature type as:
```
MethodType(argumentTypes = listOf(VoidType), VoidType)
```

and create a special constant for that: `UKNOWN_METHOD_TYPE = ...`. 

For normal methods it should not be possible to create a method type that accepts void as method parameter (you can add a `require` to regular `MethodType` constructor to check this invariant). 

Option (2) does not require any changes in (de)serialization code.
For option (1) you will need to adjust `DataOutput.writeMethodType` and `DataOutput.readMethodType`, but the changes are trivial (just add one byte for true/false to encode null/non-null method type).

CC @lev-serebryakov-jetbrains

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: done

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
lev-serebryakov-jetbrains: Yep, I'm for special signature and using class and not string as argument to context machinery. It will require more changes, but it will be more clear and bulletproof code.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: I'm for bit flags. Maybe 32 bit is too much, and 16 is enough for all our future needs? We'll can extend it later.
But I think two bytes per trace point is not a tragedy.

lev-serebryakov-jetbrains (changes requested):
Please, add flags to TRMethodCallTracrpoint instead of new tracepoint type.

And don't forget to bump serialization version.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: refactored

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
dmitrii-artuhov: added short (16bit) `flags` field and `INCOMPLETE_METHOD_FLAG`

eupp:
@dmitrii-artuhov thank you for addressing all the comments and concerns. 

How do you want to proceed with adding (integration?) tests?
In this PR or separately?

dmitrii-artuhov:
@eupp I want to at least add some here. Will do tomorrow morning

PR #825, opened 2025-09-12
Disabled field and array reads by default for Trace Recorder

bbrockbernd:
As discussed

### DIFF
--- a/integration-test/trace-recorder/src/main/resources/integrationTestData/TraceDebuggerExamples/org.examples.integration.bugs.LinkedHashSetTest_putAnObjectWithoutDefinedHashCode.txt
+++ b/integration-test/trace-recorder/src/main/resources/integrationTestData/TraceDebuggerExamples/org.examples.integration.bugs.LinkedHashSetTest_putAnObjectWithoutDefinedHashCode.txt
@@ -4,4 +4,3 @@ LinkedHashSetTest.putAnObjectWithoutDefinedHashCode() at :0
   SetsKt.linkedSetOf(LinkedHashSetTest.A;@1): LinkedHashSet@1 at LinkedHashSetTest.kt:10
   linkedHashSet.add(LinkedHashSetTest.A@1): true at LinkedHashSetTest.kt:11
   linkedHashSet.size(): 2 at LinkedHashSetTest.kt:12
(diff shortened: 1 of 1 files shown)


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/SharedMemoryAccessTransformer.kt
```diff
@@ -208,14 +202,16 @@ internal class SharedMemoryAccessTransformer(
...
+                if (configuration.trackSharedMemoryReads) {
+                    invokeIfInAnalyzedCode(
+                        original = {
+                            super.visitInsn(opcode)
+                        },
+                        instrumented = {
+                            processArrayLoad(opcode)
+                        }
+                    )
+                } else super.visitInsn(opcode)
```
eupp:
```suggestion
                } else { 
                    super.visitInsn(opcode)
                }
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -23,6 +23,7 @@ class TransformationConfiguration(
 
     var trackLocalVariableAccesses: Boolean = false,
     var trackSharedMemoryAccesses: Boolean = false,
+    var trackSharedMemoryReads: Boolean = false,
```
eupp, line 25: So as not to create more work in the future, could you please add `trackSharedMemoryWrites` right away?

Also, then could you please make `trackSharedMemoryAccesses` a derived property:
```
val trackSharedMemoryAccesses: Boolean
    get() = trackSharedMemoryReads || trackSharedMemoryWrites
    // NO SETTER!
```

And also add `trackAllSharedMemoryAccesses` for convinience (see `trackAllSynchronizationOperations` as an example):
```
var trackAllSharedMemoryAccesses: Boolean
        get() =
            trackSharedMemoryReads &&
            trackSharedMemoryWrites
        set(value) {
            trackSharedMemoryReads = value
            trackSharedMemoryWrites = value
        }
```

eupp (approved):

PR #826, opened 2025-09-12
Add trace recorder integration test with Kotlin Compiler and Ktor

zhelenskiy:


integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/KtorTraceRecorderIntegrationTest.kt
```diff
@@ -0,0 +1,75 @@
...
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.trace.recorder.test
+
+import org.junit.Ignore
+import org.junit.Test
+import java.nio.file.Paths
+
+class KtorTraceRecorderIntegrationTest : AbstractTraceRecorderIntegrationTest() {
```
ndkoval: How many tests do you run here?

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/KtorTraceRecorderIntegrationTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    @Ignore("Out of bounds")
+    fun `ktor-io`() = runKtorTests("io")
+
+    @Test
+    @Ignore("Inline method error")
+    fun `ktor-network`() = runKtorTests("network")
+
+    @Test
+    @Ignore("No output")
```
ndkoval: Please don't ignore tests

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/KtorTraceRecorderIntegrationTest.kt
zhelenskiy: AFAIK, 100-300. If you want, I can recalculate. Do you want me to?

zhelenskiy: Calculated, there are 95 now.

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -29,22 +29,26 @@ abstract class AbstractTraceIntegrationTest {
...
 
     private fun buildGradleInitScriptToDumpTrace(
+        gradleCommands: List<String>,
         testClassName: String,
         testMethodName: String,
         fileToDump: File,
         extraJvmArgs: List<String>,
         extraAgentArgs: Map<String, String>,
     ): String {
+        fun String.replaceDollar() = replace("$", "\\$")
```
lev-serebryakov-jetbrains: s/replace/escape/ ?

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -117,38 +141,67 @@ abstract class AbstractTraceIntegrationTest {
...
+
+                if (tmpFile.exists()) {
+                    checkNonEmptyNess(tmpFile)
                 } else {
-                    Assert.fail(
-                        "The gold data file was not found at '${expectedOutput.absolutePath}'. " +
-                        "Please rerun the test with \"overwriteRepresentationTestsOutput\" option enabled."
-                    )
+                    if (packedTraceFile.exists()) {
+                        checkNonEmptyNess(packedTraceFile)
```
lev-serebryakov-jetbrains, line 194: Why do we check for binary, packed output now?!

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -253,8 +253,8 @@ private fun splitArgs(args: String): List<String> {
 internal fun parseKVArgs(args: String): Map<String, String>? {
     var idx = 0
 
-    fun error(m: String): Map<String, String>? {
-        Logger.error { "Agent arguments error at position $idx: $m" }
+    fun warn(m: String): Map<String, String>? {
+        Logger.warn { "Agent arguments error at position $idx: $m" }
         return null
```
lev-serebryakov-jetbrains, line 258: Why? It is error in my eyes: we stop processing arguments. Really, we should stop running after that, and it is not done so only to allow backward-compatible parsing.

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
zhelenskiy: We check that it exists and is nonempty. It was Nikita's requirement.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
zhelenskiy: Before rebase it worked successfully after running the script. And as it worked, it's not error but a warning.

lev-serebryakov-jetbrains (approved):
LGTM!

PR #830, opened 2025-09-18
Fix bug with missing transformation of classes when static reads happen

dmitrii-artuhov:


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/SharedMemoryAccessTransformer.kt
```diff
@@ -51,13 +51,31 @@ internal class SharedMemoryAccessTransformer(
             return
         }
         when (opcode) {
-            GETSTATIC if configuration.trackSharedMemoryReads -> {
```
eupp, line 54: Let's please instead add more flags to profile to independently enable: `static` fields, instance fields, all field (static + instance), and array accesses. 

Then for trace recorder enable only `static` field reads, and put `ensureClassHierarchyIsTransformed` there.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -18,6 +18,8 @@
  */
 public interface EventTracker {
 
+    void ensureClassHierarchyIsTransformed(Class<?> clazz);
```
eupp: Let's rollback (see comment above).

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/KotlinCoroutinesTraceRecorderIntegrationTest.kt
```diff
@@ -10,13 +10,15 @@
...
 
+import org.junit.Ignore
 import org.junit.Test
 import java.nio.file.Paths
 
 class KotlinCoroutinesTraceRecorderIntegrationTest : AbstractTraceRecorderIntegrationTest() {
     override val projectPath: String = Paths.get("build", "integrationTestProjects", "kotlinx.coroutines").toString()
 
     @Test
+    @Ignore($$"java.lang.VerifyError: Failed to retransform class kotlinx.coroutines.EventLoopImplBase$DelayedTask")
```
eupp: These tests started failing after the fix?
Are they still failing if inline method calls transformer is disabled?

eupp (changes requested):

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/KotlinCoroutinesTraceRecorderIntegrationTest.kt
dmitrii-artuhov: ~~The inline transformer is disabled by the transformation profile as I saw, so these tests start to fail after the fix~~
Nevermind, I did not find it at first. I will try to disable

dmitrii-artuhov: After PR fixes I see not VerifyError, but now the tests are non-deterministic (because now previously missing `System.nanoTime()` calls are recorded)

ndkoval:
@dmitrii-artuhov please rebase onto recent develop to make the tests pass

eupp (approved):

PR #831, opened 2025-09-18
Prettify static method calls (DR-398)

dmitrii-artuhov:


common/src/main/org/jetbrains/lincheck/trace/TraceUtils.kt
```diff
@@ -10,10 +10,12 @@
 
 package org.jetbrains.lincheck.trace
 
+import com.sun.org.apache.xpath.internal.operations.Bool
```
eupp: redundant import

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -129,16 +129,35 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
...
         val ownerName = CodeLocations.accessPath(tracePoint.codeLocationId)
         if (ownerName != null) {
             ownerName.filterThisAccesses().takeIf { !it.isEmpty() }?.let {
-                appendAccessPath(it)
-                appendSpecialSymbol(".")
+                if (it.isObjectInstanceAccess()) {
+                    appendClassName(tracePoint.classDescriptor)
+                    appendSpecialSymbol(".")
+                }
+                else if (it.isCompanionAccess()) {
```
eupp: please put `else if` and `else` on the same line as previous `}`

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -129,16 +129,35 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
...
+                    appendSpecialSymbol(".")
+                }
             }
         } else if (tracePoint.obj != null) {
             appendObject(tracePoint.obj)
             appendSpecialSymbol(".")
-        } else {
+        }
+        // TODO: some refactoring is required here, because users can define classes, which end with 'Kt' as well
+        else if (!(tracePoint.isStatic() && tracePoint.className.endsWith("Kt"))) {
```
eupp: Please extract `Kt` check into separate method, similar how it is done for `isObjectInstanceAccess`  and `isCompanionAccess`

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -184,6 +191,11 @@ class TRMethodCallTracePoint(
     companion object {
         // Flag which tells that method was not tracked from its start and has some missing tracepoints
         const val INCOMPLETE_METHOD_FLAG: Int = 1
+        // Flag which tells that method is the test method from which trace recording started
+        const val IS_INITIAL_TEST_METHOD: Int = 1 shl 1
+        // Flag which tells that method is defined in the same class as the method which called this one
+        // or in its Companion
+        const val IS_CALLED_FROM_DEFINING_CLASS: Int = 1 shl 2
```
eupp: Do we need separate flags for these two? 
Couldn't this information be derived from the trace at trace printing stage?

dmitrii-artuhov: I can remove `IS_INITIAL_TEST_METHOD` because it serves the same purpose as `IS_CALLED_FROM_DEFINING_CLASS`, but the second flag alone is also enough

dmitrii-artuhov: Add `parentTracePoint` field to the `TRMethodCallTracePoint`, but serialization size did not change, this field is restored on loading, when children points are append to some method call trace point. Then this trace point will become a parent for those cihldren tracepoints (though, this field is only stored now in the method call trace points, so that runtime memory overhead is not too big)

eupp: > Add parentTracePoint field to the TRMethodCallTracePoint

That is good idea. 

Only one question though: how this `parent` field should interact with `TracePostprocessor` that can remove intermediate method call trace points from the tree?

dmitrii-artuhov: In place, where that happens, I reassign the parent as well

dmitrii-artuhov: here https://github.com/JetBrains/lincheck/pull/831/files#diff-d0b64ac54644e762fc903a2319d90e07ff117d609e705a7b5ef2725a54b33743

eupp:
For the future: let's consider making `parent` be a property of the base class `TRTracePoint`.

PR #832, opened 2025-09-19
Fix empty output file error in integration tests

zhelenskiy:
- Added functionality to support class hierarchy analysis in `TraceAgent` for inherited test methods.
- Removed ignored test method entries from `ktorTests.json`.


zhelenskiy:
I'm pretty sure that there are better ways to do this, but I don't know which ones. Point me at them, if you see them. @eupp

ndkoval:
@zhelenskiy, could you please explain the issue this PR fixes?

zhelenskiy:
The issue is empty output files generated by trace recorder.

There are two reasons for that: I accidentally added ignored test which expectedly pristine no output, the second is when the test method is inherited by the test class from some of ancestors.

ndkoval:
So, in the situation above
```
abstract class AbstractTest {
  @Test fun test() = ...
}

class MyTest : AbstractTest { .. }
```
If I call `MyTest.test`, Trace Recorder will try to inject the recording-related logic into a nonexistent `MyTest.test`. What we need is to find the closest in the hierarchy `test` method and instrument it, right?

zhelenskiy:
After this PR yes, before it did nothing. I'll update the PR to take only the first class with the method, as we discussed with @eupp, because we don't support multiple tests anyway now.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -139,6 +146,30 @@ object TraceAgentParameters {
             namedArgs.putAll(kvArguments)
         }
     }
+    
+    private fun setClassUnderTraceDebuggingMethodOwner(startClass: String = classUnderTraceDebugging, method: String = methodUnderTraceDebugging) {
+        classUnderTraceDebuggingMethodOwner = if (isInTraceDebuggerMode) {
```
eupp: Why adding check for `isInTraceDebuggerMode` here? Shouldn't the behavior be the same anyway?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -139,6 +146,30 @@ object TraceAgentParameters {
...
+            startClass
+        } else when (val clazz = runCatching { Class.forName(startClass) }.getOrNull()) {
+            null -> startClass
+            else -> calculateMethodOwner(clazz, method) ?: startClass
+        }
+    }
+    
+    private fun calculateMethodOwner(startClass: Class<*>, methodName: String, set: MutableSet<String> = mutableSetOf()): String? {
```
eupp: Let's rename to `findDeclaringClassOrInterface`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -139,6 +146,30 @@ object TraceAgentParameters {
...
+            when {
+                declaredMethod.name != methodName -> continue
+                !Modifier.isPublic(declaredMethod.modifiers) -> continue
+                Modifier.isAbstract(declaredMethod.modifiers) -> continue
+                Modifier.isStatic(declaredMethod.modifiers) -> continue
+                else -> return startClass.name
+            }
+        }
+        startClass.superclass?.let { calculateMethodOwner(it, methodName, set) }?.let { return it }
+        return startClass.interfaces.firstNotNullOfOrNull { calculateMethodOwner(it, methodName, set) }
```
eupp: In general, better to write DFS using stack instead of recursion.

eupp (changes requested):

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
zhelenskiy: It seems it makes little sense here. I see no practical example where the depth would be > 100_000 (hardly even more >10). Besides that, the recursive approach is more straight forward and easier to understand, so I chose it here.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
zhelenskiy: Done

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
zhelenskiy: It initially somehow broke some trace debugger test. So, I added the line, as we are not developing the trace debugger anymore. Maybe it will work this time. I'll try it again.

zhelenskiy: https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_KotlinxLincheck_BuildLinuxOnJava17TraceDebuggerIntegrationTests/5512020
No, it does not

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -83,6 +85,9 @@ object TraceAgentParameters {
 
     @JvmStatic
     lateinit var classUnderTraceDebugging: String
+    
+    @JvmStatic
+    lateinit var classUnderTraceDebuggingMethodOwner: String
```
eupp: Now there are two fileds with the same purpose `classUnderTraceDebugging` and  `classUnderTraceDebuggingMethodOwner`, and they are used inconsistently throughout codebase. 

Can we please leave only `classUnderTraceDebugging` ?

eupp (changes requested):

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
zhelenskiy: It seems they are not the same because replacing one with another would lead to calling all the ancestors of the base test class.

eupp: > It seems they are not the same because replacing one with another would lead to calling all the ancestors of the base test class.

Let's preserve the old behavior for the cases when `classUnderTestDebugging` is the declaring class of the method. 

For the cases when `classUnderTestDebugging` is not the declaring method, let's set `classUnderTestDebugging` to be the declaring class. 

All usages of `classUnderTestDebugging` should remain the same --- it is the class we are instrumenting.  

Currently there is a lot of confusion, because sometimes in transformers we lookup for `classUnderTraceDebuggingMethodOwner` as a class to instrument, and sometimes for `classUnderTestDebugging`.

eupp:
@zhelenskiy thanks for addressing all comments

PR #834, opened 2025-09-19
Add features to support loading trace in IDEA plugin

lev-serebryakov-jetbrains:


trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -36,10 +36,13 @@ internal const val TRACE_MAGIC : Long = 0x706e547124ee5f70L
 internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
 internal const val TRACE_VERSION : Long = 12
 
-internal const val INDEX_FILENAME_SUFFIX = ".idx"
-internal const val PACK_FILENAME_SUFFIX = ".packedtrace"
+// This suffix is not enforced, but IDEA plugin rely on it
+const val DATA_FILENAME_SUFFIX = ".trace"
+const val INDEX_FILENAME_SUFFIX = ".idx"
```
eupp: Should dot `.` be part of the suffix, or should it only contain file extension?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -283,13 +284,18 @@ class LazyTraceReader private constructor(
                         this.tmpDataFile = tmpDataFile
                     } catch (e: IOException) {
                         tmpDataFile.delete()
-                        return@run
+                        throwZipError(baseFileName)
```
eupp: Should you set `e` as cause of zip exception?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -767,6 +773,48 @@ fun loadRecordedTrace(traceFileName: String): TraceWithContext {
...
+    val buf = ByteBuffer.allocate(8)
+    if (input.read(buf.array()) != 8) return 0 // 0 is not magic for sure
+    return buf.getLong(0)
+}
+
+fun isPackedTrace(traceFileName: String): Boolean {
+    val input = openExistingFile(traceFileName) ?: return false
+    return try {
+        ZipInputStream(input).use { zip ->
+            val metaInfoEntry = zip.nextEntry ?: return@use false
```
eupp: I think you can use just `return false` (and everywhere below), here there is no difference in returning from `use` or from enclosing function.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
lev-serebryakov-jetbrains: I want to rework error handling completely to have good UX in plugin. Now it is silently "swallow" all exceptions, and I don't like to use exceptions in this area at all.

PR #835, opened 2025-09-19
Display value of primitive types

dmitrii-artuhov:
Done:
- Fixed a bug with displaying `$Ref` objects via access paths

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
@@ -188,7 +188,7 @@ abstract class AbstractTRFieldTracePointPrinter {
     protected fun TRAppendable.appendTracePoint(tracePoint: TRFieldTracePoint): TRAppendable {
         val isLambdaCaptureSyntheticField = isLambdaCaptureSyntheticField(tracePoint)
 
-        appendOwner(tracePoint)
+        appendOwner(tracePoint, isLambdaCaptureSyntheticField)
         appendFieldName(tracePoint, isLambdaCaptureSyntheticField)
         append(" ")
         appendSpecialSymbol(tracePoint.accessSymbol())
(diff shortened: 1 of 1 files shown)


ndkoval:
@dmitrii-artuhov please add a test

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -188,7 +188,7 @@ abstract class AbstractTRFieldTracePointPrinter {
     protected fun TRAppendable.appendTracePoint(tracePoint: TRFieldTracePoint): TRAppendable {
         val isLambdaCaptureSyntheticField = isLambdaCaptureSyntheticField(tracePoint)
 
-        appendOwner(tracePoint)
+        appendOwner(tracePoint, isLambdaCaptureSyntheticField)
```
eupp: Couldn't you just move:
```
val isLambdaCaptureSyntheticField = isLambdaCaptureSyntheticField(tracePoint)
```

to `appendOwner` directly, and do not add new method parameter?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -197,19 +197,22 @@ abstract class AbstractTRFieldTracePointPrinter {
         return this
     }
 
-    protected fun TRAppendable.appendOwner(tracePoint: TRFieldTracePoint): TRAppendable {
+    protected fun TRAppendable.appendOwner(tracePoint: TRFieldTracePoint, isLambdaCaptureSyntheticField: Boolean): TRAppendable {
         val ownerName = CodeLocations.accessPath(tracePoint.codeLocationId)
+        val appendDot = {
+            if (!isLambdaCaptureSyntheticField) appendSpecialSymbol(".")
+        }
```
eupp, line 207: Please add a comment why this is necessary, with an example of how trace looks like with/without this change

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
dmitrii-artuhov: I move the calculation to points where it is needed

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
dmitrii-artuhov: Added a description

dmitrii-artuhov:
Added a test here: https://github.com/ivandev0/TraceDebuggerExamples/pull/2
@eupp @ndkoval

eupp (approved):

PR #843, opened 2025-10-06
Disable global threads descriptors creation on the trace end

dmitrii-artuhov:
When the main thread finishes and disables analysis in all other running threads, we need to ensure that we do not register any more threads in `EventTracker::registerRunningThread`

### DIFF
--- a/common/src/main/org/jetbrains/lincheck/util/AnalysisSections.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/AnalysisSections.kt
@@ -168,7 +168,7 @@ inline fun <R> runInsideIgnoredSection(block: () -> R): R {
  * @return result of the [block] invocation or `null` if the analysis is disabled for the current thread.
  */
 inline fun <R> runInsideInjectedCode(block: () -> R): R? {
-    val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return block()
+    val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return null
 
     // General synchronization schema here is as follows:
     //   * This thread sets the flag, that it will now run some code from instrumentation (e.g. method from `EventTracker`)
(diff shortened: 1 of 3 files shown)


common/src/main/org/jetbrains/lincheck/util/AnalysisSections.kt
```diff
@@ -168,7 +168,7 @@ inline fun <R> runInsideIgnoredSection(block: () -> R): R {
  * @return result of the [block] invocation or `null` if the analysis is disabled for the current thread.
  */
 inline fun <R> runInsideInjectedCode(block: () -> R): R? {
-    val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return block()
+    val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return null
```
eupp, line 171: Could you please re-check. I think this method is used incorrectly for `TraceCollectingEventTracker::onMethodCallReturn`.

For `onMethodCallException` it will use `t: Throwable` as default value, but for `onMethodCallReturn` it will return `null`.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
```diff
@@ -69,6 +69,7 @@ object TraceRecorder {
         // and 'eventTracker.finishAndDumpTrace()' is called after analysis is disabled
         val desc = ThreadDescriptor.getCurrentThreadDescriptor() ?: return
         desc.removeAsRootDescriptor()
+        Injections.disableGlobalThreadsTracking()
```
eupp, line 71: It probably does not matter here, but in general a good practice to initialize/clean-up resources in reverse order.

So, in `installAndStartTrace` the order is:
```
desc.setAsRootDescriptor()
Injections.enableGlobalThreadsTracking(eventTracker)
```

Then here it should be in reverse order:
```
Injections.disableGlobalThreadsTracking()
desc.removeAsRootDescriptor()
```

Or vice versa, whatever order is more natural here.

common/src/main/org/jetbrains/lincheck/util/AnalysisSections.kt
dmitrii-artuhov: Methods `TraceCollectingEventTracker::onMethodCall[Return/Exception]` use another overload of this function:
```kotlin
inline fun <R> runInsideInjectedCode(default: R, block: () -> R): R {
    return runInsideInjectedCode(block) ?: default
}
```
and it would return `default` in such case (which is the `result: Any?` or `t: Throwable`)

PR #846, opened 2025-10-07
Improve performance and integration test infrastructure

zhelenskiy:
- Increased buffer sizes and optimized hot paths for better performance.
- Enhanced file emptiness checks in integration tests using buffered readers.
- Added cleanup for temporary index files during integration tests.
- Optimized hot paths.

### DIFF
--- a/integration-test/common/src/main/AbstractTraceIntegrationTest.kt
+++ b/integration-test/common/src/main/AbstractTraceIntegrationTest.kt
@@ -11,6 +11,7 @@
 import org.gradle.tooling.GradleConnector
 import org.gradle.tooling.ProjectConnection
 import org.jetbrains.kotlinx.lincheck_test.util.OVERWRITE_REPRESENTATION_TESTS_OUTPUT
+import org.junit.After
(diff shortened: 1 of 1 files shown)


integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -142,64 +149,76 @@ abstract class AbstractTraceIntegrationTest {
...
-                        *gradleCommands.toTypedArray(),
-                        "--tests",
-                        "$testClassName.$testMethodName",
-                    ).run()
-            }
+        taskQueue.add {
+            tmpFile.delete()
+            packedTraceFile.delete()
+            indexFile.delete()
+        }
```
lev-serebryakov-jetbrains, line 159: I don't understand, why do we create a lot of empty files in the first place!

zhelenskiy: I'm not sure, but it seems to be done by IDEA?

PR #848, opened 2025-10-09
Add `lazy` transformation toggle to jvm agent API

bbrockbernd:
- Introduced `lazy` parameter for trace agent configuration.
- Enabled conditional handling for lazy transformations.
- Added validation to flag potential risks with include filters when lazy mode is enabled.
- Updated related comments and documentation.

### DIFF
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckJavaAgent.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckJavaAgent.kt
@@ -186,15 +186,6 @@ object LincheckJavaAgent {
      * Also, retransforms already loaded classes.
      */
     fun install(instrumentationMode: InstrumentationMode) {
-        if (TraceAgentParameters.getLazyTransformationEnabled() && TraceAgentParameters.getIncludePatterns().isNotEmpty()) {
-            Logger.error { """
-                An `include` filter is provided but lazy transformation is enabled. 
-                This can lead to unexpected behaviour.
-                To disable lazy transformation provide `lazy=false` to the Jvm agent.
-            """.trimIndent() }
-        }
-        
-        
         this.instrumentationMode = instrumentationMode
         setInstrumentationStrategy()
         // The bytecode injections must be loaded with the bootstrap class loader,
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
@@ -152,6 +152,13 @@ object TraceAgentParameters {
 
             namedArgs.putAll(kvArguments)
         }
+        if (getLazyTransformationEnabled() && getIncludePatterns().isNotEmpty()) {
+            Logger.error { """
+                An `include` filter is provided but lazy transformation is enabled. 
+                This can lead to unexpected behaviour.
+                To disable lazy instrumentation provide `lazyInstrumentation=false` to the Jvm agent.
+            """.trimIndent() }
+        }
     }
     
     private fun setClassUnderTraceDebuggingToMethodOwner(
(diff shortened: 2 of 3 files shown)


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -32,15 +32,21 @@ import java.lang.reflect.Modifier
...
  *
- * - include — semicolon-separated list of include patterns (optional)
+ * - include — semicolon-separated list of include patterns (optional).
  *       Example: `include="org.example.*;com.acme.util.*"`
  *
- * - exclude — semicolon-separated list of exclude patterns (optional)
+ * - exclude — semicolon-separated list of exclude patterns (optional).
  *       Example: `exclude="org.example.internal.*;**.generated.*"`
  *
+ * - lazy — boolean that can disable lazy transformation, it is true by default.
```
eupp: Let's rename to `lazyTransformation` or `lazyInstrumentation`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentParameters.kt
```diff
@@ -32,15 +32,21 @@ import java.lang.reflect.Modifier
...
+ *      Example: `lazy=false`
+ *      
+ * - pack — boolean that enables zipping trace artifact files, it is false by default.
```
eupp, line 47: There are also other parameters (`format` and `formatOption`), which are specified in `TraceRecorderAgent.kt` (because they only work for trace recorder agent, but not trace debugger). 

I think since we are focusing on trace recorder right now, it rather makes sense to document all trace recorder parameters here for clarity.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckJavaAgent.kt
```diff
@@ -409,16 +418,17 @@ object LincheckJavaAgent {
...
      * Both stress and model checking modes implement some optimizations
      * to avoid re-transforming all loaded into VM classes on each run of a Lincheck test.
      * When this flag is set, these optimizations are disabled, and so
      * the Lincheck agent re-transforms all the loaded classes on each run.
      */
-    internal val INSTRUMENT_ALL_CLASSES =
-        System.getProperty("lincheck.instrumentAllClasses")?.toBoolean() ?: false
+    internal val INSTRUMENT_ALL_CLASSES = 
+        System.getProperty("lincheck.instrumentAllClasses")?.toBoolean() 
+            ?: !TraceAgentParameters.getLazyTransformationEnabled()
```
eupp: Let's rather leave `INSTRUMENT_ALL_CLASSES` intact for backward compatibility.

bbrockbernd: It is still intact. If `lincheck.instrumentAllClasses` property is set it will not be null and the extracted boolean will be set. If it is not set we fall back on the lazy parameter.

eupp: I meant, let's not change fallback behavior. Rather put this logic to `instrumentationStrategy` initialization.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckJavaAgent.kt
```diff
@@ -149,7 +186,17 @@ object LincheckJavaAgent {
      * Also, retransforms already loaded classes.
      */
     fun install(instrumentationMode: InstrumentationMode) {
+        if (TraceAgentParameters.getLazyTransformationEnabled() && TraceAgentParameters.getIncludePatterns().isNotEmpty()) {
+            Logger.error { """
+                An `include` filter is provided but lazy transformation is enabled. 
+                This can lead to unexpected behaviour.
+                To disable lazy transformation provide `lazy=false` to the Jvm agent.
+            """.trimIndent() }
+        }
```
eupp: If it's not too much trouble, could you also please move parameters validation to `TraceAgentParameters`.

PR #800, opened 2025-08-12
Loop detection via control-flow graph

ndkoval:

PR #851, opened 2025-10-17
Extract loops from CFG

dmitrii-artuhov:


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -169,6 +169,168 @@ class BasicBlockControlFlowGraph(
...
+     * Compute loops using dominators and back-edge detection.
+     * A back edge is an edge u -> h (non-exception) where h dominates u.
+     * For each header h, the loop body is the union of natural loops of its back edges.
+     */
+    private fun computeLoopsFromDominators(dominators: Array<Set<BasicBlockIndex>>): MethodLoopsInformation {
+        val n = basicBlocks.size
+        if (n == 0) return MethodLoopsInformation()
+
+        // Build predecessor maps (normal and all) and collect non-exception edges for exits and back-edges.
+        // TODO: this info is shared between dominators and back-edges calculation, its calculation could united
```
dmitrii-artuhov: Optimize here

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -169,6 +169,168 @@ class BasicBlockControlFlowGraph(
...
+        }
+        return Array(n) { doms[it].toSet() }
+    }
+
+    /**
+     * Compute loops using dominators and back-edge detection.
+     * A back edge is an edge u -> h (non-exception) where h dominates u.
+     * For each header h, the loop body is the union of natural loops of its back edges.
+     */
+    private fun computeLoopsFromDominators(dominators: Array<Set<BasicBlockIndex>>): MethodLoopsInformation {
```
dmitrii-artuhov: Split body into separate functions

jvm-agent/src/test/resources/analysis/controlflow/golden/loops/doWhileTrueLoop.txt
```diff
@@ -0,0 +1,13 @@
+LOOPS
+LOOP 1
+  HEADER: B1
+  BODY: B1
+  BACK EDGES:
+    B1 -> B1 : JUMP(opcode=GOTO)
+  NORMAL EXITS:
+
```
dmitrii-artuhov: Extra `\n` which should not be here

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -81,6 +81,11 @@ void cacheInvokeDynamicCallSite(
 
     BootstrapResult<?> invokeDeterministicallyOrNull(long descriptorId, Object descriptor, Object receiver, Object[] params);
 
+    void beforeLoopEnter(int codeLocation, int loopId);
+    void onLoopIteration(int codeLocation, int loopId);
+    void afterLoopExit(int codeLocation, int loopId);
+    void afterLoopExceptionExit(int codeLocation, int loopId, Throwable exception, boolean canEnterFromOutsideLoop);
+
```
dmitrii-artuhov: Should be part of the separate PR, not here. See Nikita's PR https://github.com/JetBrains/lincheck/pull/800

jvm-agent/src/test/org/jetbrains/lincheck/jvm/agent/ControlFlowGraphLoopsTests.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    @Test
+    fun whileWithTryCatchOutside() = test("whileWithTryCatchOutside", "(I)I")
+
+    @Test
+    fun nestedWhileWithTryCatchBetween() = test("nestedWhileWithTryCatchBetween", "(I)I")
+
+    @Test
+    fun nestedWhileWithTryCatchOutside() = test("nestedWhileWithTryCatchOutside", "(I)I")
+}
```
eupp, line 95: TODO: add a test case with `try { mayThrow(); while() {...}; mayThrow() } catch (...) { ... }`

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -0,0 +1,186 @@
...
+
+fun MethodLoopsInformation.prettyPrint(): String =
+    MethodLoopsInformationPrinter(this).prettyPrint()
+
+private class MethodLoopsInformationPrinter(val loopInfo: MethodLoopsInformation) {
+
+    fun prettyPrint(): String {
+        val sb = StringBuilder()
+
+        sb.appendLine("LOOPS")
```
eupp: Do we need this `LOOPS` header? Maybe let's just start printing loops right away.

jvm-agent/src/test/resources/analysis/controlflow/golden/loops/doWhileLoopContinueBreak.txt
```diff
@@ -0,0 +1,17 @@
...
+  HEADER: B1
+  BODY: B1, B2, B3, B5, B6
+  BACK EDGES:
+    B6 -> B1 : JUMP(opcode=IF_ICMPLT)
+  NORMAL EXITS:
+    B3 -> B4
+    B6 -> B7
+  EXCEPTION EXIT HANDLERS: NONE
+
+LOOPS BY BLOCK
```
eupp: Do we need to print this info?
Alternatively, we can just check right away in the tests that this bb-to-loop mapping is consistent with the loop-to-bbs mapping.

jvm-agent/src/test/resources/analysis/controlflow/golden/loops/doWhileTrueLoop.txt
```diff
@@ -0,0 +1,13 @@
...
+    NONE
+  EXCEPTION EXIT HANDLERS: NONE
```
eupp: Inconsistency here, for `NORMAL EXITS` `NONE` is printed on the next line, but for `EXCEPTION EXIT HANDLERS` on the same line. 

IMO, better to always print on the next line.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+     */
+    private fun computeDominators(): Array<Set<BasicBlockIndex>> {
+        val n = basicBlocks.size
+        if (n == 0) return emptyArray()
+
+        // Build predecessor lists; include only normal (non-exception) predecessors for loop analysis dominators.
+        // Dominators for loop/back-edge detection are typically based on normal control-flow.
+        val preds: Array<MutableSet<BasicBlockIndex>> = Array(n) { mutableSetOf<BasicBlockIndex>() }
+        for (e in edges) {
+            // Note: we count exception edges as well
```
eupp: Contradiction with the previous comment.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+        // Entry is block 0 when present
+        doms[0] = setOf(0)
+
+        var changed = true
+        while (changed) {
+            changed = false
+            for (b in 1 until n) {
+                val newSet = buildSet {
+                    // Intersection of dominators of predecessors: dom(b) = {b} + intersect(dom(p1), dom(p2), ...)
+                    val it = preds[b].iterator()
```
eupp: Please rewrite without using iterators explicitly. 
Also it probably makes sense to extract this into a common utility to build intersection of a collection of sets.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+                }
+            }
+        }
+        return Array(n) { doms[it].toSet() }
+    }
+
+    /**
+     * Compute loops using dominators and back-edge detection.
+     * A back edge is an edge u -> h (non-exception) where h dominates u.
+     * For each header h, the loop body is the union of natural loops of its back edges.
```
eupp: What are "natural loops"? Do we mention and explain this term somewhere in our docs?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+            // Start from each back edge source; perform reverse DFS over normal predecessors until reaching h
+            for (e in backEdges) {
+                val u = e.source
+                val stack = ArrayDeque<BasicBlockIndex>()
+                // Natural loop includes both u and h initially
+                if (body.add(u)) stack.add(u)
+                while (stack.isNotEmpty()) {
+                    val x = stack.removeLast()
+                    for (p in predsNormal[x]) {
+                        if (p == h) continue
```
eupp: Is this check necessary? Because header should already be in `body`, so the next `if` should fail anyway.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+            val h = e.target
+            val domU = dominators.getOrElse(u) { emptySet() }
+            if (h in domU) {
+                backEdgesByHeader.updateInplace(h, default = mutableSetOf()) { add(e) }
+            }
+        }
+        if (backEdgesByHeader.isEmpty()) return MethodLoopsInformation()
+
+        // For each header, compute loop body as a union of natural loops for each back edge to that header
+        data class LoopDescriptor(
```
eupp: Is it different from `LoopInformation`? Can we re-use the same class here?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+            }
+            loopsDescriptors += LoopDescriptor(h, body, backEdges)
+        }
+
+        // Build final LoopInformation list
+        val loops = mutableListOf<LoopInformation>()
+        loopsDescriptors.sortBy { it.header }
+        for ((id, desc) in loopsDescriptors.withIndex()) {
+            val header = desc.header
+            val headers = setOf(header) // reducible by construction
```
eupp: Where do we check it?
There should be a check for reducibility somewhere in our pipeline.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -0,0 +1,483 @@
...
+        // Map blocks to loop ids
+        val loopsByBlock = mutableMapOf<BasicBlockIndex, MutableList<LoopId>>()
+        for (loop in loops) {
+            for (b in loop.body) {
+                loopsByBlock.updateInplace(b, default = mutableListOf()) { add(loop.id) }
+            }
+        }
+        return MethodLoopsInformation(loops = loops, loopsByBlock = loopsByBlock)
+    }
+}
```
eupp, line 251: BTW, let's extract loop information construction (and all related utility functions) into separate file (or move it to `LoopInformation.kt`).

eupp (changes requested):
Could you please also rebase the branch on develop (specifically, only subset of commits need to be rebased).

jvm-agent/src/test/org/jetbrains/lincheck/jvm/agent/ControlFlowGraphLoopsTests.kt
eupp: Perhaps also a test like: `if (...) { while (...) { ... } }; doSomething()` to check for normal-exit edges to `doSomething()` both from or outside of the loop.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
dmitrii-artuhov: seems simplified enough now

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
dmitrii-artuhov: Added a check for reducability. By the way, in the one of the articles you've sent me, this term is described for the CFG and not its loops. The dominators algorithm cannot calculate loops in irreducable graphs at all

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
dmitrii-artuhov: removed from this PR

jvm-agent/src/test/org/jetbrains/lincheck/jvm/agent/ControlFlowGraphLoopsTests.kt
dmitrii-artuhov: added

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -97,20 +114,100 @@ class BasicBlockControlFlowGraph(
...
+                    info.loops.forEach {
+                        it.validateLoopEdgesInvariants()
+                    }
+                }
+            }
+            else {
+                _isReducible = false
+                // CFG is irreducible, our loop calculation algorithm will not work with it
+                // so we report that fact and don't compute any loop information
+                Logger.warn { "Irreducible CFG detected, loop information will not be computed:\n${toFormattedString()}" }
```
dmitrii-artuhov, line 173: Print method name to which CFG corresponds to, instead of the CFG itself 
https://youtrack.jetbrains.com/agiles/153-5670/current?issue=DR-464

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -83,6 +83,23 @@ class BasicBlockControlFlowGraph(
     var loopInfo: MethodLoopsInformation? = null
         private set
 
+    /**
+     * Whether this CFG is reducible.
+     */
+    val isReducible: Boolean
```
eupp: Let's just make it optional:
```
var isReducible: Boolean? = null
     private set
```

and return `null` in case when loop information has not yet been computed. 

And document that `isReducible` is known (!= null) only after `computeLoopInformation()` was called.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -141,3 +141,168 @@ class MethodLoopsInformation(
...
+                changed = true
+            }
+        }
+    }
+    return doms
+}
+
+/**
+ * **This algorithm cannot detect loops in irreducible graphs. It only produces reducible loops.
+ * Thus, expects reducible CFG.**
```
eupp: Let's remove `**` from the comments.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -141,3 +141,168 @@ class MethodLoopsInformation(
...
+    val loopsByBlock = mutableMapOf<BasicBlockIndex, MutableList<LoopId>>()
+    for (loop in loops) {
+        for (b in loop.body) {
+            loopsByBlock.updateInplace(b, default = mutableListOf()) { add(loop.id) }
+        }
+    }
+    return MethodLoopsInformation(loops = loops, loopsByBlock = loopsByBlock)
+}
+
+// == LOOPS PRETTY-PRINTING ==
```
eupp, line 330:
```suggestion
// == LOOPS PRETTY-PRINTING ==

```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -141,3 +141,168 @@ class MethodLoopsInformation(
     fun getLoopInfo(id: LoopId): LoopInformation? =
         loops.getOrNull(id)
 }
+
+// == LOOPS EXTRACTION FROM CFG ==
```
eupp, line 146:
```suggestion
// == LOOPS EXTRACTION FROM CFG ==

```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -97,20 +114,100 @@ class BasicBlockControlFlowGraph(
...
      * Computes loop-related information for this method.
      */
     fun computeLoopInformation(): MethodLoopsInformation {
         if (loopInfo == null) {
-            // TODO: this is a stub: loop detection is not implemented yet; the method returns empty sites.
-            loopInfo = MethodLoopsInformation().apply {
-                loops.forEach { it.validateLoopEdgesInvariants() }
+            // compute predecessors for all basic blocks
+            val allPredecessors: Array<MutableSet<BasicBlockIndex>> = Array(basicBlocks.size) { mutableSetOf() }
+            val normalPredecessors: Array<MutableSet<BasicBlockIndex>> = Array(basicBlocks.size) { mutableSetOf() }
```
eupp: It looks like predecessors calculation is shared between several stages of the algorithm. 
Let's then unify this. 

Let's make `allPredecessors` and  `normalPredecessors` properties of `BasicBlockControlFlowGraph` and override `addEdge` to just maintain these maps incrementally on each update. We probably do not want to add predecessors to `InstructionControlFlowGraph`, because on instruction-level cfg there are a lot more edges in the graph, and we end up storing too much unnecessary information.

It also then makes sense to: 
- rename `edgeMap` to `allSuccessors` (and maybe introduce `normalSuccessors`?);
- rename `EdgeMap` to `AdjacencyMap`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -97,20 +114,100 @@ class BasicBlockControlFlowGraph(
...
+            for (e in edges) {
+                val u = e.source
+                val v = e.target
+                allPredecessors[v].add(u)
+                if (e.label !is EdgeLabel.Exception) {
+                    normalPredecessors[v].add(u)
+                }
+            }
+            // compute loop information
+            val dominators = computeDominators(allPredecessors)
```
eupp: Same observation about dominators and back-edges: they are re-computed multiple times in different methods.

We can add `dominators` and `backEdges` properties directly to CFG, similarly how it is done for `loopInfo`. They should be initially `null`, and they should be initialized after a call to `computeLoopInformation()`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -97,20 +114,100 @@ class BasicBlockControlFlowGraph(
...
+     * CFG is called *reducible* if removal of its back edges leads to a graph that:
+     * * is acyclic (in terms of graph theory)
+     * * each basic block of CFG can be reached from the initial one
+     *
+     * The initial basic block in our case is the basic block at index 0.
+     * And reachability calculation can use both normal (except for back-edges) and exceptional edges.
+     *
+     * See also https://www.csd.uwo.ca/~mmorenom/CS447/Lectures/CodeOptimization.html/node6.html.
+     */
+    private fun isReducible(dominators: Array<Set<BasicBlockIndex>>): Boolean {
```
eupp: Let's move to `LoopInformation.kt` --- it is conceptually related to other methods that compute loop information.

eupp (changes requested):

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
dmitrii-artuhov: added all except for `normalSuccessors` -- there are no places where it is needed

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
dmitrii-artuhov: added `dominators` and `backEdges` fields

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
```diff
@@ -83,6 +84,54 @@ class BasicBlockControlFlowGraph(
...
+    /**
+     * Dominators of each basic block.
+     *
+     * When this field is `null` it means that [computeLoopInformation] method was not called on this CFG.
+     */
+    var dominators: Array<Set<BasicBlockIndex>>? = null
+        private set
+
+    /**
+     * Set of all back-edges in the CFG stored by their target block index.
```
eupp: Outdated comment, `backEdges` are stored as `Set`, not `Map`

eupp (approved):

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/BasicBlockControlFlowGraph.kt
dmitrii-artuhov: fixed

PR #852, opened 2025-10-22
Add option to enable "old" threading recording.

lev-serebryakov-jetbrains:
Add option not to trace all "old" threads (as was before), useful for parallel tests execution.


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderAgent.kt
```diff
@@ -32,6 +32,7 @@ internal object TraceRecorderAgent {
     const val ARGUMENT_FORMAT = "format"
     const val ARGUMENT_FOPTION = "formatOption"
     const val ARGUMENT_PACK = "pack"
+    const val ARGUMENT_TRACE_ALL_THREADS = "traceAllThreads"
```
dmitrii-artuhov: Let's change "trace" to "track" in symbols related to this new flag to match the "Injections.enableGlobalThreadsTracking":
- ARGUMENT_TRACK_ALL_THREADS
- trackAllThreads

dmitrii-artuhov (approved):

PR #854, opened 2025-10-22
Allow of attaching trace recorder to unknown class.

lev-serebryakov-jetbrains:
Sometimes only convenient way to attach Trace Recorder agent is to pass arguments to several JVM instances (i.e. via gradle script).

Some of these instances can don't have target class, and it is Ok in such cases.

Don't fail if target class is not found.


eupp:
@lev-serebryakov-jetbrains should we merge this?

PR #855, opened 2025-10-22
Loops injections via CFG

dmitrii-artuhov:
I have temporarily removed trace debugger test configs for JDK 11 and 21, because they were TLing for some unclear reason for me. I will dig into them later, for now just disabled

### DIFF
--- a/buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
+++ b/buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
@@ -27,7 +27,7 @@ private val projectsToTest = listOf(
     GithubProjectSnapshot(
         organization = "ivandev0",
         repositoryName = "TraceDebuggerExamples",
-        commitHash = "11ccb79cf39f570d650919a86783847c957eee34"
+        commitHash = "1923624a2ba2a3954e5c5c3504ec9e8233ffee40"
     ),
     GithubProjectSnapshot(
         organization = "JetBrains",
--- a/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/TraceDebuggerExamplesTraceRecorderIntegrationTest.kt
+++ b/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/TraceDebuggerExamplesTraceRecorderIntegrationTest.kt
@@ -313,4 +313,76 @@ class TraceDebuggerExamplesTraceRecorderIntegrationTest : AbstractTraceRecorderI
             gradleCommands = listOf(":test"),
         )
     }
+
+    @Test
+    fun `org_examples_integration_loops_ListBreakedForAndMapLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListBreakedForAndMapLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_ListBreakedForLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListBreakedForLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_ListContinuedForAndMapLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListContinuedForAndMapLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_ListContinuedForLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListContinuedForLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_ListForEachBreakedLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListForEachBreakedLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_ListForEachContinuedLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.ListForEachContinuedLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_SequenceBreakedForAndMapLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.SequenceBreakedForAndMapLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
+
+    @Test
+    fun `org_examples_integration_loops_SequenceContinuedForAndMapLoopRepresentationTest operation`() {
+        runGradleTest(
+            testClassName = "org.examples.integration.loops.SequenceContinuedForAndMapLoopRepresentationTest",
+            testMethodName = "operation",
+            gradleCommands = listOf(":test"),
+        )
+    }
 }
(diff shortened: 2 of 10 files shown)


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/MethodCallTransformerBase.kt
```diff
@@ -212,7 +212,8 @@ internal abstract class MethodCallTransformerBase(
         className == "java/util/Locale" ||
         className == "org/slf4j/helpers/Util" ||
         className == "java/util/Properties" ||
-        className == "java/lang/invoke/MethodHandles"
+        // TODO: perhaps should use `isIgnoredMethodHandleMethod` instead
+        className.startsWith("java/lang/invoke/")
```
dmitrii-artuhov: Why is this needed? Many lincheck tests fail because of it as some important classes are not instrumented

eupp: I don't know, it was taken from this https://github.com/JetBrains/lincheck/pull/800 PR. 
Let's revert this change to `develop`.

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -641,6 +641,24 @@ public static void onInlineMethodCallException(int methodId, Throwable t) {
...
+        getEventTracker().onLoopIteration(codeLocation, loopId);
+    }
+
+    /**
+     * Called on a normal (non-exceptional) exit from a loop body and
+     * at an exception handler entry that is reachable from within a loop body and lies outside it.
+     *
+     * @param isReachableFromOutsideLoop true if the handler can also be reached from outside the loop body;
+     *   false if it is exclusive to the loop body.
+     */
```
eupp, line 658: Update the documentation (can be called on both normal and exceptional exits), and also document `exception` parameter.

integration-test/trace-recorder/src/main/resources/integrationTestData/TraceDebuggerExamples/org.examples.integration.loops.WhileWithIfLoopRepresentationTest_operation.txt
```diff
@@ -0,0 +1,72 @@
+# Thread 1
+WhileWithIfLoopRepresentationTest.operation() at :0
+  escape = "START" at WhileWithIfLoopRepresentationTest.kt:17
+  i = 0 at WhileWithIfLoopRepresentationTest.kt:18
+  total = 0 at WhileWithIfLoopRepresentationTest.kt:19
+  loop(11 iterations) at WhileWithIfLoopRepresentationTest.kt:20
+    <iteration-1> at WhileWithIfLoopRepresentationTest.kt:20
```
eupp: Let's print iteration numbers without `-` sign:
```
    <iteration 1> at WhileWithIfLoopRepresentationTest.kt:20
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -652,11 +729,73 @@ class TraceCollectingEventTracker(
...
+            val tracePoint = TRLoopTracePoint(
+                threadId = threadData.threadId,
+                codeLocationId = codeLocation,
+                loopId = loopId,
+            )
+            strategy.tracePointCreated(threadData.currentTopTracePoint(), tracePoint)
+            threadData.enterLoop(tracePoint)
+        }
+
+        val currentLoopTracePoint = threadData.currentLoopTracePoint().ensureNotNull {
```
eupp: This `ensureNotNull` is no longer needed (you create loop trace point just above it).

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -652,11 +729,73 @@ class TraceCollectingEventTracker(
...
+            "Unexpected loop iteration: no loop trace point found"
+        }
+        // complete previous iteration, if any
+        threadData.currentLoopIterationTracePoint()?.also { previousIteration ->
+            strategy.completeContainerTracePoint(Thread.currentThread(), previousIteration)
+        }
```
eupp, line 761: You should also close last loop iteration when you create new loop trace point (`if (loopId != threadData.currentLoopTracePoint()?.loopId)` branch above).

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -831,4 +970,23 @@ class TraceCollectingEventTracker(
...
         return analysisProfile.getAnalysisSectionFor(ownerName, methodName)
     }
+
+    /**
+     * Method takes the currently open loop trace point and finishes
+     * its last open iteration (if it exists) and then the loop itself.
+     *
+     * Note: assumes there is an open loop trace point.
+     */
+    private fun exitCurrentLoop(threadData: ThreadData) {
```
eupp: Please move this method near other loop-related methods above.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderMethodTransformer.kt
```diff
@@ -48,20 +48,23 @@ internal class TraceRecorderMethodTransformer(
...
     override fun onMethodExit(opcode: Int) {
-         invokeStatic(TraceRecorderInjections::stopTraceRecorderAndDumpTrace)
-         super.onMethodExit(opcode)
+        super.onMethodExit(opcode)
+        if (opcode == ATHROW) return
+        invokeStatic(TraceRecorderInjections::stopTraceRecorderAndDumpTrace)
     }
 
     override fun visitMaxs(maxStack: Int, maxLocals: Int) {
+        // TODO: repair exception handling here
```
eupp: Already repaired?

eupp (approved):
@dmitrii-artuhov  I've put "Approve" status as to not block the merge while I'll be away. 
But please fix all the remaining comments before merging.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
dmitrii-artuhov: Why so?
What if the new loop is just part of the current iteration of the containing loop:
```
for (i) {
  foo()
  for (j) {
     a()
     b()
  }
  bar()
}
```

dmitrii-artuhov:
@zhelenskiy there are tests for break/continue, nested loops, see the TraceDebuggerExamples project and representation files inside this PR.
But there are no tests for repeat, break/continue by label, repeat inside loop, and break/continue of the wrapping loop

zhelenskiy:
@dmitrii-artuhov can you add them?

dmitrii-artuhov:
@zhelenskiy there is a problem with breaks/continues of outer loops now from inner ones: the outer cycle is getting slices in two instead of being a single one. I am inverstigating the cause  


Also note, that I have updated the description the PR: I've temporarily removed 2 test configs from the CI

dmitrii-artuhov:
@zhelenskiy fixed the bug + added 6 more tests, see the https://github.com/ivandev0/TraceDebuggerExamples/tree/master/src/test/kotlin/org/examples/integration/loops and their outputs in this PR

- BreakedNestedRepeatLoopRepresentationTest
- BreakedOuterForLoopRepresentationTest
- ContinuedNestedRepeatLoopRepresentationTest
- ContinuedOuterForLoopRepresentationTest
- NestedRepeatLoopRepresentationTest
- RepeatLoopRepresentationTest

zhelenskiy:
Add also repeat RepeatBreakedNestedLoopRepresentationTest, RepeatContnuedNestedLoopRepresentationTest, also add all the same for coroutines, for iterating over object collections, over sequences, over collections and sequences `.withIndex()`.

eupp:
> all the same for coroutines

Coroutines are out of scope for this PR.

> for iterating over object collections, over sequences, over collections and sequences .withIndex()

Adding couple of tests for collections/sequences is a good idea. But there is little sense to add all possible combinations of collections + break/continue/etc. 
1. If we will consider combination of every feature with any other possible feature, we will get exponential blow-up on the number of tests. 
2. On bytecode level (where we detect loops), all these features are desugared into similar representation, so break/continue will work similarly for e.g., collection loops as for regular loops.

dmitrii-artuhov:
I will add a couple of collections tests in TraceDebuggerExamples and then merge this PR

PR #860, opened 2025-11-04
Revert removal of `alreadyStopped` flag

dmitrii-artuhov:
Insertion of `TraceDebuggerInjections::startTraceRecorder` method call still allows for its double invocation even though there was an attempt to fix the instrumentation in #855.

I revert that change, so that when exception is thrown from the `startTraceRecorder` we see it logged in the console instead of some unrelated error caused by second invocation of `startTraceRecorder`


lev-serebryakov-jetbrains (approved):
LGTM

PR #861, opened 2025-11-04
Fix bug with completing threads with loops

dmitrii-artuhov:
This PR makes sure that all open container trace points are closed when daemon-threads (those that keep executing while Main is finishing) are completed. There was an error related to unbalanced container tracepoints footers closing without this change.

I have also added a reproduction test, which now passes with no errors, and it has consistent trace output

### DIFF
--- a/buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
+++ b/buildSrc/src/main/kotlin/TraceAgentIntegrationTestsTasks.kt
@@ -27,7 +27,7 @@ private val projectsToTest = listOf(
     GithubProjectSnapshot(
         organization = "ivandev0",
         repositoryName = "TraceDebuggerExamples",
-        commitHash = "1923624a2ba2a3954e5c5c3504ec9e8233ffee40"
+        commitHash = "4eb6b032f3ced65f432eaf77016cf686cdb8084d"
     ),
     GithubProjectSnapshot(
         organization = "JetBrains",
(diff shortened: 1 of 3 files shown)


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -879,11 +881,26 @@ class TraceCollectingEventTracker(
         spinner.spinWaitUntil { !threadDescriptor.isInsideInjectedCode }
         // now, we are sure that another thread has finished its injected code
         // and will not attempt to execute anything else, because we disabled analysis in it
-        val stackFrames = threadData.getStack().asReversed()
-        stackFrames.forEach { frame ->
+        val stack = threadData.getStack()
+        while (stack.isNotEmpty()) {
```
lev-serebryakov-jetbrains, line 885: Maybe change condition to `stack.size() > 1` and get rid of `if` and `break` at the end of the loop?

dmitrii-artuhov: When stask size is 1, I still need to complete the loops from that single method call (loops could be inside the `Thread.run` invocation itself) and also complete the method call as well.
The only difference is that in this case I do not pop the method call from stack, but all other actions should be performed anyway.

Getting rid of `if` and changing `while` condition will change the semantics

lev-serebryakov-jetbrains: Oh, I see.

PR #863, opened 2025-11-04
Ensure that dynamically loaded classes are transformed

zhelenskiy:


ivandev0:
I am not sure that this is the approach that we want to take. It is better to use eager initialization mode that will fix most of the issues. We just need to fix the performance of it.

ivandev0 (changes requested):
See above comment.

PR #864, opened 2025-11-05
[JBRes-6525] Implement loops reachability analysis

dmitrii-artuhov:
This PR implements the following:
- Now when loops are extracted from CFG of method, they first will be sorted from outer-most to inner-most (Meaning that if some loop is outer for some other inner loop, then it will go first. The ordering between unrelated loops is not guaranteed). This feature is needed to assign smaller ids to outer loops, and then insert `afterLoopExit` hook from bigger loop ids to smaller ones in case they should be inserted at the same loop-exit location.
- Reachability analysis: for each loop we just check that some exit block has no incoming edges from outside this loop. Later this fact will checked in runtime: if `afterLoopExit` hook is called with the some loop id and this hook is only reachable from this loop id, then the same loop id must be on our stored stack. In case if not, we will detect an error and log it to the terminal

### DIFF
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
@@ -245,7 +245,7 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
     val backEdgesByHeader = backEdges.groupBy { it.target }
     val loops = mutableListOf<LoopInformation>()
     var nextLoopId = 0
-    val loopBodiesByHeader = mutableListOf<Pair<BasicBlockIndex, Set<BasicBlockIndex>>>()
+    val loopBodiesWithHeader = mutableListOf<Pair<BasicBlockIndex, Set<BasicBlockIndex>>>()
     for ((h, adjacentBackEdges) in backEdgesByHeader) {
         val body = mutableSetOf<BasicBlockIndex>()
         body.add(h)
@@ -262,12 +262,12 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
                 }
             }
         }
-        loopBodiesByHeader.add(h to body.toSet())
+        loopBodiesWithHeader.add(h to body.toSet())
     }
 
     // Sort loop bodies by containment and header values, so that for any two loops a and b we could say that
     // if b is an inner loop of a, then id(a) < id(b), so "outer" loops have smaller ids than "inner" loops
-    loopBodiesByHeader.sortWith { a, b ->
+    loopBodiesWithHeader.sortWith { a, b ->
         val aBody = a.second
         val bBody = b.second
         // the "outer" loops will come before their "inner" loops
@@ -276,7 +276,7 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
         else a.first.compareTo(b.first)
     }
 
-    for ((h, body) in loopBodiesByHeader) {
+    for ((h, body) in loopBodiesWithHeader) {
         // Normal exits: edges from body to outside, non-exception
         val normalExits = buildSet {
             for (e in normalEdges) {


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -782,21 +782,17 @@ class TraceCollectingEventTracker(
...
+        }
+        if (currentLoopTracePoint?.loopId == loopId) {
             exitCurrentLoop(thread, threadData)
-        } else {
-            if (threadData.currentLoopTracePoint()?.loopId != loopId) return
-            do {
-                val currentLoopTracePoint = threadData.currentLoopTracePoint() ?: break
-                exitCurrentLoop(thread, threadData)
-            }
-            while (currentLoopTracePoint.loopId != loopId)
```
dmitrii-artuhov, line 799: I have no idea why I wrote such code 🤷‍♂️ It is semantically equivalent to 
```
if (threadData.currentLoopTracePoint()?.loopId != loopId) {
  exitCurrentLoop(thread, threadData)
}
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -236,6 +244,7 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
     val backEdgesByHeader = backEdges.groupBy { it.target }
     val loops = mutableListOf<LoopInformation>()
     var nextLoopId = 0
+    val loopBodiesByHeader = mutableListOf<Pair<BasicBlockIndex, Set<BasicBlockIndex>>>()
```
eupp: Why not `Map<BasicBlockIndex, Set<BasicBlockIndex>>`?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -266,14 +290,21 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
                 }
             }
         }
+        // All exits that are only reachable from the loop body
+        val exclusiveExits = sequence {
```
eupp: Why you decided to use `sequence` here, is there any reason?
In general, `sequence` builder is not cheap, unless you really need lazy elements processing better to use regular lists.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -207,5 +210,36 @@ private fun BasicBlockControlFlowGraph.computeExceptionExitSites(
...
+ */
+private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
+    insnIndexRemapping: IntArray,
+    loopInfo: MethodLoopsInformation
+): Map<InstructionIndex, Set<Int>> {
+    if (!loopInfo.hasLoops()) return emptyMap()
+    val cfg = this
+    val result = mutableMapOf<InstructionIndex, MutableSet<Int>>()
+    for (loop in loopInfo.loops) {
+        val exitBlocks = sequence {
```
eupp: Same question here about `sequence`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
```diff
@@ -252,6 +261,21 @@ internal fun BasicBlockControlFlowGraph.computeLoopsFromDominators(): MethodLoop
                 }
             }
         }
+        loopBodiesByHeader.add(h to body.toSet())
+    }
+
+    // Sort loop bodies by containment and header values, so that for any two loops a and b we could say that
+    // if b is an inner loop of a, then id(a) < id(b), so "outer" loops have smaller ids than "inner" loops
```
eupp, line 269: Could you also please add information about this invariant (loop ids and their nestedness) to `MethodLoopsInformation` documentation.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -99,9 +103,7 @@ internal class LoopTransformer(
                 original = {},
                 instrumented = {
                     for (loopId in loopIds) {
-                        // conservative default without exclusivity analysis
-                        // TODO: implement exclusivity analysis and use it here
-                        val isReachableFromOutsideLoop = true
+                        val isReachableFromOutsideLoop = opcodesReachableFromOutsideLoops[nonPhonyIndex]?.contains(loopId) ?: true
```
eupp, line 106: Could you please unify naming of this flag across codebase?
In some places it is called `isReachableFromOutsideLoop` and in some other places `canEnterFromOutsideLoop`.

eupp (changes requested):

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
dmitrii-artuhov: Map's `toSortedMap` only accepts a comparator for keys, but I need to sort the collection by values as well (the set's containment). I would need to call `map.entries.sortedWith(...)`, which is basically the same as having a list of pairs

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
dmitrii-artuhov: I want to have an iterator over 2 collections with different stored types, but I don't want to create a new collection from them, I just need an iterator. That is what I am trying to achieve here

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
dmitrii-artuhov: to have an iterator instead of creating new collections

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
dmitrii-artuhov: Alternatively, I could have something like this:
```kotlin
val exclusiveExits = mutableSetOf<BasicBlockIndex>().apply {
  normalExits.map { it.target }.filterTo(this) { isExclusiveExitFromLoopBody(it, body) }
  exceptionalExitHandlers.filterTo(this) { isExclusiveExitFromLoopBody(it, body) }
}
```

I think sequence variant is cleaner

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
eupp: >  I would need to call map.entries.sortedWith(...), 

What's wrong with `map.entries.sortedWith(...)` ?)

> which is basically the same as having a list of pairs

Yeap, any map can be represented as a list of pairs. 
But map allows efficient lookup by key (this is what name of variable `loopBodiesByHeader` suggests). Even though lookup is currently not needed in this code, it might be convenient to have for future. 

In general, map looks kinda more natural here.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
eupp: You can use either what you proposed above, or this:
```kotlin
(normalExits.asSequence().map { it.target } + exceptionalExitHandlers.asSequence())
    .filterTo(...) { ... }
```

`sequence` builder under the hood is implemented via the same machinery as coroutines (a state machine) --- it is non-trivial and not cheap. `asSequence()` is just one allocation wrapping `Iterable` into `Sequence`.

dmitrii-artuhov: yeah, this is what I need, I will rewrite then

eupp: But for small collections typically there is no difference between lists vs sequences.

eupp: - https://chrisbanes.me/posts/use-sequence/
- https://youtrack.jetbrains.com/issue/KT-75201

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/analysis/controlflow/LoopInformation.kt
dmitrii-artuhov: `map.entries` cannot be sorted inplace, it only exposes `sortedWith` api which creates a new collection. I suggest just renaming the variable to `loopBodiesWithHeader` so there are no associations with map. Also I see no reason to make it a map, because it is a local variable which is only required for proper assignment of loops id. It will not be used any further

eupp (approved):

Issue #850, opened 2025-10-13
Using 3.3 with maven


adamw:
When trying to add lincheck as a test dependency using maven:

```
  <properties>
    <kotlin.compiler.jvmTarget>21</kotlin.compiler.jvmTarget>
  </properties>

  <dependencies>
    <dependency>
      <groupId>org.jetbrains.lincheck</groupId>
      <artifactId>lincheck</artifactId>
      <version>3.3.2</version>
      <scope>test</scope>
    </dependency>
    <dependency>
      <groupId>org.junit.jupiter</groupId>
      <artifactId>junit-jupiter</artifactId>
      <scope>test</scope>
    </dependency>
  </dependencies>

  <build>
    <testSourceDirectory>src/test/kotlin</testSourceDirectory>

    <plugins>
      <plugin>
        <groupId>org.jetbrains.kotlin</groupId>
        <artifactId>kotlin-maven-plugin</artifactId>
        <version>2.2.20</version>
        <executions>
          <execution>
            <id>compile</id>
            <phase>compile</phase>
            <goals>
              <goal>compile</goal>
            </goals>
          </execution>
          <execution>
            <id>test-compile</id>
            <phase>test-compile</phase>
            <goals>
              <goal>test-compile</goal>
            </goals>
            <configuration>
              <sourceDirs>
                <sourceDir>${project.basedir}/src/test/kotlin</sourceDir>
              </sourceDirs>
            </configuration>
          </execution>
        </executions>
        <configuration>
          <jvmTarget>${kotlin.compiler.jvmTarget}</jvmTarget>
        </configuration>
      </plugin>
    </plugins>
  </build>
```

I'm getting the following error:

```
[ERROR] dependency: org.jetbrains.kotlin:kotlin-stdlib-common:jar:2.2.0 (test)
[ERROR]         Could not find artifact org.jetbrains.kotlin:kotlin-stdlib-common:jar:2.2.0 in central (https://repo.maven.apache.org/maven2)
```

and indeed, there's no jar for kotlin-stdlib-common in the repository: https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-common/2.2.0/

Is it possible to use lincheck with maven?

ilya-g:
[...]

PR #866, opened 2025-11-07
Simplify Gradle build files

ivandev0:


build.gradle.kts
```diff
@@ -113,9 +112,7 @@ sourceSets {
         implementation(project(":jvm-agent"))
         implementation(project(":trace"))
 
-        api("org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion")
-        api("org.jetbrains.kotlin:kotlin-stdlib-common:$kotlinVersion")
```
eupp, line 117: Am I right this will also fix this problem?: https://github.com/JetBrains/lincheck/issues/850#issuecomment-3403447026

ivandev0: Looks like it! I am not exactly sure but we can check

PR #869, opened 2025-11-07
[JBRes-6756] Implement "null" output strategy for TR benchmarking

lev-serebryakov-jetbrains:
This trategy is named `null` and simply throws away all new tracepoints (do nothing).


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -182,6 +187,8 @@ fun parseOutputMode(outputMode: String?, outputOption: String?): TraceCollectorM
         } else {
             return TraceCollectorMode.TEXT
         }
+    } else if ("null".equals(outputMode, true)) {
```
eupp:
```suggestion
    } else if ("null".equals(outputMode, ignoreCase = true)) {
```

is a bit cleaner. 

Same for other `if` branches here, e.g. `"dump".startsWith(outputOption, ignoreCase = true)`

PR #873, opened 2025-11-09
Implement background I/O for streaming trace recorder.

lev-serebryakov-jetbrains:
Now all block saving for binary format in streaming mode is performing in separate thread.

Also, this strategy implements its own `DataOutput` as `DataOutputStream` wrapper is very (5-10x) slower than `ByteBuffer` methods.

"Collect and dump" strategy still uses `DataOutputStream`, as loading still use `DataInputStream`. It should be optimized separately.

These changes decrease stress benchmark from 40 to 27-28 seconds.


trace/src/main/org/jetbrains/lincheck/trace/Serialization.kt
```diff
@@ -195,8 +197,9 @@ internal interface ContextSavingState {
 private sealed class TraceWriterBase(
     private val context: TraceContext,
     private val contextState: ContextSavingState,
-    protected val data: DataOutputStream
-): TraceWriter, DataOutput by data {
+    protected val dataStream: OutputStream,
```
ndkoval: Please leave a comment explaining why this separation is necessary and what dataStream and dataOutput are responsible for.

trace/src/main/org/jetbrains/lincheck/trace/Serialization.kt
```diff
@@ -774,6 +785,128 @@ class NullTraceCollecting(private val context: TraceContext): TraceCollectingStr
     }
 }
 
+private class FileStreamingThread(
```
ndkoval: Consider splitting this file into multiple ones for easier navigation

ndkoval (approved):
Nice work! I am approving the PR but leaving a couple of minor issues.

ndkoval:
> These changes decrease stress benchmark from 40 to 27-28 seconds.

@lev-serebryakov-jetbrains, I remember different numbers from our sync. Can you please clarify?

lev-serebryakov-jetbrains:
> > These changes decrease stress benchmark from 40 to 27-28 seconds.
> 
> @lev-serebryakov-jetbrains, I remember different numbers from our sync. Can you please clarify?

It is conservative numbers, they differ from run to run, I suspect my laptop power management, as they are stable for several runs in a row but can be smaller after a hour or two. I can reproduce both numbers with some luck. Need more tests on other hardware (not-heat-dissipation-restricted).

trace/src/main/org/jetbrains/lincheck/trace/Serialization.kt
lev-serebryakov-jetbrains: I've thought about it already.

lev-serebryakov-jetbrains:
> Nice work! I am approving the PR but leaving a couple of minor issues.

I've fixed these issues:  new comment & split huge `Serialization.kt` into logical parts which are much more manageable.

PR #879, opened 2025-11-12
Clean-up global event tracker logic

eupp:
Clean-ups the event tracker logic. 

- Introduce two modes of event tracking: `GLOBAL` and `THREAD_LOCAL`. 
- Simplifies event tracking querying logic based on used mode.

Also:
- Renames thread tracking injections to match Java's terminology (use `start`/`run` names instead of fork/start).

### DIFF
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformers.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformers.kt
@@ -45,7 +45,7 @@ internal class ThreadRunTransformer(
 
     override fun visitCode() = adapter.run {
         super.visitCode()
-        if (configuration.trackThreadRun && isThreadRunMethod(methodName, descriptor)) {
+        if (isThreadRunMethod(methodName, descriptor)) {
             // STACK: <empty>
             visitTryCatchBlock(runMethodTryBlockStart, runMethodTryBlockEnd, runMethodCatchBlock, null)
             visitLabel(runMethodTryBlockStart)
@@ -56,7 +56,7 @@ internal class ThreadRunTransformer(
     }
 
     override fun visitInsn(opcode: Int) = adapter.run {
-        if (configuration.trackThreadRun && isThreadRunMethod(methodName, descriptor) && opcode == Opcodes.RETURN) {
+        if (isThreadRunMethod(methodName, descriptor) && opcode == Opcodes.RETURN) {
             // STACK: <empty>
             invokeStatic(Injections::afterThreadRunReturn)
             // STACK: <empty>
@@ -65,7 +65,7 @@ internal class ThreadRunTransformer(
     }
 
     override fun visitMaxs(maxStack: Int, maxLocals: Int) = adapter.run {
-        if (configuration.trackThreadRun && isThreadRunMethod(methodName, descriptor)) {
+        if (isThreadRunMethod(methodName, descriptor)) {
             visitLabel(runMethodTryBlockEnd)
             visitLabel(runMethodCatchBlock)
             // STACK: exception


bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -141,32 +253,110 @@ public static void leaveIgnoredSection() {
...
+     * @param eventTracker the event tracker to associate with the thread.
+     * @return the thread descriptor associated with the given thread.
+     */
+    private static ThreadDescriptor registerRunningThread(EventTracker eventTracker, Thread thread) {
+        ThreadDescriptor descriptor = registerThread(eventTracker, thread);
+        ThreadDescriptor.setCurrentThreadDescriptor(descriptor);
+        eventTracker.registerRunningThread(thread, descriptor);
+        return descriptor;
+    }
+
```
lev-serebryakov-jetbrains, line 336: Why does these two methods (`registerCurrentThread` and `registerRunningThread`) differs? Maybe, clarify in comments?

eupp: - `registerCurrentThread` is part of public API that can be used to register current thread. 

- `registerRunningThread` is internal API, it is used to register already running thread, when this thread was started before global event tracking was enabled. Unlike `registerCurrentThread` it also calls `EventTracker.registerRunningThread` --- its implementation in trace recorder attempts to reconstruct missing stack frames from methods called earlier (before recording has started). 

I expanded the docs a bit, hope it is clearer now.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
```diff
@@ -73,29 +73,30 @@ object TraceRecorder {
...
+            ?: Injections.registerCurrentThread(eventTracker)
 
-        eventTracker!!.enableTrace()
-        desc.enableAnalysis()
         if (trackAllThreads) {
-            Injections.enableGlobalThreadsTracking(eventTracker)
+            Injections.enableGlobalEventTracking(eventTracker)
+        } else {
+            Injections.enableThreadLocalEventTracking()
+            ThreadDescriptor.setCurrentThreadAsRoot(descriptor)
```
ndkoval: Why do you need it for Trace Recorder?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformer.kt
```diff
@@ -47,15 +47,15 @@ internal class ThreadTransformer(
             // STACK: <empty>
             loadThis()
             // STACK: forkedThread
-            invokeStatic(Injections::beforeThreadFork)
+            invokeStatic(Injections::beforeThreadStart)
```
ndkoval: ThreadTransformer should no longer be used for Trace Recorder

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
eupp: @lev-serebryakov-jetbrains added an option to disable global events tracking some time ago. 
@lev-serebryakov-jetbrains what was the motivation back then?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformer.kt
eupp: Instrumentation of `Thread::run()` method (`beforeThreadRun` and `afterThreadRunReturn`/`afterThreadRunException`) is still needed to properly initialize/clean-up new thread. This is different from `registerRunningThread` that is used to register already running thread. This function tries to reconstruct missing stack trace frames that were on the thread before tracing started. This is not needed for newly started thread. 

I can, however, minimize the instrumentation for Trace Recorder, leaving only `Thread::run()` instrumentation.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
ndkoval: I am strongly against having multiple modes when it is not necessary

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformer.kt
ndkoval: Let's minimize it, please

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
lev-serebryakov-jetbrains: Motivation was to enable tracing 1 test but which is run in parallel with other tests. Ir allowed me to trace flacky tests which are flacky only with parallel run. Without it tests runs forever.

eupp: So what should I do? Remove this mode from Trace Recorder and remove `trackAllThreads` flag?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformer.kt
eupp: Implemented.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
lev-serebryakov-jetbrains: I think, it is useful at least for internal usage.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -299,20 +299,20 @@ class TraceCollectingEventTracker(
...
-        val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: return
+
+        // Create new thread data
+        val threadData = ThreadData(threads.size)
         val thread = Thread.currentThread()
+        // Register thread data
+        threadDescriptor.eventTrackerData = threadData
+        threads[thread] = threadData
+
         // just like in `registerRunningThread` we first need to enable analysis
```
lev-serebryakov-jetbrains, line 300: If we don't have `threadData` without stack frame anymore (as it is always created with stack frame), we don't need `ThreadData.startThread()` and this flag at all.

eupp: Ok, I removed it.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
eupp: Per @ndkoval request I removed the flag.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -110,14 +112,27 @@ class TransformationConfiguration(
...
             trackAllFieldsWrites && trackArrayElementWrites
-
         set(value) {
             trackAllFieldsReads = value
             trackArrayElementReads = value
             trackAllFieldsWrites = value
             trackArrayElementWrites = value
         }
 
+    val trackThreadsOperations: Boolean
```
ndkoval, line 122: Do you still need it?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformers.kt
```diff
@@ -26,73 +26,99 @@ import sun.nio.ch.lincheck.*
...
         }
         super.visitInsn(opcode)
     }
 
     override fun visitMaxs(maxStack: Int, maxLocals: Int) = adapter.run {
-        // we only need to handle `Thread::run()` method, exit early otherwise
-        if (!isThreadRunMethod(methodName, desc)) {
-            super.visitMaxs(maxStack, maxLocals)
-            return
+        if (configuration.trackThreadRun && isThreadRunMethod(methodName, descriptor)) {
```
ndkoval: Shouldn't we always track Thread.run?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/ThreadTransformers.kt
```diff
@@ -26,73 +26,99 @@ import sun.nio.ch.lincheck.*
...
+            visitLabel(runMethodTryBlockEnd)
+            visitLabel(runMethodCatchBlock)
+            // STACK: exception
+            invokeStatic(Injections::afterThreadRunException)
+            // STACK: <empty>
+
+            // Notify that the thread has finished.
+            // TODO: currently does not work, because `ManagedStrategy::onThreadFinish`
```
ndkoval, line 76: Should we solve this TODO?

ndkoval: Oh, I see, it was here before

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -314,7 +331,7 @@ object TraceDebuggerDefaultTransformationProfile : TransformationProfile {
             trackMethodCalls = true
             trackInlineMethodCalls = true
 
-            trackThreadsOperations = true
+            trackAllThreadsOperations = true
```
ndkoval, line 334: Let's remove Trace Debugger from the repo in a separate PR? :)

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -133,7 +148,9 @@ internal fun TransformationConfiguration.shouldApplyVisitor(visitorClass: Class<
         MethodCallTransformerBase::class.java -> trackMethodCalls
         InlineMethodCallTransformer::class.java -> trackInlineMethodCalls
 
-        ThreadTransformer::class.java -> trackThreadsOperations
+        ThreadRunTransformer::class.java -> trackThreadRun
+        ThreadStartJoinTransformer::class.java -> trackThreadStart || trackThreadJoin
```
ndkoval, line 152: Does it make sense to track only one of the events?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
eupp: Let's keep it for consistency with other flags.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
eupp: Let's decide what to do with it separately from this PR : )

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
eupp: Let's keep two flags for completeness and consistency with other flags.

PR #882, opened 2025-11-12
[JBRes-6878] Multiple concurrent calls to tracing method lead to error

lev-serebryakov-jetbrains:
Add protection from multiple calls with atomic counter.

### DIFF
--- a/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
+++ b/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
@@ -44,8 +44,9 @@ object TraceRecorder {
     private var eventTracker: TraceCollectingEventTracker? = null
 
     private val installCount = AtomicInteger(0)
+
     @Volatile
-    private var traceStarter: Thread? = null
+    private var traceStarterThread: Thread? = null
 
     fun installAndStartTrace(
         className: String,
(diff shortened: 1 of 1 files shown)


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
```diff
@@ -41,6 +43,10 @@ import sun.nio.ch.lincheck.ThreadDescriptor
 object TraceRecorder {
     private var eventTracker: TraceCollectingEventTracker? = null
 
+    private val installCount = AtomicInteger(0)
+    @Volatile
+    private var traceStarter: Thread? = null
```
eupp:
```suggestion
    private val installCount = AtomicInteger(0)
    
    @Volatile
    private var traceStarterThread: Thread? = null
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
```diff
@@ -50,6 +56,12 @@ object TraceRecorder {
         pack: Boolean,
         trackAllThreads: Boolean
     ) {
+        val startedCount = installCount.incrementAndGet()
+        if (startedCount > 1) {
+            Logger.debug { "Trace recorder has been started multiple times: $startedCount (${Thread.currentThread().name})" }
+            return
+        }
+        Logger.debug { "Trace recorder has been started first time: $startedCount (${Thread.currentThread().name})" }
```
eupp: Let's just log unconditionally under `INFO`:
```kotlin
Logger.info { "Trace recording has been started from $className::$methodName in thread ${Thread.currentThread().name} (installCount=${installCount})" }
```

Because starting recording multiple times under global tracing should be considered normal situation.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorder.kt
```diff
@@ -75,11 +88,31 @@ object TraceRecorder {
     }
 
     fun finishTraceAndDumpResults() {
+        val startedCount = installCount.decrementAndGet()
+        if (startedCount > 0) {
+            Logger.debug { "Recording has been stopped less times than started: $startedCount (${Thread.currentThread().name})" }
```
eupp, line 94: Same here:
```kotlin
Logger.info { "Trace recording has been stoped from $className::$methodName in thread ${Thread.currentThread().name} (installCount=${installCount})" }
```

eupp (changes requested):

PR #885, opened 2025-11-13
Add TR integration tests with enabled kover and filtering of kover instrumentation via `TracePostprocessor`

dmitrii-artuhov:
- Added a submodule to the TraceDebuggerExamples with turned-on kover. Also I've added 2 tests there and check their representation in this PR, so that no kover instrumentation appears
- Added a kotlinx-datetime test which used to fail because of the kover, now it does not. I did not add any representation tests for datetime because datetime uses stress testing with a lot of iterations (the output file results in 110k+ lines) and they are non-deterministic as well

### DIFF
--- a/integration-test/common/src/main/AbstractTraceIntegrationTest.kt
+++ b/integration-test/common/src/main/AbstractTraceIntegrationTest.kt
@@ -58,6 +58,9 @@ abstract class AbstractTraceIntegrationTest {
                         val jvmArgs = options.jvmArgs?.toMutableList() ?: mutableListOf()
                         jvmArgs.addAll(listOf(${extraJvmArgs.joinToString(", ") { "\"$it\"" }}))
                         jvmArgs.add("-javaagent:$pathToFatJar=$agentArgs")
+                        // We need to append our agent args the latest, so it comes
+                        // after any user-specified args (including other agents).
+                        // The only reliable way to do this is via `jvmArgumentProviders`
                         options.jvmArgumentProviders.add(CommandLineArgumentProvider { jvmArgs })
                     }
                 }


trace/src/main/org/jetbrains/lincheck/trace/TracePostprocessor.kt
```diff
@@ -237,6 +239,26 @@ object CompressingPostprocessor : TracePostprocessor {
         return this
     }
 
+    /**
+     * Removes accesses to local variable `__$coverage_local$__` inserted by intellij coverage agent (or kover plugin).
+     *
+     * The accesses might contain array writes and reads `__$coverage_local$__[idx] = value` and assignments
+     * `__$coverage_local$__ = array`.
+     */
+    private fun TRTracePoint.removeCoverageInstructions(reader: LazyTraceReader): TRTracePoint? {
```
ndkoval, line 248: This part changes the logic; please adjust the PR title accordingly.

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -58,7 +58,7 @@ abstract class AbstractTraceIntegrationTest {
                         val jvmArgs = options.jvmArgs?.toMutableList() ?: mutableListOf()
                         jvmArgs.addAll(listOf(${extraJvmArgs.joinToString(", ") { "\"$it\"" }}))
                         jvmArgs.add("-javaagent:$pathToFatJar=$agentArgs")
-                        options.jvmArgs = jvmArgs
+                        options.jvmArgumentProviders.add(CommandLineArgumentProvider { jvmArgs })
```
eupp, line 64: Please add a comment to explain why `jvmArgumentProviders` is required.

PR #887, opened 2025-11-15
Collect bytecode instrumentation statistics

eupp:


eupp:
Example of the printed transformation statistics:
```
==== Transformation statistics ====
Total transformed classes: 9462
Total transformed methods: 104767
Total transformation time: 6533 ms
Average transformation time per class: 1 ms
Average class size:
	before: 6639 B
	after: 8695 B
Average method instructions count:
	before: 19
	after: 59
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationStatistics.kt
```diff
@@ -0,0 +1,315 @@
...
+    val averageMethodInstructionsCountBefore: Double,
+    val averageMethodInstructionsCountAfter: Double,
+
+    val totalTransformationTimeNanos: Long,
+    val averageTransformationTimeNanos: Double,
+)
+
+fun TransformationStatistics.writeTo(writer: Writer) {
+    val avgClassDelta = averageClassBytesSizeAfter - averageClassBytesSizeBefore
+    val avgInsnDelta = averageMethodInstructionsCountAfter - averageMethodInstructionsCountBefore
```
dmitrii-artuhov: These are not used, could you remove them

common/src/main/org/jetbrains/lincheck/util/Collections.kt
```diff
@@ -186,6 +186,13 @@ fun <T> List<T>.isSortedWith(comparator: Comparator<in T>): Boolean {
     return true
 }
 
+/**
+ * Returns the average value of the elements in the list, or `null` if the list is empty.
+ */
+fun List<Long>.averageOrNull(): Double? {
```
dmitrii-artuhov, line 192: This could be `averageOrDefault(() -> Double): Double` in order to get rid of multiple `?:` operators in the statistics file

dmitrii-artuhov: this one is up-to-you

eupp: Kotlin stdlib usually defines only `orNull` version, so let's stick to that.

PR #888, opened 2025-11-17
JBRes-6905 [Bug] java.io.IOException: Cannot read ObjectKind: unknown ordinal -1

lev-serebryakov-jetbrains:
Fix seek() implementation.

Previous implementation can lead to skipped buffer refill as a result of two consecutive seek() calls without any other stream operations in between.

Also, small cosmetic fixes while I'm here.

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/Streams.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/Streams.kt
@@ -87,7 +87,7 @@ internal class SeekableChannelBufferedInputStream(
         if (pos < bufferStartPosition || pos >= bufferStartPosition + buffer.limit()) {
             // It will trigger re-fill on next read(), as remaining() will return 0.
             buffer.limit(0)
-            // Refill will start form given position
+            // Refill will start from the given position
             bufferStartPosition = pos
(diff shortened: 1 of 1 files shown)


trace/src/main/org/jetbrains/lincheck/trace/Streams.kt
```diff
@@ -84,14 +84,12 @@ internal class SeekableChannelBufferedInputStream(
...
 
-        if (pos < bufferStartPosition || pos >= bufferStartPosition + buffer.capacity()) {
-            buffer.clear()
-            // Mark as empty for reading
-            buffer.position(buffer.capacity())
-
+        if (pos < bufferStartPosition || pos >= bufferStartPosition + buffer.limit()) {
+            // It will trigger re-fill on next read(), as remaining() will return 0.
+            buffer.limit(0)
+            // Refill will start form given position
```
dmitrii-artuhov:
```suggestion
            // Refill will start from the given position
```

lev-serebryakov-jetbrains: All reproducers I have is multi-gigabyte traces, unfortunately...

PR #891, opened 2025-11-18
Make integration tests generatable

zhelenskiy:
If someone forgets to regenerate the test, the build will fail automatically. Existing generators are found automatically using reflection.

With these changes:
* The user can run a single test.
* When a TeamCity bug is fixed, it would be possible to parallelize tests, as test classes are now different.
* It is possible to temporarily configure some tests regardless of the JSON.
* As tests are still generated from JSON, it is straightforward to add and bulk change the tests.

### DIFF
--- a/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinCompilerGeneratedTests.kt
+++ b/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinCompilerGeneratedTests.kt
@@ -2,7 +2,7 @@
 
 package org.jetbrains.trace.recorder.test.impl.generated
 
-import org.jetbrains.trace.recorder.test.impl.*
+import org.jetbrains.trace.recorder.test.runner.*
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.Timeout
--- a/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinImmutableCollectionsGeneratedTests.kt
+++ b/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinImmutableCollectionsGeneratedTests.kt
@@ -2,7 +2,7 @@
 
 package org.jetbrains.trace.recorder.test.impl.generated
 
-import org.jetbrains.trace.recorder.test.impl.*
+import org.jetbrains.trace.recorder.test.runner.*
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.Timeout
--- a/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KtorGeneratedTests.kt
+++ b/integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KtorGeneratedTests.kt
@@ -2,7 +2,7 @@
 
 package org.jetbrains.trace.recorder.test.impl.generated
 
-import org.jetbrains.trace.recorder.test.impl.*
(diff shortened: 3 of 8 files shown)


zhelenskiy:
Rebased on develop

integration-test/trace-recorder/build.gradle.kts
```diff
@@ -48,15 +49,16 @@ tasks {
...
+            excludeCategories("org.jetbrains.trace.recorder.test.runner.KtorTraceRecorderTest")
         }
 
         outputs.upToDateWhen { false } // Always run tests when called
         dependsOn(traceAgentIntegrationTestsPrerequisites)
         dependsOn(copyTraceRecorderFatJar)
     }
 
-    val traceRecorderIntegrationTestExtended = register<Test>("traceRecorderIntegrationTestExtended") {
+    val traceRecorderIntegrationTestKotlinCompiler = register<Test>("traceRecorderIntegrationTestKotlinCompiler") {
```
ivandev0: I hate this in Kotlin, I will hate this here. Why do we need separate categories for all tests?

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinCompilerTests.kt
```diff
@@ -0,0 +1,493 @@
...
+import org.junit.Test
+import org.junit.experimental.categories.Category
+
+/*
+  THIS FILE IS AUTOGENERATED!
+  DO NOT MODIFY!
+*/
+
+@Category(KotlinCompilerTraceRecorderTest::class)
+class KotlinCompilerTestsTraceRecorderIntegrationTest_FirLightTreeBlackBoxCodegenTestGenerated_testAllFilesPresentInBox : KotlinCompilerTraceRecorderJsonTests() {
```
ivandev0: Why do we have separation by classes instead of one class with several methods? Is this TC restriction?

ivandev0:
Please rebase. I pushed my refactoring changes to the develop.

zhelenskiy:
> Please rebase. I pushed my refactoring changes to the develop.

I rebased

integration-test/trace-recorder/build.gradle.kts
zhelenskiy: This is manual test parallelization because of TeamCity bugs. I just have multiple run configurations.

integration-test/trace-recorder/src/main/org/jetbrains/trace/recorder/test/impl/generated/KotlinCompilerTests.kt
zhelenskiy: Yes, it is necessary for TC to be able to parallelize the tests.

integration-test/trace-recorder/build.gradle.kts
ivandev0: Do you need both: multiple configurations and a new test class for each test?...

zhelenskiy: Yes, because multiple configurations are needed now for the current temporary solution with manual parallelization.

zhelenskiy:
Now there is a single test configuration, the tests are organized properly. All test suites and sub-suits are runnable separately.
image

A failing test is muted. There is a way to mute tests now.
Immutable collections JSON tests are also converted to generated from parametrized.

ivandev0:
Please move `AbstractJsonTraceRecorderIntegrationTest` and other non test files into `runner` package

ivandev0:
I see warnings when running tests `JUnit5 tests are parallel by default and its configured with `junit-platform.properties`, please remove `parallel=true` argument`. Is it possible to fix them?

zhelenskiy:
> I see warnings when running tests `JUnit5 tests are parallel by default and its configured with `junit-platform.properties`, please remove `parallel=true` argument`. Is it possible to fix them?

They existed before my PR and are unrelated.

PR #892, opened 2025-11-19
[JBRes-6527] Fix kover instrumentation postprocessing

dmitrii-artuhov:


dmitrii-artuhov:
> Can we add a test to show that was changed?

@ivandev0 I have added the representation

dmitrii-artuhov:
> I don't see any changes in the test data after your fix. Did you apply commits in the correct order?

See the new file `integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx-datetime/kotlinx.datetime.test.ConvertersTest_instant.txt` it has the trace for datetime test that we had without trace previously. Without my fix there would be `__$hits$__` field assignment and call to `CoverageRuntime.getHitsMask(...)`

Of what do you mean here?

ivandev0:
> > I don't see any changes in the test data after your fix. Did you apply commits in the correct order?
> 
> See the new file `integration-test/trace-recorder/src/main/resources/integrationTestData/kotlinx-datetime/kotlinx.datetime.test.ConvertersTest_instant.txt` it has the trace for datetime test that we had without trace previously. Without my fix there would be `__$hits$__` field assignment and call to `CoverageRuntime.getHitsMask(...)`
> 
> Of what do you mean here?

Ah, I see. You make a change and after that add a test. In the future please add tests before making changes. This way I can see that actually changed in the test.

PR #894, opened 2025-11-20
JBRes-6687 Trace recorder calls project code

lev-serebryakov-jetbrains:
CharSequence interface can be implemented in user code, and then Trace Recorder will call it to get string representation.

3rd party implementations can be incomplete or contain side effects.

Don't try to call implementations not from standard library, filter by packages.

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
@@ -730,6 +730,23 @@ private fun CharSequence.trimToString(): String {
     return if (string.length > MAX_TROBJECT_STRING_LENGTH) "${string.take(MAX_TROBJECT_STRING_LENGTH)}..." else string
 }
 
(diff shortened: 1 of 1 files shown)


trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -730,6 +730,23 @@ private fun CharSequence.trimToString(): String {
...
+    // CharBuffer
+    "java.nio.",
+    // Segment
+    "javax.swing.",
+    // Kotlin "wrappers"
+    "kotlin.text.",
+    "kotlin."
+)
+
+private fun classNameWhitelisted(obj: Any, whiteList: List<String>): Boolean {
```
dmitrii-artuhov: I would use `WHITELIST_PACKAGES_FOR_TO_STRING` directly in the body of this function, since there is a single place where it is needed

lev-serebryakov-jetbrains: Maybe we will need it in the future... Function is more generic for future-proof.

PR #896, opened 2025-11-20
JBRes-6983 io.ktor.client.engine.okhttp.OkHttpEngineTests::testRequestAfterRecreate fails

lev-serebryakov-jetbrains:
## Analysis:

The test failed because the Trace Recorder complained about too many calls to `TraceRecorder.finishTraceAndDumpResults()`.

The Second call to `TraceRecorder.finishTraceAndDumpResults()` was originated in the `finally` block in test method due to exception thrown on first call to `TraceRecorder.finishTraceAndDumpResults()`.

Exception was caused by conflict between `afterThreadRunReturn()` of one of the threads and `completeRunningThread(),` because `afterThreadRunReturn()` closes all method call trace points and `completeRunningThread()` try to close last trace point without any checks.

## Solution:

1. Add `try {} catch(...) {}` to start/stop injections to report all exceptions in these methods.

2. Rework thread finishing & stack processing:
    1. Store the first call on the thread (root of call tree) into a separate field in `ThreadData`. 
    2. Always pop the whole stack, don't treat top-level frame specifically (except for the main thread where it is set to `VOID` result).
    3. Skip `completeRunningThread()` for threads with empty stacks in `finishAndDumpTrace()`: it means «natural» thread finish wins, it is Ok.

### DIFF
--- a/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
+++ b/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
@@ -328,7 +328,7 @@ class TraceCollectingEventTracker(
 
     override fun afterThreadRunReturn(threadDescriptor: ThreadDescriptor) = threadDescriptor.runInsideInjectedCode {
         val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: return
-        completeInvokedMethodCalls(Thread.currentThread(), threadData, reportLoops = false) { _, tp -> tp.result = TR_OBJECT_UNTRACKED_METHOD_RESULT }
+        completeInvokedMethodCalls(Thread.currentThread(), threadData) { _, tp -> tp.result = TR_OBJECT_UNTRACKED_METHOD_RESULT }
         threadDescriptor.disableAnalysis()
     }
 
@@ -337,7 +337,7 @@ class TraceCollectingEventTracker(
         exception: Throwable
     ) = threadDescriptor.runInsideInjectedCode {
         val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: throw exception
-        completeInvokedMethodCalls(Thread.currentThread(), threadData, reportLoops = false) { _, tp -> tp.setExceptionResult(exception) }
+        completeInvokedMethodCalls(Thread.currentThread(), threadData) { _, tp -> tp.setExceptionResult(exception) }
         threadDescriptor.disableAnalysis()
         throw exception
     }
@@ -907,10 +907,17 @@ class TraceCollectingEventTracker(
     }
 
     /**
-     * Completes all method calls of the current thread via [tracePointFinisher] and empties the stack.
+     * Completes all method calls of the current thread via [onMethodCallCompletion] and empties the stack.
+     *
+     * @param loopsCompletionExpected when set to `false` will report an error if there were any loop tracepoints completed along the process.
      */
-    private fun completeInvokedMethodCalls(thread: Thread, threadData: ThreadData, reportLoops: Boolean, tracePointFinisher: (stackLevel: Int, tracePoint: TRMethodCallTracePoint) -> Unit) {
-        // End all method calls, for which we did not track the method return,
+    private fun completeInvokedMethodCalls(
+        thread: Thread,
+        threadData: ThreadData,
+        loopsCompletionExpected: Boolean = true,
+        onMethodCallCompletion: (stackLevel: Int, tracePoint: TRMethodCallTracePoint) -> Unit
+    ) {
+        // End all method calls, for which we did not track the method return
         while (threadData.getStack().isNotEmpty()) {
             val hasLoops = threadData.currentLoopTracePoint() != null
             while (threadData.currentLoopTracePoint() != null) {
@@ -918,11 +925,11 @@ class TraceCollectingEventTracker(
             }
 
             val tracePoint = threadData.popStackFrame()
-            if (hasLoops && reportLoops) {
+            if (hasLoops && !loopsCompletionExpected) {
                 Logger.error { "Forced exit from method ${tracePoint.className}.${tracePoint.methodName} breaks loops." }
             }
 
-            tracePointFinisher(threadData.getStack().size, tracePoint)
+            onMethodCallCompletion(threadData.getStack().size, tracePoint)
             strategy.completeContainerTracePoint(thread, tracePoint)
         }
         strategy.completeThread(thread)
@@ -940,7 +947,7 @@ class TraceCollectingEventTracker(
         // Early exit because we must skip `strategy.completeThread(thread)` for this thread too.
         if (threadData.getStack().isEmpty()) return
 
-        completeInvokedMethodCalls(thread, threadData, reportLoops = false) { _, tp -> tp.result = TR_OBJECT_UNFINISHED_METHOD_RESULT }
+        completeInvokedMethodCalls(thread, threadData) { _, tp -> tp.result = TR_OBJECT_UNFINISHED_METHOD_RESULT }
     }
 
     private fun completeMainThread(thread: Thread, threadDescriptor: ThreadDescriptor) {
@@ -957,7 +964,7 @@ class TraceCollectingEventTracker(
         }
 
         val overflowStack = mutableListOf<String>()
-        completeInvokedMethodCalls(thread, threadData, reportLoops = true) { stackLevel, tp ->
+        completeInvokedMethodCalls(thread, threadData, loopsCompletionExpected = false) { stackLevel, tp ->
             if (stackLevel != 0) {
                 overflowStack.add("${tp.className}.${tp.methodName}")
             }


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +925,73 @@ class TraceCollectingEventTracker(
...
-                threadData.popStackFrame() // remove the current frame from the stack
-            } else {
-                // basically, the last `Thread.run()` call is preserved as a root node
-                // for 'text' tracing mode to work properly
-                break
+        }
+        strategy.completeThread(thread)
+    }
+
+    private fun completeMainThread(thread: Thread, threadData: ThreadData) {
```
dmitrii-artuhov: Please unify the argument type of `completeRunningThread`, `completeMainThread`: either both accept `ThreadData` of `ThreadDewcriptor`

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +925,73 @@ class TraceCollectingEventTracker(
...
             strategy.completeContainerTracePoint(thread, tracePoint)
-            // do not pop the last method call in order for 'text' tracing mode to work
-            // it is used for debugging in the `TRPlayground.kt` file
-            if (stack.size != 1) {
-                threadData.popStackFrame() // remove the current frame from the stack
-            } else {
-                // basically, the last `Thread.run()` call is preserved as a root node
-                // for 'text' tracing mode to work properly
-                break
+        }
```
dmitrii-artuhov: The logic inside this while loop could be united with the `forceThreadEnd` method. 
They only differentiate in assigning method call result/exception and loops handling

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -313,26 +321,39 @@ class TraceCollectingEventTracker(
...
     }
 
-    override fun afterThreadRunReturn(threadDescriptor: ThreadDescriptor) = threadDescriptor.runInsideInjectedCode {
-        val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: return
+    /**
+     * Method ends all methods of current thread nd empties stack.
+     *
+     * How trace points are finished is determined by second parameter.
+     */
+    private fun forceThreadEnd(threadData: ThreadData, tracePointFinisher: (tracePoint: TRMethodCallTracePoint) -> Unit) {
```
dmitrii-artuhov: should probably be smth like: `completeInvokedMethodCalls`

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +925,73 @@ class TraceCollectingEventTracker(
...
+            while (threadData.currentLoopTracePoint() != null) {
+                exitCurrentLoop(thread, threadData)
             }
+            // Pop stack
+            val tracePoint = threadData.popStackFrame()
+            errorStack.add("${tracePoint.className}.${tracePoint.methodName}")
+            // complete the method call
+            tracePoint.result = TR_OBJECT_UNFINISHED_METHOD_RESULT
+            strategy.completeContainerTracePoint(thread, tracePoint)
+        }
```
dmitrii-artuhov: Code duplication with `forceThreadEnd` and the same loop in the `completeRunningThread`

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +925,73 @@ class TraceCollectingEventTracker(
...
-            } else {
-                // basically, the last `Thread.run()` call is preserved as a root node
-                // for 'text' tracing mode to work properly
-                break
+        }
+        strategy.completeThread(thread)
+    }
+
+    private fun completeMainThread(thread: Thread, threadData: ThreadData) {
+        val errorStack = mutableListOf<String>()
```
dmitrii-artuhov: `errorStack` misleads the purpose of the variable, `stackLeftover`/`overflowStack` suits better (I would use the second variant)

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -962,11 +1009,15 @@ class TraceCollectingEventTracker(
...
-        tracePoint.result = TR_OBJECT_VOID
-        strategy.completeContainerTracePoint(Thread.currentThread(), tracePoint)
+
+        // Close this thread call stack (it must be 1 element, complain about problems)
+        val threadData = ThreadDescriptor.getCurrentThreadDescriptor()?.eventTrackerData as? ThreadData?
+        if (threadData == null) {
+            Logger.error { "Main tracing thread ${mainThread.name} doesn't have event tracker data" }
+        } else {
+            completeMainThread(mainThread, threadData)
+        }
```
dmitrii-artuhov: I would put the `if` in the body of the `completeMainThread` and leave here only the methd call

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -313,26 +321,39 @@ class TraceCollectingEventTracker(
...
             threadData.pushStackFrame(tracePoint, thread, isInline = false)
         }
     }
 
-    override fun afterThreadRunReturn(threadDescriptor: ThreadDescriptor) = threadDescriptor.runInsideInjectedCode {
-        val threadData = threadDescriptor.eventTrackerData as? ThreadData? ?: return
+    /**
+     * Method ends all methods of current thread nd empties stack.
+     *
+     * How trace points are finished is determined by second parameter.
```
dmitrii-artuhov:
```suggestion
     * Completes all method calls of the current thread via [tracePointFinisher] and empties the stack.
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -313,26 +321,39 @@ class TraceCollectingEventTracker(
...
         // End all method calls, for which we did not track the method return,
-        // except for the very first one (`java.lang.Thread::run` method)
-        while (threadData.getStack().size > 1) {
+        while (threadData.getStack().isNotEmpty()) {
+            val loop = threadData.currentLoopTracePoint()
             val tracePoint = threadData.popStackFrame()
-            tracePoint.result = TR_OBJECT_UNTRACKED_METHOD_RESULT
+
+            // There should be no loops in these methods, they should not be here if they were instrumented,
+            // amd loops should be closed naturally for instrumented methods
```
dmitrii-artuhov:
```suggestion
            // and loops should be closed naturally for instrumented methods
```

dmitrii-artuhov (changes requested):

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
lev-serebryakov-jetbrains: I've thought about it, but loop handling is what stopped me.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderInjections.kt
```diff
@@ -38,19 +31,27 @@ internal object TraceRecorderInjections {
...
+        } catch (t: Throwable) {
+            Logger.error { "Cannot start Trace Recorder: $t"}
+        }
     }
 
     @JvmStatic
     fun stopTraceRecorderAndDumpTrace() {
-        if (alreadyStopped) return
-        TraceRecorder.finishTraceAndDumpResults()
+        // This method never should throw execution, or tracer state is undetermined
```
dmitrii-artuhov:
```suggestion
        // This method should never throw an exception, or tracer state is undetermined
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +900,81 @@ class TraceCollectingEventTracker(
...
-            if (stack.size != 1) {
-                threadData.popStackFrame() // remove the current frame from the stack
-            } else {
-                // basically, the last `Thread.run()` call is preserved as a root node
-                // for 'text' tracing mode to work properly
-                break
+
+        // Skip thread completely if it is finished "naturally"
+        // Early exit because we must skip `strategy.completeThread(thread)`
+        // for this thread too.
```
dmitrii-artuhov:
```suggestion
        // Skip this thread completely if it is finished "naturally" (i.e. by concurrent afterThreadRun[Return|Exception]).
        // Early exit because we must skip `strategy.completeThread(thread)` for this thread too
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -962,11 +992,10 @@ class TraceCollectingEventTracker(
                 completeRunningThread(thread, threadDescriptor)
             }
 
-        // Close this thread callstack
-        val threadData = ThreadDescriptor.getCurrentThreadDescriptor()?.eventTrackerData as? ThreadData? ?: return
-        val tracePoint = threadData.currentMethodCallTracePoint()!!
-        tracePoint.result = TR_OBJECT_VOID
-        strategy.completeContainerTracePoint(Thread.currentThread(), tracePoint)
+
+        // Close this thread call stack (it must be 1 element, complain about problems)
```
dmitrii-artuhov:
```suggestion
        // Close this thread call stack (it must be 1 element, complain about problems otherwise)
```

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -906,45 +900,81 @@ class TraceCollectingEventTracker(
...
         threadData.pushStackFrame(tracePoint, null, isInline = false)
 
         startTime = System.currentTimeMillis()
     }
 
-    fun completeRunningThread(thread: Thread, threadDescriptor: ThreadDescriptor) {
+    /**
+     * Completes all method calls of the current thread via [tracePointFinisher] and empties the stack.
+     */
+    private fun completeInvokedMethodCalls(thread: Thread, threadData: ThreadData, reportLoops: Boolean, tracePointFinisher: (stackLevel: Int, tracePoint: TRMethodCallTracePoint) -> Unit) {
```
dmitrii-artuhov: Please make it:
```
/**
 * Completes all method calls of the current thread via [onMethodCallCompletion] and empties the stack.
 * 
 * @param loopsCompletionExpected when set to `false` will report an error if there were any loop tracepoints completed along the process.
 */
completeInvokedMethodCalls(
  thread: Thread,
  threadData: ThreadData,
  loopsCompletionExpected: Boolean = true,
  onMethodCallCompletion: (stackLevel: Int, tracePoint: TRMethodCallTracePoint) -> Unit
) { ... }
```

Pass `loopsCompletionExpected = false`, in a single place where it is required (when Main thread finishes)

PR #897, opened 2025-11-21
JBRes-6650 Invalid code location of the method which is run via button click

lev-serebryakov-jetbrains:
Pass code location for first synthetic tracepoint.

There are some problems with resolving (still) in plugin, but code location is correct now.

### DIFF
--- a/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderMethodTransformer.kt
+++ b/trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderMethodTransformer.kt
@@ -54,14 +54,14 @@ internal class TraceRecorderMethodTransformer(
         super.onMethodEnter()
         // Make code location
         // Line number may be unknown yet, update it later if needed
-        val ste = StackTraceElement(
+        val stackTraceElement = StackTraceElement(
             /* declaringClass = */ className.toInternalClassName(),
             /* methodName = */ name,
             /* fileName = */ fileName,
             /* lineNumber = */ firstLine
         )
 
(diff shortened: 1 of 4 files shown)


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderMethodTransformer.kt
```diff
@@ -38,15 +40,41 @@ import org.objectweb.asm.commons.GeneratorAdapter
...
+    private var codeLocationId: Int = -1
+
+    override fun visitLineNumber(line: Int, start: Label?) {
+        super.visitLineNumber(line, start)
+        // Code location has been created without known line number, update line number
+        if (lineNumber == 0 && codeLocationId >= 0) {
+            CodeLocations.updateCodeLocationLineNumber(codeLocationId, line)
+        }
+        lineNumber = line
+    }
```
eupp: Is it possible to somehow avoid mutable update of code location mappings, and prepare line number data in advance? Like, for instance, using the same idea with first reading a `ClassNode`, like we do in `LincheckClassFileTransformer`?

I am afraid exposing mutable API is dangerous. Even if we use it in a single place currently, people may (unknowingly) abuse it in the future.

lev-serebryakov-jetbrains: I don't like it either, but it looks like double-processing is much more effort... I'll try.

lev-serebryakov-jetbrains: Done.

PR #898, opened 2025-11-26
Enable eager instrumentation for Trace Recorder 

eupp:


trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceCollectingEventTracker.kt
```diff
@@ -378,14 +378,8 @@ class TraceCollectingEventTracker(
...
-    override fun beforeNewObjectCreation(
-        threadDescriptor: ThreadDescriptor,
-        className: String
-    ) = runInsideIgnoredSection {
-        LincheckJavaAgent.ensureClassHierarchyIsTransformed(className)
-    }
-
-    override fun afterNewObjectCreation(threadDescriptor: ThreadDescriptor, obj: Any) = Unit
+    override fun beforeNewObjectCreation(threadDescriptor: ThreadDescriptor, className: String) {}
+    override fun afterNewObjectCreation(threadDescriptor: ThreadDescriptor, obj: Any) {}
```
dmitrii-artuhov, line 382: There are 2 more places in this file where no-op hooks are written via: `hookName(...) = Unit`, could you rewrite them to `hookName(...) {}` as well

PR #899, opened 2025-12-01
Collect packages from source files of the project to which the agent is attached

dmitrii-artuhov:
This PR adds 2 new agent parameters:
- `projectPath`: path to the project from which packages should be collected
- `excludeDirPaths`: `;`-separated paths to folders, which should be excluded from the sources search 

They should be provided when the agent is run with Junie for scoping the include patterns only to those that are part of the project. If `projectPath` is not provided, then this computation will not be run.

--------

Not supported:
- does not add class names to include patterns from source files which package name is empty


bbrockbernd:
@dmitrii-artuhov Should we still merge this after @eupp told us today that the improvements are not significant?

dmitrii-artuhov:
We can decline it for now, I will close the PR

eupp:
Why declining though? Lets get back to it next week.

PR #901, opened 2025-12-02
IJ integration tests

bbrockbernd:
- Fix: propagate trace format settings
- Fix: increase memory for IJ tests
- Added 15 tests (more will come, and not all do currenlty pass)
- Checks wether some keyword is part of the trace, to verify that "important" part has actually been traced. This is very helpfull when using cases from benchmarks as there we know what files were changed in the fix. Now we can verify that those files are actually traced.

### DIFF
--- a/integration-test/common/src/main/AbstractIntellijTraceIntegrationTest.kt
+++ b/integration-test/common/src/main/AbstractIntellijTraceIntegrationTest.kt
@@ -30,18 +30,19 @@ abstract class AbstractIntellijTraceIntegrationTest: AbstractTraceIntegrationTes
         // We need to escape it twice, as our argument parser will de-escape it when split into array
         val pathToOutput = outputFile.absolutePath.escape().escape()
         
-        val cmd = mutableListOf(
-            "./tests.cmd",
-            "-Dintellij.build.test.patterns=${testClassName.escapeDollar()}",
-            "-Dintellij.build.test.trace.recorder.enabled=true",
-            "-Dintellij.build.test.trace.recorder.className=${testClassName.escapeDollar()}",
-            "-Dintellij.build.test.trace.recorder.methodName=${testMethodName.escapeDollar()}",
-            "-Dintellij.build.test.trace.recorder.traceDump=$pathToOutput",
-            "-Dintellij.build.test.trace.recorder.agentJar=$pathToFatJar",
-            "-Dintellij.build.test.jvm.memory.options=-Xmx4g"
-        )
-        extraAgentArgs["format"]?.let { cmd.add("-Dintellij.build.test.trace.recorder.format=$it") }
-        extraAgentArgs["formatOption"]?.let { cmd.add("-Dintellij.build.test.trace.recorder.formatOption=$it") }
+        val cmd = buildList {
+            add("./tests.cmd")
+            add("-Dintellij.build.test.patterns=${testClassName.escapeDollar()}")
+            add("-Dintellij.build.test.trace.recorder.enabled=true")
+            add("-Dintellij.build.test.trace.recorder.className=${testClassName.escapeDollar()}")
+            add("-Dintellij.build.test.trace.recorder.methodName=${testMethodName.escapeDollar()}")
+            add("-Dintellij.build.test.trace.recorder.traceDump=$pathToOutput")
+            add("-Dintellij.build.test.trace.recorder.agentJar=$pathToFatJar")
+            add("-Dintellij.build.test.jvm.memory.options=-Xmx4g")
+
+            extraAgentArgs["format"]?.let { add("-Dintellij.build.test.trace.recorder.format=$it") }
+            extraAgentArgs["formatOption"]?.let { add("-Dintellij.build.test.trace.recorder.formatOption=$it") }
+        }
 
         ProcessBuilder(cmd).directory(File(projectPath))
             .inheritIO().start().waitFor()


integration-test/common/src/main/AbstractIntellijTraceIntegrationTest.kt
```diff
@@ -29,15 +29,21 @@ abstract class AbstractIntellijTraceIntegrationTest: AbstractTraceIntegrationTes
 
         // We need to escape it twice, as our argument parser will de-escape it when split into array
         val pathToOutput = outputFile.absolutePath.escape().escape()
-        ProcessBuilder(
+        
+        val cmd = mutableListOf(
```
ivandev0: You can use buildList here.
```
val cmd = buildList {
    add("./tests.cmd")
    // ...
    extraAgentArgs["format"]?.let { add("-Dintellij.build.test.trace.recorder.format=$it") }
    // ...
```

Not sure it is much better though. Your call.

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
```diff
@@ -131,34 +139,57 @@ abstract class AbstractTraceIntegrationTest {
...
+                            "Please rerun the test with \"overwriteRepresentationTestsOutput\" option enabled."
+                )
             }
         } else {
-            fun checkNonEmptyNess(file: File, filePurpose: String = "output") {
-                val fileContainsContent = file.bufferedReader().lineSequence().any { it.isNotEmpty() }
-                if (!fileContainsContent) {
-                    Assertions.fail<Unit>("Empty $filePurpose file was produced by the test: $file.")
-                }
+            val file = if (outputFile.exists()) outputFile
```
ivandev0, line 153: I think it is better to move under `checkRepresentation`. If we have `traceShouldContain`, then check only these lines. It is is empty, then check the whole file with golden data.

bbrockbernd:
All added tests should pass since latest fix of dima. Only the test that was already there fails.

It fails because hashCodes are not constant and relies on comparing output with file.
The reason hashCodes are not constant is that the `extraJvmArgs` are not propagated to the IJ `test.cmd` script. 
Technically it should be possible with `-Dintellij.build.test.jvm.memory.options`. But it somehow fails once the `-XX:hashCode=2` arg is added. (yes i have enabled experimental vm options)

In any case I don't think it is worth it to investigate this further as IJ test outputs are generally very large and searching for some `shouldContain` keywords is more effective.

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
bbrockbernd: When recording for checkRepresentation we use `binary stream` fromat options, when we record for file comparison we use `text`. So I don't think it makes sense to combine them.

Also, for how I see it, the `shouldContain` functionality is to verify that trace recorder has reached and recorded the "important" part of the execution. While text file comparison is to flag regressions in trace representation when we make changes to lincheck.

But, if you would still like to see it combined, it can still be done ofcourse.

integration-test/common/build.gradle.kts
```diff
@@ -2,6 +2,10 @@ repositories {
     mavenCentral()
     maven { url = uri("https://repo.gradle.org/gradle/libs-releases/") }
 }
+dependencies {
+    implementation(project(":common"))
+    implementation(project(":trace"))
+}
```
eupp, line 8: Why do you need to add these libraries as dependencies for integration tests?

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
eupp: Am I right that:
- when `checkRepresentation` is true, the `text` format option is used, and we compare the recorded trace with golden data as plain text?
- when `shouldContain` is true, you use binary stream recording, and then iterate through the trace using `LazyTraceReader` searching for given strings?

I have a bunch of questions then:
- Why not to use `text` format when `shouldContain` is passed, and use simple text search? 
- Is it because for IJ tests the text output is too large? 
- Should we then check for `shouldContain` on IJ tests at all?
- Maybe we can still use `text` output, but reduce the trace size by using include/exclude filters?
- Can we somehow check `shouldContain`, or perform similar sanity check, but avoid reading the binary trace?

integration-test/common/build.gradle.kts
eupp: Oh, I see, is it to use `LazyTraceReader` for decoding the binary trace?

integration-test/common/src/main/AbstractTraceIntegrationTest.kt
bbrockbernd: Hi! 

> when checkRepresentation is true, the text format option is used, and we compare the recorded trace with golden data as plain text?
when shouldContain is true, you use binary stream recording, and then iterate through the trace using LazyTraceReader searching for given strings?

Yes and yes. However, as discussed yesterday we can probably remove the text variant all together for IJ tests since setting the hashCodes to a constant doesn't work.

> Why not to use text format when shouldContain is passed, and use simple text search?
Is it because for IJ tests the text output is too large?

The reason why we don't use text here is because when we use text format, the tests time out (longer than 20 min before trying to compare any file). While, when we use a binary it is just a few minutes in total. And probably the text files are very very big indeed.

> Should we then check for shouldContain on IJ tests at all?

Yes, I think this is important. We have been recording large traces for a while without verifying that they actually reach important parts of the execution. Which is in the end what we want. The benchmark tests are an easy source for this. They have tests with large complicated traces. But also the classes where a fix should come is known. We can use this info to verify that Trace Recorder actually records those classes when it is recording very large and complex traces.

> Maybe we can still use text output, but reduce the trace size by using include/exclude filters?

The plugin does not allow users to include/exclude. Therefore, I think it is good to have integration tests that replicate real world plugin scenarios. Scenarios, that we promise to support.

> Can we somehow check shouldContain, or perform similar sanity check, but avoid reading the binary trace?

I don't really see what the problem is with using the binary format. Other integration tests (kotlin for isntance) use binaries as well. We just never check the output of those. Moreover, I think it is good to have some integration tests that record into and read from binaries. This is also what the plugin does and so it is probably good to have this as tests.

eupp: > I don't really see what the problem is with using the binary format.

I am a bit concerned about adding the `common` and `trace` modules as dependencies for integration tests. But if we don't see any problems in practice with this (at least currently), let it be. 

I would also say that `shouldContain` option should not depend on which output format is used. Perhaps, output option should be a separate test option altogether. At least, let's create a YT issue about supporting both binary/text options in our testing infra uniformly.

bbrockbernd:
> Looks fine. I am only scared of `compareOutput` method. I would recommend refactoring and simplifying it as much as possible (leave only one file instead of two, try to minimize nesting and so on). But this can be done separately.

We created an issue to redo the output validation, such that we only use binary or text. `compareOutput` should be simplified here as well. [JBRes-7291](https://youtrack.jetbrains.com/issue/JBRes-7291) Refactor integration test output validation

PR #903, opened 2025-12-12
Shadow asm and bytebuddy

bbrockbernd:
There was a [bug](https://github.com/GradleUp/shadow/issues/276) with older versions of the shadow plugin where bootstrap.jar was expanded and not availbale anymore at `/bootstrap.jar`

This has been resolved in newer versions. To support this I had to upgrade to gradle 8.11.

Two questions:
- Should we just upgrade to latest gradle?
- I now shade asm and bytebuddy, are there other packages that we would like to hide?

### DIFF
--- a/buildSrc/src/main/kotlin/TraceAgentTasks.kt
+++ b/buildSrc/src/main/kotlin/TraceAgentTasks.kt
@@ -38,15 +38,15 @@ fun Project.registerTraceAgentTasks(fatJarName: String, fatJarTaskName: String, 
 
     
     val copyBootstrapJar = tasks.register<Copy>("copyBootstrapJar") {
-        dependsOn(":bootstrap:jar")
+        dependsOn(":bootstrapJar")
         from(file(
             listOf(boostrapBuildDir, "libs", "bootstrap.jar").joinToString(separator = File.separator)
         ))
-        println(processedBootstrapJarPath)
         into(file(processedBootstrapJarPath))
     }
     
-    // Hack to not unpack bootstrap.jar during shadow task.
+    // Hack to prevent unpacking bootstrap.jar during shadowing task.
+    // When relocation starts, it will unwrap the outer archive and extract the inner one without change
     val jarWrapper = tasks.register<Jar>("jarWrapper") {
         destinationDirectory.set(file(
             listOf(processedBootstrapJarPath).joinToString(separator = File.separator)


bbrockbernd:
Upgrading gradle seems to be problematic. Will check if we can find a workaround for the older shadow plugin

dmitrii-artuhov:
You can wrap the `bootstrap.jar` in a jar archive and the old version should not extract its classes
https://jetbrains.team/p/data-race-detection/repositories/data-race-detection/files/5ee1e9622236971bf18f94cbc25cd0fe0358a2ff/build.gradle.kts?tab=source&line=59&lines-count=12

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -10,6 +10,6 @@
 
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-all.zip
```
dmitrii-artuhov, line 13: Why is this required?

bbrockbernd: Please read the PR description ;)

dmitrii-artuhov: Could you please try to implement it without bumping the gradle version (via the hack I've sent). If that suits there is no need for gradle version increase

buildSrc/src/main/kotlin/TraceAgentTasks.kt
```diff
@@ -10,30 +10,60 @@
...
+    val mainBuildDir: String = layout.buildDirectory.get().asFile.path
+
+
+    val processedBootstrapJarPath = listOf(mainBuildDir, "bootstrap-tmp").joinToString(separator = File.separator)
+    val boostrapBuildDir: String = project(":bootstrap").layout.buildDirectory.get().asFile.path
 
-    val traceAgentFatJar = tasks.register<Jar>(fatJarTaskName) {
+    
+    val copyBootstrapJar = tasks.register<Copy>("copyBootstrapJar") {
+        dependsOn(":bootstrap:jar")
```
dmitrii-artuhov: 'dependsOn(":bootstrapJar")'

I think

buildSrc/src/main/kotlin/TraceAgentTasks.kt
```diff
@@ -10,30 +10,60 @@
...
+        from(file(
+            listOf(boostrapBuildDir, "libs", "bootstrap.jar").joinToString(separator = File.separator)
+        ))
+        println(processedBootstrapJarPath)
```
dmitrii-artuhov: This println is probably not needed

buildSrc/src/main/kotlin/TraceAgentTasks.kt
```diff
@@ -10,30 +10,60 @@
...
+        into(file(processedBootstrapJarPath))
+    }
+    
+    // Hack to not unpack bootstrap.jar during shadow task.
```
dmitrii-artuhov:
```suggestion
    // Hack to pack bootstrap.jar inside another jar archive during shadowing. When relocation starts, it will unwrap the outer archive and extract the inner one without change
```

more like this?

buildSrc/src/main/kotlin/TraceAgentTasks.kt
bbrockbernd: oops

buildSrc/src/main/kotlin/TraceAgentTasks.kt
bbrockbernd: Both options seem to work

PR #904, opened 2025-12-17
Fix problem with targeting methods where source files are not available

bbrockbernd:
We would like to target a method from `com.intellij.rt.junit.JUnitStarter` to allow tested logic in spring project to appear in trace. This class however, does not have source files available. This fixes that problem.

### DIFF
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentTransformer.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentTransformer.kt
@@ -115,7 +115,7 @@ private class TraceAgentClassVisitor(
 
     override fun visitSource(source: String?, debug: String?) {
         super.visitSource(source, debug)
-        fileName = source ?: ""
+        fileName = source ?: "<unknown>"
     }
 
     override fun visitMethod(


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TraceAgentTransformer.kt
```diff
@@ -115,7 +115,7 @@ private class TraceAgentClassVisitor(
 
     override fun visitSource(source: String?, debug: String?) {
         super.visitSource(source, debug)
-        fileName = source!!
+        fileName = source ?: ""
```
dmitrii-artuhov: Could you please check that in plugin everything works as expected: the trace is shown, navigation does not break (it would not open the file, but at least nothing bad should happen)

Also `<unknown>` is better, I think. Or is it important for plugin to be an empty string?

bbrockbernd: Plugin seems to be okay. Will do some more checks to be sure.

Please note that tracepoints from these source less methods could already appear when called. 

I chose for an empty string because that is how it is initialized at line 102.

dmitrii-artuhov (approved):

PR #907, opened 2025-12-19
[JBRes-7391] Enhance the Trace API to support diffs

lev-serebryakov-jetbrains:
Add new field to all tracepoints: `DiffStatus`. If trace is not a diff, it is always `COMMON`.

Also, extent TRMethodCallTracePoint with flags to mark if it has added or removed tracepoints.

Add stub for diff-creating API.

Add API for diff metainfo (always false for now).

Bump serialization version.

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
@@ -170,7 +170,7 @@ sealed class TRContainerTracePoint(
         }
         // Should we check for override? Lets skip for now
         children[index] = child
-        addChild(child)
+        addChildStatus(child)
         child.setParentIfContainer(this)
     }
 


trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -337,6 +337,20 @@ class LazyTraceReader private constructor(
 
     val metaInfo: TraceMetaInfo? get() = input.metaInfo
 
+    val isDiff: Boolean get() = input.metaInfo?.isDiff ?: false
```
eupp, line 334: Minor cosmetics request: there is `readTracePointWithPostprocessor()` method above, could you please move it past initialization code.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -337,6 +337,20 @@ class LazyTraceReader private constructor(
...
+
+    val leftMetaInfo: TraceMetaInfo?
```
```suggestion
    val leftTraceMetaInfo: TraceMetaInfo?
```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -337,6 +337,20 @@ class LazyTraceReader private constructor(
...
+        get() {
+            check(isDiff) { "Cannot provide left trace meta info if trace is not a diff" }
+            return input.metaInfo?.leftMetaInfo
+        }
+
+    val rightMetaInfo: TraceMetaInfo?
```
```suggestion
    val rightTraceMetaInfo: TraceMetaInfo?
```

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -27,18 +27,54 @@ import java.io.DataInput
...
 var INJECTIONS_VOID_OBJECT: Any? = null
 
+/**
+ * Describes status of tracepoint in trace diff
+ */
+enum class DiffStatus {
+    /**
+     * Trace is not diff at all, or this tracepoint is identical in two traces.
+     */
+    COMMON,
```
eupp:
```suggestion
    NONE,
```
or could as well just use `DiffStatus?` with `null` denoting non-diff data.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -27,18 +27,54 @@ import java.io.DataInput
...
+
+    /**
+     * This tracepoint was not found in the first, but found in the second trace, when diff was created.
+     */
+    ADDED,
+
+    /**
+     * This is call tracepoint, and it has the same method in both traces but differs in arguments values.
+     */
+    ARGS_DIFFERENCE,
```
eupp: Let's make it more future proof. We can name it `EDITED` where edited has open meaning: it could be arguments change, or could be some other change of trace point payload (without structural change).

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -237,8 +290,12 @@ class TRMethodCallTracePoint(
     }
 
     companion object {
-        // Flag which tells that the method was not tracked from its start and has some missing tracepoints
+        // Flag that tells that the method was not tracked from its start and has some missing tracepoints
         const val INCOMPLETE_METHOD_FLAG: Int = 1
+        // Flag that tells that method call has child(ren) with [DiffStatus.ADDED] diff status
+        const val HAS_ADDED_CHILDREN_FLAG: Int = 2
+        // Flag that tells that method call has child(ren) with [DiffStatus.REMOVED] diff status
+        const val HAS_REMOVED_CHILDREN_FLAG: Int = 4
```
eupp: Why do you think we need these two flags? Can't reader just read children of the diff tree node and see if there were some modifications?

trace/src/main/org/jetbrains/lincheck/trace/SerializationDiff.kt
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * This Source Code Form is subject to the terms of the
+ * Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.lincheck.trace
+
+fun diffTwoTraces(left: LazyTraceReader, right: LazyTraceReader, outputBaseName: String) {
+    TODO()
```
eupp:
```suggestion
    // TODO
```

otherwise any attempt to call this stub from plugin code will result in exception.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: And what about really COMMON case - it is diff but this lime is common between two?

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: Lazy loading. We could want to color node in the tree (and Nikita specifically mentioned this status) before reading all children.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
lev-serebryakov-jetbrains: I'm leaving `COMMON` for diffs and enable `null` for simple traces.

eupp: Oh, I see. I thought COMMON stands for "common case of non-diff-trace". Let's use `UNCHANGED` then for clarity.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
eupp: > We could want to color node in the tree ... before reading all children

Hmm, maybe then we should cover all `DiffStatus` cases? That is, also include `HAS_EDITED_CHILDREN_FLAG` ?

Also, we now have other types of `TRContainerTracePoint` (e.g., for loops) --- shouldn't these flags be applicable to all container trace points as well?

lev-serebryakov-jetbrains: I've thought about Enum of statuses. Maybe it is better to add one to all Containers, yes.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -27,18 +27,58 @@ import java.io.DataInput
...
+
+    /**
+     * This tracepoint was not found in the first, but found in the second trace, when diff was created.
+     */
+    ADDED,
+
+    /**
+     * Tracepoint was edited.
+     *
+     * For example, fi it is a method called tracepoint, it has the same method in both traces but differs in arguments values.
```
dmitrii-artuhov:
```suggestion
     * For example, if it is a method called tracepoint, it has the same method in both traces but differs in arguments values.
```

PR #908, opened 2025-12-19
New Result Intercepting API

viliamDokov:
@eupp 

From our last meeting I have the following notes:

1. I separated the different if branches in `MethodCallTransformer`.
2. I added the flag to enable/disable method call result interception in `TransformationProfile`. I also set it to true for some of the profiles (I am not 100% sure If all of the profiles I have changed should actually be changed)
Also I am not sure that the case where we are not intercepting results is well tested. It seems that in most test cases either we do not track method calls at all or we should track everything about method calls.
3. (See last commit) While messing around with test cases I found out that the tests in `FilesSpecialCasesTest` all fail (`traceDebuggingMode=true` is on).  This is because in the old interception method we sometimes intercepted the method call result in `onMethodCallReturn`. I added a hacky solution that fixes this issue, but I do not like it so maybe can help make it cleaner.


eupp (approved):

eupp:
@viliamDokov congrats with the first contribution!

PR #909, opened 2025-12-23
Draft: Add memory tracker interface and supporting cast

viliamDokov:
@eupp 

Some notes:

1. There are 2 `Utils` files. I beleive that they should be merged into one, but I decided against it for the time being because I thought it would make the rest of the rebasing easier. Now I don't think that is the case so whatever.
2. I feel like that `UnsafeNames` and `VarHandleNames` and `AtomicFieldUpdaterNames` are either already implemented somewhere or should be moved to a different place. Did not think about it too much.

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
@@ -33,7 +33,7 @@ import java.lang.management.ManagementFactory
 
 internal const val TRACE_MAGIC : Long = 0x706e547124ee5f70L
 internal const val INDEX_MAGIC : Long = TRACE_MAGIC.inv()
-internal const val TRACE_VERSION : Long = 14
+internal const val TRACE_VERSION : Long = 15
 
 // This suffix is not enforced, but IDEA plugin rely on it
 const val DATA_FILENAME_EXT = "trace"


bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -59,13 +59,16 @@ void cacheInvokeDynamicCallSite(
 
     void updateSnapshotBeforeConstructorCall(Object[] objs);
 
-    void beforeReadField(ThreadDescriptor descriptor, int codeLocation, Object obj, int fieldId);
-    void beforeReadArrayElement(ThreadDescriptor descriptor, int codeLocation, Object array, int index);
+    void beforeReadField(ThreadDescriptor descriptor, int codeLocation, Object obj, int fieldId, String typeDescriptor);
+    void beforeReadArrayElement(ThreadDescriptor descriptor, int codeLocation, Object array, int index, String typeDescriptor);
+
+    Object interceptReadResult();
```
eupp: Let's not add `interceptReadResult()` as a separate injection. 
Instead, let's implement similarly how it is done for `onMethodCall(...)` by utilizing `ResultInterceptor` API and passing the interceptor as an additional argument to `beforeReadField` and `beforeReadArrayElement`. 
You will need to modify `SharedMemoryAccessTransformer` accordingly.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -59,13 +59,16 @@ void cacheInvokeDynamicCallSite(
 
     void updateSnapshotBeforeConstructorCall(Object[] objs);
 
-    void beforeReadField(ThreadDescriptor descriptor, int codeLocation, Object obj, int fieldId);
-    void beforeReadArrayElement(ThreadDescriptor descriptor, int codeLocation, Object array, int index);
+    void beforeReadField(ThreadDescriptor descriptor, int codeLocation, Object obj, int fieldId, String typeDescriptor);
+    void beforeReadArrayElement(ThreadDescriptor descriptor, int codeLocation, Object array, int index, String typeDescriptor);
```
eupp: There is no need to add `String typeDescriptor` as an additional argument, it can be passed via `FieldDescriptor` class. 

I've already implemented alternative approach in my mirror branch.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1821,6 +1870,78 @@ internal abstract class ManagedStrategy(
         enterAnalysisSection(threadId, methodSection)
     }
 
+    private fun trackAtomicMethodMemoryAccess(
```
eupp: May I please ask you to instead make this method an extension method of `MemoryTracker` and move it to `MemoryTracker.kt` (you might need to add few more parameters to pass necessary information, e.g., `threadId`). 

`ManagedStrategy.kt` is already quite large and complicated, better to avoid enlarging it further. 

Also, I think you forgot to add a call to this method to `onMethodCall` body.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/MemoryTracker.kt
```diff
@@ -46,6 +53,76 @@ interface MemoryTracker {
 
     fun reset()
 
+    // TODO: This should probably be moved to be handled by the implementations of memory tracker when we actually add them
```
viliamDokov: As you see I have added this method directly on the interface. Not sure if it is great idea or if it should be moved to an implementation (which has not been brought in yet)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1778,19 +1778,35 @@ internal abstract class ManagedStrategy(
...
         val deterministicMethodDescriptor = getDeterministicMethodDescriptorOrNull(receiver, params, methodCallInfo)
-        val shouldInterceptAtomicMethod = atomicMethodDescriptor != null && trackAtomicMethodMemoryAccess(
-            receiver,
-            methodDescriptor.className,
-            methodDescriptor.methodName,
-            codeLocation,
-            params,
-            atomicMethodDescriptor
-        )
+        var location: MemoryLocation? = null
```
viliamDokov: I dont like how this code looks, perhaps you have a better suggestion?

eupp (approved):

PR #910, opened 2025-12-24
Rebasing the new model checker

viliamDokov:
@eupp 
Here is the rest of my limited progress on rebasing the new model checker...

Notes: 

1. It includes the commit where we add the new memory tracker
2. I have tried to keep fixes separated between commits. If some fix does not actually make sense usually other fixes that are similar should be in the same/neighboring commits.
3. Main thing that is missing/blocking is that the methods/signatures of ManagedStrategy have changed compared to the old version and I did not look into equivalent replacements in the new branch.


viliamDokov:
This is old version I will close it, since I decided to continue working on a different branch and this one is not worth saving (I can just make this branch to point to the new one, but closing the PR is easier).

PR #911, opened 2025-12-24
Implement serde for locals names

dmitrii-artuhov:


common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
```diff
@@ -12,14 +12,17 @@ package org.jetbrains.lincheck.descriptors
 
 import org.jetbrains.lincheck.trace.TraceContext
 import java.util.Objects
+import kotlin.String
+import kotlin.collections.List
```
eupp: I think this stuff is imported by default

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
```diff
@@ -12,14 +12,17 @@ package org.jetbrains.lincheck.descriptors
...
 
 class CodeLocation(
     val stackTraceElement: StackTraceElement,
     val accessPath: AccessPath? = null,
 
     // TODO: this only makes sense for method call code locations,
     //   consider introducing proper type hierarchy for code locations
-    val argumentNames: List<AccessPath?>? = null
+    val argumentNames: List<AccessPath?>? = null,
+    val activeLocals: List<String>? = null
```
eupp, line 23: Would it be difficult to pass variable ids instead (and decode them to `VariableDescriptor` on plugin side)?
I think it would be more consistent approach in the long run. 

If it is too complicated for now, please create a YT ticket.

eupp (approved):

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
dmitrii-artuhov: yup, removed

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
dmitrii-artuhov: Created an issue https://youtrack.jetbrains.com/issue/JBRes-7508/Pass-VariableDescriptors-to-the-plugin-instead-of-pure-strings-when-locals-are-calculated-for-each-trace-point
😔

dmitrii-artuhov:
@eupp Hi! I've answered on each comment, please take a look

Issue #905, opened 2025-12-18
ModelCheckingStrategy: NoSuchElementException


ben-manes:
In `3.3.2` (latest on Maven Central), I received the following error on JDK-25. A refactoring of the build had disabled the test task, where I wanted it disabled for JDK-11 due to #842 but inverted the condition. It worked fine with JDK-21 interestingly enough.

```console
./gradlew caffeine:lincheckTest -PjavaVersion=25 --tests '*UnboundedLincheckTest*' --console colored
```

```console
UnboundedLincheckTest > modelCheckingTest() FAILED
    java.util.NoSuchElementException: List contains no element matching the predicate.
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$InterleavingTreeNode.chooseUnexploredNode(ModelCheckingStrategy.kt:750)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$SwitchChoosingNode.nextInterleaving(ModelCheckingStrategy.kt:288)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$ThreadChoosingNode.nextInterleaving(ModelCheckingStrategy.kt:229)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$SwitchChoosingNode.nextInterleaving(ModelCheckingStrategy.kt:290)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$ThreadChoosingNode.nextInterleaving(ModelCheckingStrategy.kt:229)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$SwitchChoosingNode.nextInterleaving(ModelCheckingStrategy.kt:290)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy$InterleavingTreeNode.nextInterleaving(ModelCheckingStrategy.kt:159)
        at org.jetbrains.kotlinx.lincheck.strategy.managed.modelchecking.ModelCheckingStrategy.nextInvocation(ModelCheckingStrategy.kt:78)
        at org.jetbrains.kotlinx.lincheck.strategy.StrategyKt.runIteration(Strategy.kt:121)
[...]

PR #912, opened 2025-12-25
Fix ignored section constructor instrumentation

eupp:


eupp:
Partly fixes https://github.com/JetBrains/lincheck/issues/905

Issue #842, opened 2025-10-04
[3.3] Fails to progress on JDK-17 or lower


ben-manes:
I'm unsure why, but Lincheck passes on JDK-25 and spins forever on JDK-11. A Java Flight Recording is [attached](https://github.com/user-attachments/files/22704111/profile_caffeine.jfr.zip) after it was running for 15 minutes with no progress. In its thread dumps or when using jstack, it looks like Lincheck is looping and not making any progress in whatever its goal is.

<details><summary>JDK 11</summary>

```console
➜  caffeine git:(v3.dev) gradle caffeine:lincheckTest
executing gradlew instead of gradle
Parallel Configuration Cache is an incubating feature.
Configuration on demand is an incubating feature.
Reusing configuration cache.

> Task :caffeine:lincheckTest

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.BoundedBufferTest > concurrent_drain_expandOrRetry STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.BoundedBufferTest > concurrent_drain_offer STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.BoundedBufferTest > concurrent_offer STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.MpscGrowableArrayQueueLincheckTest > modelCheckingTest STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.cache.MpscGrowableArrayQueueLincheckTest > stressTest STARTED

Gradle suite > Gradle test > com.github.benmanes.caffeine.lincheck.CaffeineLincheckTest$BoundedLincheckTest > modelCheckingTest STARTED
<============-> 94% EXECUTING [26m 1s]
> :caffeine:lincheckTest > 5 tests completed
> :caffeine:lincheckTest > Executing test com...lincheck.CaffeineLincheckTest$BoundedLincheckTest
```

```console
"Test worker" #1 prio=5 os_prio=31 cpu=1024209.39ms elapsed=1331.27s tid=0x0000000105052800 nid=0x1203 runnable  [0x000000016b8e2000]
   java.lang.Thread.State: RUNNABLE
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(java.base@11.0.27/ConcurrentHashMap.java:1691)
	at org.jetbrains.lincheck.util.UnsafeHolderKt.readFieldViaUnsafe(UnsafeHolder.kt:289)
[...]

PR #913, opened 2025-12-25
Optimize class retransformation

eupp:


eupp:
Should fix https://github.com/JetBrains/lincheck/issues/842

PR #914, opened 2025-12-25
Fix unsafe field resolve bug

eupp:


eupp:
Fixes https://github.com/JetBrains/lincheck/issues/905.

With this PR and PR https://github.com/JetBrains/lincheck/pull/912 caffeine tests pass on JDK-25.

PR #918, opened 2025-12-27
[jbres-7392] Prototype trace diff

lev-serebryakov-jetbrains:
WIP!

```
<diff>.<diff>(): null at :0
    SomeClass.test(): void@0 at SomeClass.kt:64
      new SomeClass(): SomeClass@2082807696 at SomeClass.kt:64
    - SomeClass@2082807696.someMethod(1): void@0 at SomeClass.kt:64
    + SomeClass@2082807696.someMethod(3): void@0 at SomeClass.kt:64
        SomeClass.out.println("Start"): void@0 at SomeClass.kt:10
      - SomeClass.out.println(i ➜ 1): void@0 at SomeClass.kt:11
      + SomeClass.out.println(i ➜ 3): void@0 at SomeClass.kt:11
      - field = 1 at SomeClass.kt:16
      - SomeClass.out.println("Message 1"): void@0 at SomeClass.kt:17
      - callTree1(): void@0 at SomeClass.kt:18
        - callTreeField = 1 at SomeClass.kt:35
        - callTree11(): void@0 at SomeClass.kt:36
          - callTreeField = 2 at SomeClass.kt:47
          - callTree111(): void@0 at SomeClass.kt:48
            - callTreeField = 3 at SomeClass.kt:53
            - callTree1111(): void@0 at SomeClass.kt:54
              - SomeClass.out.println("Hello 4"): void@0 at SomeClass.kt:59
            - SomeClass.out.println("Hello 3"): void@0 at SomeClass.kt:55
          - SomeClass.out.println("Hello 2"): void@0 at SomeClass.kt:49
        - SomeClass.out.println("Hello 1"): void@0 at SomeClass.kt:37
      - SomeClass.out.println("Message 2"): void@0 at SomeClass.kt:19
      + field = 2 at SomeClass.kt:24
      + SomeClass.out.println("Text 1"): void@0 at SomeClass.kt:25
      + SomeClass.out.println("Text 2"): void@0 at SomeClass.kt:26
      + callTree2(): void@0 at SomeClass.kt:27
        + callTreeField = 1 at SomeClass.kt:41
        + callTree11(): void@0 at SomeClass.kt:42
          + callTreeField = 2 at SomeClass.kt:47
          + callTree111(): void@0 at SomeClass.kt:48
            + callTreeField = 3 at SomeClass.kt:53
            + callTree1111(): void@0 at SomeClass.kt:54
              + SomeClass.out.println("Hello 4"): void@0 at SomeClass.kt:59
            + SomeClass.out.println("Hello 3"): void@0 at SomeClass.kt:55
          + SomeClass.out.println("Hello 2"): void@0 at SomeClass.kt:49
        + SomeClass.out.println("Hello 1"): void@0 at SomeClass.kt:43
      + SomeClass.out.println("Text 3"): void@0 at SomeClass.kt:28
        SomeClass.out.println("End"): void@0 at SomeClass.kt:31
```

### DIFF
--- a/trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
+++ b/trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
@@ -15,9 +15,9 @@ import org.jetbrains.lincheck.trace.*
 import java.io.DataOutputStream
 import java.io.File
 import java.io.FileOutputStream
+import java.util.concurrent.ThreadLocalRandom
 
-private typealias ThreadMap = Map<String, Triple<Int, Int, Int>>
-
+internal data class ThreadIds(val leftId: Int, val rightId: Int, val diffId: Int)
 
 fun diffTwoTraces(left: LazyTraceReader, right: LazyTraceReader, outputBaseName: String) {
     require(!left.isDiff) { "Cannot diff other diffs: left trace is diff" }
@@ -140,22 +140,22 @@ private fun diffOneThread(
 private fun correlateThreadsByName(
     left: LazyTraceReader,
     right: LazyTraceReader
-): ThreadMap {
-    val threadMap = mutableMapOf<String, Triple<Int, Int, Int>>()
+): Map<String, ThreadIds> {
+    val threadMap = mutableMapOf<String, ThreadIds>()
     var diffThreadId = 0
     left.context.threadNames()
         .forEach { tn ->
-            threadMap[tn] = Triple(left.context.getThreadId(tn), right.context.getThreadId(tn), diffThreadId++)
+            threadMap[tn] = ThreadIds(left.context.getThreadId(tn), right.context.getThreadId(tn), diffThreadId++)
         }
     right.context.threadNames()
         .filter { !threadMap.contains(it) }
         .forEach { tn ->
-            threadMap[tn] = Triple(left.context.getThreadId(tn), right.context.getThreadId(tn), diffThreadId++)
+            threadMap[tn] = ThreadIds(left.context.getThreadId(tn), right.context.getThreadId(tn), diffThreadId++)
         }
     return threadMap
 }
 
-private fun saveThreadMap(threadMap: ThreadMap): File {
+private fun saveThreadMap(threadMap: Map<String, ThreadIds>): File {
     val threadMapFile = File.createTempFile("trace-diff-", ".$THREAD_MAP_FILENAME_EXT")
         .also { it.deleteOnExit() }
     val threadMapStream = FileOutputStream(threadMapFile).buffered(OUTPUT_BUFFER_SIZE)
@@ -163,9 +163,9 @@ private fun saveThreadMap(threadMap: ThreadMap): File {
         val out = DataOutputStream(it)
         out.writeInt(threadMap.size)
         threadMap.forEach { (_, triple) ->
-            out.writeInt(triple.third)
-            out.writeInt(triple.first)
-            out.writeInt(triple.second)
+            out.writeInt(triple.diffId)
+            out.writeInt(triple.leftId)
+            out.writeInt(triple.rightId)
         }
     }
     return threadMapFile


lev-serebryakov-jetbrains:
It will be refactored in many places, don't comment on style yet :)

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -252,67 +246,164 @@ class LazyTraceReader private constructor(
...
-                    val dataEntry = zip?.nextEntry ?: throwZipError(baseFileName)
-                    if (dataEntry.name != PACKED_DATA_ITEM_NAME) throwZipError(baseFileName)
-                    val tmpDataFile = File.createTempFile("unpacked-trace-data", ".$DATA_FILENAME_EXT")
-                    tmpDataFile.deleteOnExit()
-
-                    try {
-                        val dataOutput = openNewFile(tmpDataFile.absolutePath)
-                        dataOutput.use {
-                            zip?.copyTo(it)
+                for (entry in zip.entries()) {
```
eupp: The logic here becomes more complex, it is hard to grasp what is the expected zip file structure.  Could you please document it.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -252,67 +246,164 @@ class LazyTraceReader private constructor(
      *    ot ZIP stream prepared for reading index entry. Index is only read sequential, so reading directly
      *    from ZIP is Ok.
      *  - [metaInfo] — Trace meta information, if it exists for provided trace.
+     *  - [threadIdMap] — Map from diff's thread id to left (first) and right (second) thread ids, `-1` if thread
+     *    doesn't present in corresponding trace. `null` if trace is not diff.
+     *  - [eventIdMap] — Map from diff's event id to left (first) and right (second) event ids, `-1` if this event
+     *    doesn't present in corresponding trace. `null` if trace is not diff.
      */
-    private class TraceDataProvider(val baseFileName: String): AutoCloseable {
+    private class TraceDataProvider(val baseFileName: String) : AutoCloseable {
```
eupp: This class becomes larger and more complex. Let's extract it into a separate file (`private` can be relaxed to `internal`).

eupp: Also, it looks like the zip packing/unpacking logic is split between this class and `SerializationPacker.kt`. 

Let's group it together in a single place, because pack/unpack are toughly coupled. Otherwise, I have a concern we will get more bugs in the future when somebody will update packing logic, but not unpacking, and vice versa.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -371,11 +476,16 @@ class LazyTraceReader private constructor(
         input.close()
     }
 
+    @Synchronized
```
eupp, line 299: Is it required for this PR to function properly? If not, I would suggest this change (and other related changes in `LazyTraceReader`) going into a separate PR.  

The reasons are as follows:

1. We already noticed this problem before and added a fix on plugin side (I'll share MR link to private repo). We need to synchronize this logic across lincheck/plugin codebase. 
2. I'll prefer to have a look at this separately, to review `LazyTraceReader` API in more detail and re-check that we have locks on all its methods where they are needed.

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -726,6 +844,20 @@ class LazyTraceReader private constructor(
             if (input == null) return null
             return DataInputStream(input.buffered(INPUT_BUFFER_SIZE))
         }
+
+        fun copyOutToTempFile(prefix: String, suffix: String, input: InputStream): Result<File> {
```
eupp: It looks like this is a generic utility function, can we decouple it from `LazyTraceReader` companion and make it just top-level function?

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -252,67 +246,164 @@ class LazyTraceReader private constructor(
...
+            val maybeDiff = threadIdMap != null
+                    || idMapStream != null
+                    || leftMetaInfo != null
+                    || rightMetaInfo != null
+
+            val totalDiff = threadIdMap != null
+                    && idMapStream != null
+                    && leftMetaInfo != null
+                    && rightMetaInfo != null
+
```
```suggestion
            val maybeDiff = 
                   threadIdMap != null
                || idMapStream != null
                || leftMetaInfo != null
                || rightMetaInfo != null

            val totalDiff = 
                   threadIdMap != null
                && idMapStream != null
                && leftMetaInfo != null
                && rightMetaInfo != null

```

trace/src/main/org/jetbrains/lincheck/trace/Deserialization.kt
```diff
@@ -72,16 +62,16 @@ private class DataBlock(
...
         this.physicalStart = physicalStart
         this.physicalEnd = physicalEnd
         this.accDataSize = accDataSize
     }
 
-    fun coversPhysicalOffset(offset: Long): Boolean = offset in physicalStart ..< physicalEnd
+    fun coversPhysicalOffset(offset: Long): Boolean = offset in physicalStart..<physicalEnd
 
-    fun coversLogicalOffset(offset: Long): Boolean = offset in logicalDataStart ..< logicalDataStart + dataSize
+    fun coversLogicalOffset(offset: Long): Boolean = offset in logicalDataStart..<logicalDataStart + dataSize
```
eupp, line 75: Is it necessary? I prefer ` ..< ` with spaces, otherwise it is kinda inconsistent compared to other binary operators in Kotlin.

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -448,6 +493,10 @@ internal fun DataInput.readDiffStatus(): DiffStatus? {
     return values[ordinal]
 }
 
+internal fun openNewStandardDataAndIndex(baseFileName: String): Pair<OutputStream, OutputStream> {
```
eupp: Please add `data class` with meaningful field names, instead of `Pair`

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -56,11 +56,24 @@ enum class DiffStatus {
...
+    /**
+     * Tracepoint is tracepoint from right (new) trace which was edited in respect with left (old) trace.
+     *
+     * It can be seen as "added" if no editing information is needed, and is pair for previous sibling which should be
+     * [EDITED_OLD].
+
      * For example, if it is a method called tracepoint, it has the same method in both traces but differs in arguments values.
      */
-    EDITED,
+    EDITED_NEW,
```
eupp, line 79: Please document the contract of EDITED_OLD and EDITED_NEW, and where the children are stored.

trace/src/main/org/jetbrains/lincheck/trace/diff/TracePointCloner.kt
```diff
@@ -0,0 +1,218 @@
...
+ * with this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.jetbrains.lincheck.trace.diff
+
+import org.jetbrains.lincheck.descriptors.*
+import org.jetbrains.lincheck.trace.*
+import java.io.DataOutput
+
+class TracePointCloner(
```
eupp, line 43: Where does it clone the tracepoints? 
To `idMapOutput`? Please document.

trace/src/main/org/jetbrains/lincheck/trace/diff/TracePointCloner.kt
```diff
@@ -0,0 +1,218 @@
...
+    private val context: TraceContext,
+    private val idMapOutput: DataOutput,
```
eupp, line 45: Why `idMapOutput`? What does `idMap` stands for?

trace/src/main/org/jetbrains/lincheck/trace/diff/TracePointComparator.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    private fun Hasher.addTRList(list: List<TRObject?>): Hasher {
+        list.forEach { add(it) }
+        return this
+    }
+}
+
+// mzHash64 instead of standard Java hash to expand it to long
+// https://github.com/matteo65/mzHash64
+private class Hasher {
```
eupp: I suppose this was copied from another OSS project?

If so, then the policy is as follows:

- Copy the whole file with the same name. (I would also suggest putting it into `common/util`).
- Add the license from original repo to the header of the file.  
- Add a comment at the beginning of the file:
"Copied and modified from: <link to file in target repo>"

See `WeakIdentityHashMap.java` as an example.

trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
```diff
@@ -0,0 +1,248 @@
...
+
+package org.jetbrains.lincheck.trace.diff
+
+import org.jetbrains.lincheck.descriptors.Types
+import org.jetbrains.lincheck.trace.*
+import java.io.DataOutputStream
+import java.io.File
+import java.io.FileOutputStream
+
+fun diffTwoTraces(left: LazyTraceReader, right: LazyTraceReader, outputBaseName: String) {
```
eupp, line 22: Please briefly describe the algorithm in the comment in plain English. 

From what I understood, the following scheme is used:
- On single tree level (list of trace points), a regular Myers algoritm for lists is used (diffing list of trace points without taking their children in consideration).  
- For trace points for which Myers reported "edit" action, the algorithm is applied recursively to their childrens lists. 
- For trace points for which Myers reported add/remove, the whole subtree is treated as added/removed.

trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
```diff
@@ -0,0 +1,248 @@
...
+    val threadMapStream = FileOutputStream(threadMapFile).buffered(OUTPUT_BUFFER_SIZE)
+    threadMapStream.use {
+        val out = DataOutputStream(it)
+        out.writeInt(threadMap.size)
+        threadMap.forEach { (_, triple) ->
+            out.writeInt(triple.third)
+            out.writeInt(triple.first)
+            out.writeInt(triple.second)
+        }
+    }
```
eupp: Please extract thread matching into separate function.

trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
```diff
@@ -0,0 +1,248 @@
...
+
+    val idMapFile = File.createTempFile("trace-diff-", ".$ID_MAP_FILENAME_EXT")
+        .also { it.deleteOnExit() }
+    val idMapStream = DataOutputStream(FileOutputStream(idMapFile).buffered(OUTPUT_BUFFER_SIZE))
+
+    val cloner = TracePointCloner(output.context,idMapStream)
+    output.use {
+        // Ok, we have all threads matched, work on them one by one.
+        threadMap.forEach { name, (leftThreadId, rightThreadId, outputThreadId) ->
+            cloner.setThread(outputThreadId)
```
eupp, line 51: please extract thread handling into separate function

trace/src/main/org/jetbrains/lincheck/trace/diff/MyersDiff.kt
```diff
@@ -0,0 +1,93 @@
...
+    if (n == 0 && m == 0) return emptyList()
+
+    val max = n + m
+    val v = IntArray(2 * max + 1)
+    val trace = mutableListOf<IntArray>()
+
+    v[max + 1] = 0
+
+    var x: Int
+    var y: Int
```
eupp: Having short var names for algorithmic code is ok, but let's add a comment to explain their semantics. 

Something like:
```text
/* Abbrevations:
 * n - length of left list
 * m - length of right list
 * x - ...
 * y - ...
 * d - ...
 * ...
 */
```

trace/src/main/org/jetbrains/lincheck/trace/diff/MyersDiff.kt
```diff
@@ -0,0 +1,93 @@
...
+ * This implementation uses Myers' diff algorithm.
+ */
+internal fun <T> diffLists(left: List<T>, right: List<T>, equals: (T, T) -> Boolean): List<DiffLine> {
+    val n = left.size
+    val m = right.size
+    if (n == 0 && m == 0) return emptyList()
+
+    val max = n + m
+    val v = IntArray(2 * max + 1)
+    val trace = mutableListOf<IntArray>()
```
eupp: Using `trace` name is confusing here, because of confusion with our trace data. Let's rename it somehow.

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -31,17 +29,27 @@ val UNKNOWN_METHOD_TYPE = Types.MethodType(Types.VOID_TYPE, Types.VOID_TYPE)
 private val EMPTY_STACK_TRACE = StackTraceElement("", "", "", 0)
 
 class TraceContext {
-    private val threadNames = ConcurrentHashMap<Int, String>()
+    private val threadsId2Name = mutableMapOf<Int, String>()
+    private val threadsName2Id = mutableMapOf<String, Int>()
```
eupp: If id-by-name-lookup operation is not frequent operation, I would suggest to just use something like `values.find { ... }`.

eupp: In any case, let's keep `threadNames` name, and for the second map use something like `threadIds` or `threadIdByName`.

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -31,17 +29,27 @@ val UNKNOWN_METHOD_TYPE = Types.MethodType(Types.VOID_TYPE, Types.VOID_TYPE)
...
+            threadsName2Id[name] = id
+        }
+    }
+
+    fun getThreadName(id: Int): String = synchronized(threadsId2Name) { threadsId2Name[id] ?: "" }
+
+    fun getThreadId(name: String): Int = synchronized(threadsId2Name) { threadsName2Id[name] ?: -1 }
 
-    fun getThreadName(id: Int): String = threadNames[id] ?: ""
+    val threadNames: List<String> get() = synchronized(threadsId2Name) { threadsId2Name.values.toList() }
```
eupp: IMO, `val` should be used only for "cheap" operations, where essentially you can treat it as field access. For "heavy" operations (especially when you create a copy of the accessed data) better to use function:

```kotlin
fun getThreadNames(): List<String> = ...
```

eupp:
Also, don't forget to update `TRACE_VERSION`.

lev-serebryakov-jetbrains:
> Also, don't forget to update `TRACE_VERSION`.

It is not needed, changes are backward-compatible, it will read old version Ok, and I want to leave it as is.

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -43,6 +41,10 @@ class TraceContext {
 
     fun getThreadName(id: Int): String = threadNames[id] ?: ""
 
+    fun getThreadId(name: String): Int = threadNames.filter { (_, value) -> value == name }.keys.firstOrNull() ?: -1
```
eupp, line 44:
```suggestion
fun getThreadId(name: String): Int = threadNames.entries.find { (_, value) -> value == name }?.key ?: -1
```

trace/src/main/org/jetbrains/lincheck/trace/diff/TraceDiff.kt
```diff
@@ -0,0 +1,285 @@
...
+
+package org.jetbrains.lincheck.trace.diff
+
+import org.jetbrains.lincheck.descriptors.Types
+import org.jetbrains.lincheck.trace.*
+import java.io.DataOutputStream
+import java.io.File
+import java.io.FileOutputStream
+
+private typealias ThreadMap = Map<String, Triple<Int, Int, Int>>
```
eupp: Please also use custom data type here instead of `Triple`.

eupp (approved):
Please fix few final nitpicks before merge.

PR #921, opened 2026-01-14
Implement new loop detector utilizing CFG information

acturcu:
@eupp 

This PR contains the counter-based LoopDetector APIs. The old method of detecting loops has been removed. 

Trace reporting and representation tests have been updated.


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
+    fun onMethodExit(
+        threadDescriptor: ThreadDescriptor,
+        methodId: Int,
+        receiver: Any?,
+        params: Array<Any?>,
+        result: Any?,
+    )
+
+    // TODO But, where do we call this???? beforeWriteField, beforeWriteArrayElement, afterLocalWrite (which is empty)??????
+    fun onStageChange(threadDescriptor: ThreadDescriptor, codeLocation: Int, variableId: Int, valueHash: Int)
```
eupp: We might need some API to track variables/fields changes for integrating with a new model checker, or for smarter loop detection algorithm in general. But for the first baseline implementation we won't need that. So let's remove it for now, and design this API when we will actually need that.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-            interleavingHistory = spinCycleInterleavingHistory,
-            failDueToDeadlockInTheEnd = failDueToDeadlockInTheEnd
-        )
-    }
+data class ActiveLoop (
+    val loopId: Int,
+    var iterationCount: Int = 0,
+    var switched: Boolean = false,
+    var stuck: Boolean = false,
+    var lastSignature: Int = 0,
```
eupp: I think we can just keep `iterationCount` always increasing, and return `Switch` decision when `iterationCount % N == 0` and `Stuck` when `iterationCount >= M`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
+    var progressSignature: Int = 1,
```
eupp: Let's postpone this idea for the next iterations.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-     * In the considered example, we will retain that we will switch soon after
-     * the spin cycle in thread 1, so no bug will appear.
-     */
-    fun shouldSwitch(): Boolean {
-        return replayModeLoopDetectorHelper?.shouldSwitch() ?: false
-    }
+data class ActiveMethodCall (
+    val methodId: Int,
+    var depth: Int = 1,
+    val params: Array<Any?>? = null,
```
eupp: Le'ts postpone parameters tracking to next iterations.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-        // Remove references to the interleaving tree
-        interleavingsLeadToSpinLockSet.clear()
-        loopTrackingCursor.clear()
-
-        replayModeLoopDetectorHelper = ReplayModeLoopDetectorHelper(
-            interleavingHistory = spinCycleInterleavingHistory,
-            failDueToDeadlockInTheEnd = failDueToDeadlockInTheEnd
-        )
-    }
+data class ActiveLoop (
```
eupp:
```suggestion
data class ActiveLoopInfo(
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-     * This leads to the results of another execution, compared to the original failure results.
-     * To avoid this bug when we're replaying some executions,
-     * we have to follow only loop detector's history during the last execution.
-     * In the considered example, we will retain that we will switch soon after
-     * the spin cycle in thread 1, so no bug will appear.
-     */
-    fun shouldSwitch(): Boolean {
-        return replayModeLoopDetectorHelper?.shouldSwitch() ?: false
-    }
+data class ActiveMethodCall (
```
eupp:
```suggestion
data class ActiveMethodCallInfo(
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-    }
+import sun.nio.ch.lincheck.ThreadDescriptor
+import java.util.IdentityHashMap
+
+interface LoopDetector {
+    enum class Decision {
+        IDLE,
+        SWITCH_THREAD, //n iterations
+        STUCK // m iterations
+    }
```
eupp, line 31:
```suggestion
}


```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-        currentInterleavingHistory.clear()
-        replayModeLoopDetectorHelper?.reset()
-    }
+import sun.nio.ch.lincheck.ThreadDescriptor
+import java.util.IdentityHashMap
+
+interface LoopDetector {
+    enum class Decision {
+        IDLE,
+        SWITCH_THREAD, //n iterations
```
eupp:
```suggestion
SWITCH_THREAD, // N iterations
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
+        STUCK // m iterations
```
eupp:
```suggestion
STUCK // M iterations
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
-
-        /**
-         * This decision is returned when a single thread enters a live-lock,
-         * and a switch to other threads is required to see if they can unblock the execution.
-         *
-         * @property cyclePeriod The period of the spin-loop cycle.
-         */
-        data class LivelockThreadSwitch(val cyclePeriod: Int) : Decision()
-    }
+data class LoopThreadState (
```
eupp:
```suggestion
data class LoopDetectorThreadState(
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,219 @@
...
 
-    /**
-     * Should be called before each read non-local array element write.
-     */
-    fun beforeWriteArrayElement(array: Any, index: Int, value: Any?) {
-        passValue(array)
-        passValue(index)
-        passValue(value)
+        loop.progressSignature = loop.progressSignature * 31 + variableId * 17 + valueHash
     }
```
eupp, line 123: Postpone (remove or comment)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1931,12 +1880,54 @@ internal abstract class ManagedStrategy(
...
+                LoopDetector.Decision.IDLE -> {}
+                LoopDetector.Decision.SWITCH_THREAD -> {
+                    onSwitchPoint(threadId)
+                    switchCurrentThread(threadId, BlockingReason.Waiting)
+                }
+
+                LoopDetector.Decision.STUCK -> {
+                    // What to do here? we are stuck in a loop, not really sure how to proceed
+                    onSwitchPoint(threadId)
+                    switchCurrentThread(threadId, BlockingReason.LiveLocked)
```
eupp: Add new invocation result type `LivelockInvocationResult`. 
Create new function `failDueToLivelock` similar to `failDueToDeadlock` and call it here.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1931,12 +1880,54 @@ internal abstract class ManagedStrategy(
...
+            val threadId = threadScheduler.getCurrentThreadId()
+            val methodInvocationId = currentMethodInvocationId(threadId)
+
+            val decision = loopDetector.onLoopIteration(threadDescriptor, codeLocation, loopId, methodInvocationId)
+
+            when(decision) {
+                LoopDetector.Decision.IDLE -> {}
+                LoopDetector.Decision.SWITCH_THREAD -> {
+                    onSwitchPoint(threadId)
+                    switchCurrentThread(threadId, BlockingReason.Waiting)
```
eupp:
```suggestion
switchCurrentThread(threadId, BlockingReason.LiveLocked)
```

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceNodes.kt
```diff
@@ -135,6 +135,61 @@ internal class ResultNode(val actorResult: ReturnedValueResult, eventNumber: Int
...
+    override val tracePoint: LoopIterationTracePoint get() = super.tracePoint as LoopIterationTracePoint
+
+    override fun toStringImpl(withLocation: Boolean): String =
+        tracePoint.toStringImpl(withLocation)
+
+    override fun copy(): TraceNode = IterationNode(tracePoint, eventNumber)
+}
+
+// Used to print <iterations from-to> nodes when compressing iterations
+internal class IterationRangeNode(
```
eupp: Let's keep single `IterationNode` class, and instead allow it to represent both a single iteration or a range of iterations. You can have multiple constructors of the class to create node for single iteration or range, and define helper functions like `fun isIterationRange(): Boolean`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceNodes.kt
```diff
@@ -143,6 +198,19 @@ internal fun traceToTree(threadCount: Int, trace: Trace): MultiThreadedTable<Tra
     val nodes = MutableList<MutableList<TraceNode>>(threadCount) { mutableListOf() }
     val currentNodePerThread = MutableList<CallNode?>(threadCount) { null }
 
+    val currentLoopPerThread = MutableList<LoopNode?>(threadCount) { null }
+    val currentIterPerThread = MutableList<IterationNode?>(threadCount) { null }
```
eupp: Could you keep single `currentNodePerThread` thing, and instead change its type to `MutableList<TraceNode>` ?

I think there should be no difference between the type of the container, whether it is call node, loop node, or loop iteration node --- the behavior is basically the same.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceNodes.kt
```diff
@@ -185,3 +269,97 @@ internal fun traceToTree(threadCount: Int, trace: Trace): MultiThreadedTable<Tra
 
     return nodes
 }
+
+internal fun foldEquivalentLoopIterations(
```
eupp: Please move to `trace/Utils.kt`

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/Utils.kt
```diff
@@ -149,62 +198,64 @@ internal fun Trace.moveStartingSwitchPointsOutOfMethodCalls(): Trace {
...
-            --j
-        }
-
-        newTrace.move(i, j)
-    }
-
-    return Trace(newTrace, this.threadNames)
-}
+//TODO: refactor to work for new loop detector
+//internal fun Trace.moveSpinCycleStartTracePoints(): Trace {
```
eupp: I think we don't need that anymore, let's just remove this commented code.

src/jvm/main/org/jetbrains/kotlinx/lincheck/IdeaPlugin.kt
```diff
@@ -334,6 +334,7 @@ private val LincheckFailure.type: String
         is UnexpectedExceptionFailure -> "UNEXPECTED_EXCEPTION"
         is ValidationFailure -> "VALIDATION_FAILURE"
         is ManagedDeadlockFailure, is TimeoutFailure -> "DEADLOCK"
+        is ManagedLivelockFailure -> "LIVELOCK_SPINLOOP"
```
eupp:
```suggestion
        is ManagedLivelockFailure -> "LIVELOCK"
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/Utils.kt
```diff
@@ -108,6 +108,55 @@ internal fun Trace.moveStartingSwitchPointsOutOfMethodCalls(): Trace {
     return Trace(newTrace, this.threadNames)
 }
 
+/**
+ * Idea is to convert unclosed loops method calls to [SpinCycleStartTracePoint]
+ */
+internal fun Trace.convertOpenLoopsToSpinCycles(): Trace {
```
eupp: Do we need this?

I think that instead, if we detect an unclosed loop at some point,  we should just close it at this point. For instance, if we are gonna to exit a method call with unfinished loops, we just close all the loops at this point. 

It is still not perfect, as we miss actual loop bounds in the trace, but it simplifies all other code, as we don't need to deal with unclosed loops anywhere else.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -737,11 +680,18 @@ internal abstract class ManagedStrategy(
         // scenario threads are handled separately by the runner itself
         if (currentThreadId < nScenarioThreads) return
 
+        val methodId = context.getOrCreateMethodId(
+            className = "java.lang.Thread",
+            methodName = "run",
+            methodType = Types.MethodType(Types.VOID_TYPE)
+        )
+        popMethodId(currentThreadId, methodId)
```
eupp: Let's instead add methods to loop detector:
```
fun beforeThreadRun(...)
fun afterThreadRun(...)
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -144,6 +144,9 @@ internal abstract class ManagedStrategy(
     // Stores the currently executing methods call stack for each thread.
     private val callStackTrace = mutableThreadMapOf<MutableList<CallStackTraceElement>>()
 
+    // Stores the methods calls for each thread, to be used by LoopDetector.
+    private val methodIdStack = mutableThreadMapOf<ArrayDeque<Int>>()
```
eupp: Please this instead to `LoopDetector`. 

You already have there `LoopDetectorThreadState` data, so you need to just re-use that.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2059,12 +2040,123 @@ internal abstract class ManagedStrategy(
...
+    private fun getMethodId(threadId: Int): Int {
+        return methodIdStack[threadId]?.lastOrNull() ?: -1
+    }
+
+    data class Loop(
+        val threadId: Int,
+        val loopId: Int,
+    )
+
+    private val activeLoops = mutableSetOf<Loop>()
```
eupp: All of this should be in `LoopDetector`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2059,12 +2040,123 @@ internal abstract class ManagedStrategy(
...
+
+            LoopDetector.Decision.STUCK -> {
+                failDueToLivelock()
+            }
+        }
     }
 
-    override fun afterLoopExit(threadDescriptor: ThreadDescriptor, codeLocation: Int, loopId: Int, exception: Throwable?, isReachableFromOutsideLoop: Boolean) {
-        error("Lincheck managed strategy does not support CFG-based loops tracking.")
+    override fun afterLoopExit(
```
eupp, line 2176: One important thing about `afterLoopExit` contract that I forgot to mention.

Due to details of how CFG (control-flow graph) analysis is working, `afterLoopExit` in some cases may be called not from inside the loop. This happens, because it is injected on any basic-block that CAN be a target of a jump from within loop body, but in general there can be jumps from outside the loop. 

This is why there is `isReachableFromOutsideLoop` flag. This flag tells if this `afterLoopExit` call is always called only from within a loop body, or not. And, in case if it is `false`, you need to check at runtime (that is, here, in `ManagedStrategy`) if there is currently a loop body running or not. If not, you should avoid closing the loop, because it means this injection was called from outside of the loop.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/Utils.kt
acturcu: I think with the new updates, I can completely remove it. I'll check

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceNodes.kt
acturcu: Yes, I've modified it in [1632b96](https://github.com/JetBrains/lincheck/pull/921/commits/1632b96b0a313a3f5171ea0e6d0c70038c522523)

src/jvm/test/resources/expected_logs/run_concurrent_test/kotlin_thread.txt
```diff
@@ -7,40 +7,46 @@ java.lang.IllegalStateException: Check failed.
...
-|   result: IllegalStateException #1                     |                                                         |
-| ---------------------------------------------------------------------------------------------------------------- |
+| ----------------------------------------------------------------------------------------------------------------- |
+|                      Main Thread                       |                         Thread 1                         |
+| ----------------------------------------------------------------------------------------------------------------- |
+| $this_runCatching.block(): threw IllegalStateException |                                                          |
+|   block(): threw IllegalStateException                 |                                                          |
+|     thread(block = Lambda#1): Thread#1                 |                                                          |
+|     switch (reason: waiting for Thread 1 to finish)    |                                                          |
+|                                                        | run() is looping                                         |
```
eupp: Let's remove `is looping` suffix (for simplicity of code, and also because it does not add new information). 
In some cases you might want to print `: <hung>` to indicate that function get stuck in a loop.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
```diff
@@ -38,6 +38,14 @@ internal class ManagedDeadlockFailure(
     analysisProfile: AnalysisProfile
 ) : LincheckFailure(scenario,results, trace, analysisProfile)
 
+//TODO: verify livelock handling
```
eupp: What about this TODO?
Can we remove it?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -81,6 +81,13 @@ internal class LoopTransformer(
     private val opcodesReachableFromOutsideLoops: Map<InstructionIndex, Set<LoopId>> =
         methodInfo.basicControlFlowGraph!!.computeReachabilityFromOutsideLoops(insnIndexRemapping, loopInfo)
 
+    private val isReducible = methodInfo.basicControlFlowGraph!!.isReducible ?: true
```
eupp:
```suggestion
    private val isReducible = methodInfo.basicControlFlowGraph!!.isReducible ?: true
    
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -576,6 +576,7 @@ object ModelCheckingDefaultTransformationProfile : TransformationProfile {
 
             trackCoroutineSuspensions = true
             interceptCoroutineDelays = true
+            trackLoops = true
```
eupp, line 456: Let's also add a flag `trackIrreducibleLoops` and enable it only for model checking profile.

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -84,6 +84,7 @@ void cacheInvokeDynamicCallSite(
 
     void onLoopIteration(ThreadDescriptor descriptor, int codeLocation, int loopId);
     void afterLoopExit(ThreadDescriptor descriptor, int codeLocation, int loopId, Throwable exception, boolean isReachableFromOutsideLoop);
+    void onIrreducibleLoop(ThreadDescriptor descriptor, int codeLocation);
```
eupp: Let's rename to `onIrreducibleLoopIteration`. 

Also, let's please reorder method declarations as such:
```java
void onLoopIteration(ThreadDescriptor descriptor, int codeLocation, int loopId);
void onIrreducibleLoop(ThreadDescriptor descriptor, int codeLocation);
void afterLoopExit(ThreadDescriptor descriptor, int codeLocation, int loopId, Throwable exception, boolean isReachableFromOutsideLoop);
```

bootstrap/src/sun/nio/ch/lincheck/Injections.java
```diff
@@ -1086,6 +1086,15 @@ public static void afterLoopExit(ThreadDescriptor descriptor, int codeLocation,
...
     }
 
+    /**
+     * Called at the beginning of every loop iteration for irreducible loops (loops with multiple entries)
+     */
+    public static void onIrreducibleLoop(ThreadDescriptor descriptor, int codeLocation) {
+        EventTracker tracker = getEventTracker(descriptor);
+        if (tracker == null || descriptor == null) return;
+        tracker.onIrreducibleLoop(descriptor, codeLocation);
+    }
```
eupp, line 1078: Same comment here about order of methods

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,319 @@
...
 
-    private fun updateInterleavingHistory(codeLocation: Int) {
-        val lastInterleavingHistoryNode = currentInterleavingHistory.last()
-        if (lastInterleavingHistoryNode.cycleOccurred) {
-            return /* If we already ran into cycle and haven't switched than no need to track executions */
-        }
-        lastInterleavingHistoryNode.addExecution(codeLocation)
-        loopTrackingCursor.onNextExecutionPoint()
-    }
+    fun onIrreducibleLoop(threadId: Int, codeLocation: Int): Decision
```
eupp: Let's rename to `onIrreducibleLoopIteration` and move next to `onLoopIteration`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,319 @@
...
-     */
-    private fun passParameters(params: Array<Any?>) {
-        if (mode == Mode.DEFAULT) return
-        params.forEach { param ->
-            val hash = primitiveOrIdentityHashCode(param)
-            currentThreadCodeLocationsHistory += CodeIdentity.ValueRepresentationIdentity(hash)
-        }
-    }
+    fun currentMethodId(threadId: Int): Int
+    fun loopIsInStack(threadId: Int, loopId: Int) : Boolean
```
eupp: General convention in Java/Kotlin is to put `is` at the beginning of method name:

```kotlin
fun isLoopInStack(threadId: Int, loopId: Int) : Boolean
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,319 @@
...
+) : LoopDetector {
+
+    // Maintain the following state:
+    //    threadId -> Stack<ActiveMethodCall>
+    //
+    //    ActiveMethodCall := (methodId, params?, Stack<ActiveLoop>)
+    //    ActiveLoop := (loopId, iterations)
+
+    // Questions: what methods we need to add to the interface to connect it with scheduling?
+    // shouldSwitch(): Boolean ???
```
eupp: Let's remove this comment?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,319 @@
...
+
+    private val threadStates = mutableThreadMapOf<LoopDetectorThreadState>()
+    // Per-thread stack of currently active loops
+    private val activeLoopStack = mutableThreadMapOf<ArrayDeque<LoopContext>>()
```
eupp: Do we need to have both `activeLoopStack` and `val loops: ArrayDeque<ActiveLoopInfo>` as part of `ActiveMethodCallInfo` ?

I think one should be enough. Better to leave `loops` as part of method call stack info, as this way it reflects the nestedness structure (loops inside method calls).

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
eupp: Also, let's add `loopId` argument to `onIrreducibleLoopIteration` too.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -10,975 +10,319 @@
...
-    /**
-     * Called before next thread switch
-     */
-    fun onNextSwitch() {
-        currentInterleavingNodeIndex++
-        // See threadsRan field description to understand the following initialization logic
-        executionsPerformedInCurrentThread = 0
-    }
+        // For irreducible loops we can use the placeholder loop ID -2, to make sure that they are not mixed with
+        // regular loops. We also don't need to track them in the loop stack, as they are not nested and don't have enter and exit events.
```
eupp: Actually, for irreducible loops we still can detect loop exits events. 
Moreover, the `afterLoopExit` instrumentation injection code should remain same for both reducible and irreducible loops. We also can pass real `loopId` to irreducible loops too. And we can store them in the stack same way as regular loops. 

The only difference between them is that for irreducible loops we cannot reliable determine loop iteration boundaries.

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceReporter.kt
```diff
@@ -209,12 +211,66 @@ private class TraceColumnPrinter(
             return
         }
         updateSpinCycleState(node)
+//        TODO: Iterations with ranges do not show up. In those cases the node is CallNode.
+        if ((node is LoopNode || node is CallNode) && !verbose) {
```
eupp: Am I right that the purpose of all of this new code is to display these `┌╶>` arrows? 

If yes --- then we can delete it, as well as `updateSpinCycleState()` and everything else related to these arrows. 

We don't need these arrows in the output anymore, because now we have properly structured `loop` and `iteration` trace nodes in the trace tree. Before we used these arrows only because we didn't have the CFG analysis to determine loop boundaries precisely.

src/jvm/main/org/jetbrains/kotlinx/lincheck/Reporter.kt
```diff
@@ -731,6 +733,14 @@ private fun Appendable.appendDeadlockMessageIfRequired(failure: LincheckFailure)
     }
 }
 
+//TODO: check if usage is correct
```
eupp: What about this TODO? 
Can we remove it?

src/jvm/main/org/jetbrains/kotlinx/lincheck/trace/TraceFiltering.kt
```diff
@@ -14,19 +14,26 @@ import org.jetbrains.lincheck.descriptors.CodeLocations
...
-        if (callNode.isRootCall && callNode.tracePoint.isThreadStart) return true
+//    TODO: see what else can be used to decide unfolding (LoopNode for example)
+    override fun shouldUnfold(callNode: TraceNode): Boolean {
+        if (callNode is CallNode && callNode.isRootCall && callNode.tracePoint.isThreadStart) return true
         unfoldableNodes[callNode]?.let { return it }
+
+        if (callNode.children.lastOrNull() is LoopNode || callNode.children.lastOrNull() is RecursionNode) {
+            unfoldableNodes[callNode] = true
+            return true
+        }
```
eupp: Can we remove this `if` statement?

I think because of it, almost every trace node that have loops or recursion gets unfolded, and we don't wanna that in case of short trace.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/LincheckFailure.kt
acturcu: yes, it was added when i created the failure motive

eupp (approved):

PR #970, opened 2026-03-01
Support Spring framework integration

zhelenskiy:
Support the Spring framework by initiating recording from the program's startup and ending with its shutdown.
Additionally, address and resolve several minor bugs.

### DIFF
--- a/common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
+++ b/common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
@@ -22,6 +22,7 @@ import org.jetbrains.lincheck.descriptors.MethodSignature
 import org.jetbrains.lincheck.descriptors.VariableDescriptor
 import org.jetbrains.lincheck.descriptors.Types
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicInteger
 
 const val UNKNOWN_CODE_LOCATION_ID = -1
 // This method type corresponds to the following method descriptor '(V)V'
@@ -40,8 +41,12 @@ class TraceContext {
     val methodPool = DescriptorPool<MethodDescriptor>()
     val fieldPool = DescriptorPool<FieldDescriptor>()
     val variablePool = DescriptorPool<VariableDescriptor>()
+    private val anonymousThreadCounter = AtomicInteger(0)
 
-    fun setThreadName(id: Int, name: String) { threadNames[id] = name }
+
+    fun setThreadName(id: Int, name: String?) {
+        threadNames[id] = name ?: "anonymous-thread-${anonymousThreadCounter.getAndIncrement()}"
+    }
 
     fun getThreadName(id: Int): String = threadNames[id] ?: ""
 
--- a/common/src/main/org/jetbrains/lincheck/util/Libs.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/Libs.kt
@@ -392,3 +392,38 @@ fun isMethodHandleRelatedClass(className: String): Boolean =
  */
 fun isIgnoredMethodHandleMethod(className: String, methodName: String): Boolean =
     isMethodHandleRelatedClass(className) && !methodName.contains("invoke")
+
+
+fun isSpringRelatedClass(className: String) = when {
+    className.startsWith("sun.rmi.") -> true
+    className.startsWith("javax.") -> true
+
+    className.startsWith("org.springframework.data.jpa.") -> true
+    className.startsWith("org.springframework.boot.sql.") -> true
+    className.startsWith("org.springframework.boot.autoconfigure.") -> true
+    className.startsWith("org.springframework.cache.") -> true
+    className.startsWith("org.springframework.boot.logging.") -> true
+    className.startsWith("org.springframework.boot.cache.") -> true
+    className.startsWith("org.springframework.boot.jdbc.") -> true
+
+    className.startsWith("org.apache.logging.") -> true
+    className.startsWith("org.apache.logging.log4j.") -> true
+    className.startsWith("org.apache.juli.logging.") -> true
+    className.startsWith("org.apache.tomcat.") -> true
+    className.startsWith("org.apache.catalina.") -> true
+    className.startsWith("org.apache.coyote.") -> true
+    className.startsWith("org.antlr.") -> true
+
+    className.startsWith("ch.qos.logback.") -> true
+    className.startsWith("io.micrometer.") -> true
+    className.startsWith("io.opentelemetry.") -> true
+    className.startsWith("com.google.protobuf.") -> true
+    className.startsWith("com.fasterxml.") -> true
+    className.startsWith("tools.jackson.") -> true
+
+    className.startsWith("org.hibernate.") -> true
+    className.startsWith("org.h2.") -> true
+    className.startsWith("jakarta.") -> true
+
+    else -> false
+}
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
@@ -48,7 +48,7 @@ object LincheckClassFileTransformer : ClassFileTransformer {
 
     private val statsTracker: TransformationStatisticsTracker? =
         if (collectTransformationStatistics) TransformationStatisticsTracker() else null
-    
+
     val liveDebuggerSettings = LiveDebuggerSettings()
 
     override fun transform(
@@ -67,18 +67,23 @@ object LincheckClassFileTransformer : ClassFileTransformer {
         // this can be related to the Kotlin compiler bug:
         // - https://youtrack.jetbrains.com/issue/KT-16727/
         if (internalClassName == null) return null
-        // If the class should not be transformed, return immediately.
-        if (!shouldTransform(internalClassName.toCanonicalClassName(), instrumentationMode)) {
-            return null
-        }
-        // If lazy mode is used, transform classes lazily,
-        // once they are used in the testing code.
-        if (instrumentationStrategy == InstrumentationStrategy.LAZY &&
-            // do not re-transform already instrumented classes
-            internalClassName.toCanonicalClassName() !in instrumentedClasses &&
-            // always transform eagerly instrumented classes
-            !isEagerlyInstrumentedClass(internalClassName.toCanonicalClassName())
-        ) {
+        try {
+            // If the class should not be transformed, return immediately.
+            if (!shouldTransform(internalClassName.toCanonicalClassName(), instrumentationMode)) {
+                return null
+            }
+            // If lazy mode is used, transform classes lazily,
+            // once they are used in the testing code.
+            if (instrumentationStrategy == InstrumentationStrategy.LAZY &&
+                // do not re-transform already instrumented classes
+                internalClassName.toCanonicalClassName() !in instrumentedClasses &&
+                // always transform eagerly instrumented classes
+                !isEagerlyInstrumentedClass(internalClassName.toCanonicalClassName())
+            ) {
+                return null
+            }
+        } catch (_: ClassCircularityError) {
+            // Some bootstrap class
             return null
         }
 
@@ -121,7 +126,10 @@ object LincheckClassFileTransformer : ClassFileTransformer {
         )
 
         val writer = SafeClassWriter(reader, loader, ClassWriter.COMPUTE_FRAMES)
-        val visitor = LincheckClassVisitor(writer, classInfo, instrumentationMode, profile, statsTracker, liveDebuggerSettings, LincheckInstrumentation.context)
+        val context = LincheckInstrumentation.context
+        val visitor = LincheckClassVisitor(
+            writer, classInfo, instrumentationMode, profile, statsTracker, liveDebuggerSettings, context
+        )
 
         try {
             val timeNano = measureTimeNano {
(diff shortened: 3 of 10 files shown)


zhelenskiy:
The problem is that such a prefix is added to every trace, so it is necessary to regenerate it in the tests.

image

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
```diff
@@ -40,8 +41,12 @@ class TraceContext {
     val methodPool = DescriptorPool<MethodDescriptor>()
     val fieldPool = DescriptorPool<FieldDescriptor>()
     val variablePool = DescriptorPool<VariableDescriptor>()
+    private val anonymousThreadCounter = AtomicInteger(0)
```
eupp, line 44: Why do you need a separate counter for anonymous threads?
Why not just print `"anonymous-thread-$id"`?

Also, consider changing the thread name to `"unknown-thread-$id"`.

common/src/main/org/jetbrains/lincheck/util/Libs.kt
```diff
@@ -391,4 +391,39 @@ fun isMethodHandleRelatedClass(className: String): Boolean =
...
 fun isIgnoredMethodHandleMethod(className: String, methodName: String): Boolean =
-    isMethodHandleRelatedClass(className) && !methodName.contains("invoke")
\ No newline at end of file
+    isMethodHandleRelatedClass(className) && !methodName.contains("invoke")
+
+
+fun isSpringRelatedClass(className: String) = when {
+    className.startsWith("sun.rmi.") -> true
+    className.startsWith("javax.") -> true
+
```
eupp, line 400: These 2 are not spring-related.

common/src/main/org/jetbrains/lincheck/util/Libs.kt
```diff
@@ -391,4 +391,39 @@ fun isMethodHandleRelatedClass(className: String): Boolean =
...
+    className.startsWith("org.springframework.boot.sql.") -> true
+    className.startsWith("org.springframework.boot.autoconfigure.") -> true
+    className.startsWith("org.springframework.cache.") -> true
+    className.startsWith("org.springframework.boot.logging.") -> true
+    className.startsWith("org.springframework.boot.cache.") -> true
+    className.startsWith("org.springframework.boot.jdbc.") -> true
+
+    className.startsWith("org.apache.logging.") -> true
+    className.startsWith("org.apache.logging.log4j.") -> true
+    className.startsWith("org.apache.juli.logging.") -> true
```
eupp, line 411: Let's move these to `isRecognizedLoggingLibraryClass`

common/src/main/org/jetbrains/lincheck/util/Libs.kt
```diff
@@ -391,4 +391,39 @@ fun isMethodHandleRelatedClass(className: String): Boolean =
...
+    className.startsWith("org.apache.tomcat.") -> true
+    className.startsWith("org.apache.catalina.") -> true
+    className.startsWith("org.apache.coyote.") -> true
```
eupp: Let's move these to `isRecognizedApacheLibraryClass`

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
```diff
@@ -286,9 +298,9 @@ object LincheckClassFileTransformer : ClassFileTransformer {
             return a.length == b.length
                     && a.length > 3
                     && (
-                           (a.startsWith("set") && b.startsWith("get"))
-                        || (a.startsWith("get") && b.startsWith("set"))
-                       )
+                    (a.startsWith("set") && b.startsWith("get"))
+                            || (a.startsWith("get") && b.startsWith("set"))
+                    )
```
eupp, line 303: Please rollback

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
```diff
@@ -320,7 +332,7 @@ object LincheckClassFileTransformer : ClassFileTransformer {
         allMethods.sortBy { it.third.first() }
 
         // Special case: on-line setter and getter for same name can share this line
-        for (i in 0 ..< allMethods.size - 1) {
+        for (i in 0..<allMethods.size - 1) {
```
eupp, line 335: Please rollback

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
```diff
@@ -432,7 +444,7 @@ object LincheckClassFileTransformer : ClassFileTransformer {
         isCoroutineConcurrentKtInternalClass(className)
 
     private fun readUTF(classReader: ClassReader, utfOffset: Int, utfLength: Int, buffer: ByteArray): String {
-        for (offset in 0 ..< utfLength) {
+        for (offset in 0..<utfLength) {
```
eupp, line 447: please rollback

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -767,6 +769,8 @@ private fun isRecognizedUninstrumentedClass(className: String): Boolean {
 
     if (isRecognizedUninstrumentedThirdPartyLibraryClass(className)) return true
 
+    if (isSpringRelatedClass(className)) return true
```
eupp, line 772: Let's exclude spring classes only for Trace Recorder for now.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -807,6 +811,8 @@ private fun shouldWrapInIgnoredSection(className: String, methodName: String, de
     if (isIntellijRuntimeAgentClass(className))
         return true
 
+    if (isSpringRelatedClass(className))
```
eupp, line 814: Let's exclude spring classes only for Trace Recorder for now.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/ProgramScopedTraceRecorder.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.jetbrains.lincheck.tracer.TracingSession
+import org.jetbrains.lincheck.tracer.isFileMode
+import org.jetbrains.lincheck.util.Logger
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Manages program-scoped trace recording for trace recorder mode.
+ * Recording starts at program startup and stops at shutdown via a shutdown hook.
+ */
+internal object ProgramScopedTraceRecorder {
```
eupp, line 24: Let's just port this logic to `Tracer.kt` instead of adding a new class here. 

I believe this logic is now shared and uniform for both Trace Recorder and Live Debugger, so it makes sense to move it there.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/ProgramScopedTraceRecorder.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    private val shutdownHookInstalled = AtomicBoolean(false)
+
+    fun startRecording(mode: TraceOutputMode, traceDumpFilePath: String? = null, packTrace: Boolean = true) {
+        try {
+            // Force BinaryFileDump mode for program-scoped recording.
+            // BinaryFileStream has a race condition where multiple threads can write descriptors
+            // in the wrong order due to shared ContextSavingState across threads.
+            val effectiveMode = when (mode) {
+                is TraceOutputMode.BinaryFileStream -> TraceOutputMode.BinaryFileDump()
```
eupp, line 34: Let's don't change mode silently. Moreover, binary file dump would not scale as it keeps whole trace in-memory. 

If there is a but, then please create a YT issue describing the bug and a reproducer.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/TraceRecorderAgent.kt
```diff
@@ -69,8 +79,8 @@ internal object TraceRecorderAgent {
             override fun onStreamingDisconnect() {}
         }
 
-        override val tracingEntryPointMethodVisitorProvider: TracingEntryPointMethodVisitorProvider
-            get() = ::TraceRecorderMethodTransformer
+        override val tracingEntryPointMethodVisitorProvider: TracingEntryPointMethodVisitorProvider?
+            get() = null
```
eupp, line 83: Please rollback. 

I believe, at least for backward compatibility reasons, we want to preserve the old behavior for test-like configurations in IDEA plugin. For these configurations, the trace will be started from the test function. If you switch it to static tracing from `main` right now, it will bloat the trace. We need to fix this problem separately, only after that we can migrate to always starting tracing from `main`.  

For transition period, you can add an option to javaagent arguments. Or check if `class` and `method` options were passed, and if not, start tracing from `main`.

eupp (changes requested):

common/src/main/org/jetbrains/lincheck/util/Libs.kt
zhelenskiy: Generally, Spring classes are still shown, so none of the logic makes sense.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: why?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: why?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: why?

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/ProgramScopedTraceRecorder.kt
zhelenskiy: Can you also comment [here](https://github.com/JetBrains/lincheck/pull/970#issuecomment-3980773332)?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
eupp: It's unnecessary indentation change. 
The original indentation keeps the same indentation for all disjuncts.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
eupp: IMO `0<..N` is harder to read, and it is a weird exception from other binary operators that are surrounded by spaces (e.g., you don't write `0<N`, but `0 < N`). 

I know it's recommended formatting by Kotlin styleguide, but I found it inconvenient and inconsistent.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
eupp: IMO `0<..N` is harder to read, and it is a weird exception from other binary operators that are surrounded by spaces (e.g., you don't write `0<N`, but `0 < N`). 

I know it's recommended formatting by Kotlin styleguide, but I found it inconvenient and inconsistent.

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/ProgramScopedTraceRecorder.kt
eupp: I think the answer is in this comment: https://github.com/JetBrains/lincheck/pull/970#discussion_r2884132663

Let's not start tracing for premain in case if `class` and `method` agent arguments were passed, let's keep the old behavior for this case (at least temporarily).

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
zhelenskiy: To create unique names for displaying.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: But it contradicts the official formatting guidelines.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: BTW, it can be converted to
```suggestion
        for (i in allMethods.indices) {
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassFileTransformer.kt
zhelenskiy: When you write from 1 to 10, do you write `1..10` or `1 .. 10`?
image

trace-recorder/src/main/org/jetbrains/lincheck/trace/recorder/ProgramScopedTraceRecorder.kt
zhelenskiy: So, should I exclude passing the class and method from the Idea plugin side?

common/src/main/org/jetbrains/lincheck/trace/TraceContext.kt
zhelenskiy: image
Is it ok for us?

Issue #508, opened 2025-02-07
Fix `KotlinStdlibTransformationTest` on CI for java 20 & 21


dmitrii-artuhov:
The problem occurs with the modifications done in PR #469, which adds support for `System.arraycopy` tracking for static memory snaspshot. The test `KotlinStdlibTransformationTest` causes time limit on CI for java 20 & 21 tests on Linux PCs, when on older versions it finishes in under 5 seconds.

The reason is unknown and locally I was not able to reproduce it. However, this should be handled somewhere in the future

dmitrii-artuhov:
The issue turned out to be not with this specific test, but rather with all tests in `.transformation` package. *Sometimes* tests from there fail because of `unable to transform class ...`. I am investigating the reason

eupp:
Migrated to YT

PR #972, opened 2026-03-02
Show both real and target caller for reflection-like calls.

zhelenskiy:
`methodHandle.invoke(x, y, z)` becomes `f.x(x, y, z) [by methodHandle.invoke(x, y, z)]`


zhelenskiy:
The tests don't pass due to https://github.com/JetBrains/lincheck/issues/508 despite my attempts to fix it. Do you want me to mute the test?

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -266,6 +269,12 @@ class TRMethodCallTracePoint(
     val parameters: List<TRValue?>,
     val flags: Short = 0,
     parentTracePoint: TRContainerTracePoint? = null,
+    // For reflection-like calls (Method.invoke, MethodHandle.invoke, KFunction.call, etc.)
+    // these fields store information about the target method being invoked
+    val reflectionTargetClassName: String? = null,
+    val reflectionTargetMethodName: String? = null,
+    val reflectionTargetOwner: TRValue? = null,
+    val reflectionTargetParameters: List<TRValue?>? = null,
     eventId: Int = EVENT_ID_GENERATOR.getAndIncrement()
```
eupp: Can we infer `reflectionTargetOwner` and `reflectionTargetParameters` from `parameters` ?

My point is that, if we can infer them from `parameters` (which is possible I believe), instead of passing them directly, we should pass some meta-information that will allows us to take `parameters`, "parse" it and extract relevant information.

eupp, line 321: Let's also create a special class containing all necessary info, instead of passing 4 different parameters everywhere.

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -137,7 +137,17 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
...
+        if (tracePoint.reflectionTargetMethodName != null) {
+            appendReflectionTarget(tracePoint)
+            appendSpecialSymbol(" ")
+            appendSpecialSymbol("[")
+            appendKeyword("by")
+            appendSpecialSymbol(" ")
+            appendReflectionInvoker(tracePoint)
+            appendSpecialSymbol("]")
+            appendResult(tracePoint)
+        } else if (tracePoint.isConstructor()) {
```
eupp: Can we keep `isConstructor` case first?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
```diff
@@ -156,11 +166,75 @@ abstract class AbstractTRMethodCallTracePointPrinter() {
...
+        return this
+    }
+
+    protected fun TRAppendable.appendReflectionTargetOwner(tracePoint: TRMethodCallTracePoint): TRAppendable {
+        val className = tracePoint.reflectionTargetClassName ?: return this
+        val methodName = tracePoint.reflectionTargetMethodName ?: return this
+        val owner = tracePoint.reflectionTargetOwner
+        val parameters = tracePoint.reflectionTargetParameters ?: emptyList()
+
+        // Create a fake trace point with reflection target data to reuse appendOwner logic
```
eupp: Is it possible to refactor the code so fake trace point is not needed?

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
```diff
@@ -77,7 +77,7 @@ object CodeLocations {
 
     @JvmStatic
     @Synchronized
-    fun accessPath(context: TraceContext, codeLocationId: Int): AccessPath? = context.accessPath(codeLocationId)
+    fun accessPath(context: TraceContext, codeLocationId: Int, useParameterAsReceiver: Boolean = false): AccessPath? = context.accessPath(codeLocationId, useParameterAsReceiver)
```
eupp: Why do you need to add `useParameterAsReceiver` flag to code location? How this information is used?

Can we avoid adding new flag to code location?

At the very least, we should thing about its name and semantics. `useParameterAsReceiver` is not very intuitive.

common/src/main/org/jetbrains/lincheck/util/ReflectionCallDetection.kt
```diff
@@ -0,0 +1,182 @@
...
+import java.lang.invoke.MethodHandle
+import java.lang.invoke.MethodHandleInfo
+import java.lang.invoke.MethodHandles
+import java.lang.reflect.Constructor
+import java.lang.reflect.Method
+import java.lang.reflect.Modifier
+import kotlin.reflect.KFunction
+import kotlin.reflect.jvm.javaConstructor
+import kotlin.reflect.jvm.javaMethod
+
```
eupp, line 23: Would it make sense to put this code into `Reflection.kt` or you prefer a separate file? Why do you think having a separate file is better?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -283,6 +283,10 @@ object TraceRecorderDefaultTransformationProfile : TransformationProfile {
         // Instrument thread-related classes to intercept `Thread.run` beginning/end.
         if (className == "java.lang.Thread") return true
         if (className.startsWith("kotlin.concurrent.ThreadsKt")) return true
+        if (className.startsWith("kotlin.reflect.")) return false
+        // Instrumentation of `java.util.Arrays` class causes some subtle flaky bugs.
+        // See details in https://github.com/JetBrains/lincheck/issues/717.
+        if (isJavaUtilArraysClass(className)) return false
```
eupp: Is `isJavaUtilArraysClass` check required for Trace Recorder?
The `isRecognizedUninstrumentedStandardLibraryClass` check below just filters out stdlib entirely.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
```diff
@@ -475,6 +479,9 @@ object ModelCheckingDefaultTransformationProfile : TransformationProfile {
             if (className == "java.lang.Throwable") return true
             // Instrument `java.util.concurrent` classes, except atomics.
             if (className.startsWith("java.util.concurrent.") && className.contains("Atomic")) return false
+            if (isJavaUtilArraysClass(className)) return false
+            if (className.startsWith("java.lang.invoke.")) return false
+            if (className.startsWith("java.lang.reflect.")) return false
```
eupp: Could you please reoder the checks:
```kotlin
...
if (className == "java.lang.Throwable") return true
if (className.startsWith("java.lang.invoke.")) return false
if (className.startsWith("java.lang.reflect.")) return false

// Instrumentation of `java.util.Arrays` class causes some subtle flaky bugs.
// See details in https://github.com/JetBrains/lincheck/issues/717.
if (isJavaUtilArraysClass(className)) return false
if (className.startsWith("java.util.concurrent.") && className.contains("Atomic")) return false
// Instrument  other`java.util` classes.
if (className.startsWith("java.util.")) return true
```
Also would make sense to extract: 
```kotlin
className.startsWith("java.util.concurrent.") && className.contains("Atomic")
```
into a function:
```kotlin
isJavaUtilAtomicPrimitiveClass(...)
```

eupp: Also, now that you moved `isJavaUtilArraysClass` check here, do we still need it in `shouldNotInstrument(...)` function?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2353,10 +2374,26 @@ internal abstract class ManagedStrategy(
             methodInvocationId = methodInvocationId
         )
         callStackTrace.add(stackTraceElement)
-        pushShadowStackFrame(owner)
+        pushShadowStackFrame(if (reflectionData == null) owner else reflectionData.owner)
         return tracePoint
     }
 
+    private fun resolveReflectionTargetOwnerName(
```
eupp, line 2391: Could you please put this function somewhere else, e.g., in `OwnerNames.kt` ?

`ManagedStrategy` is already bloated, it is better avoid adding new stuff here as much as possible.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
```diff
@@ -119,6 +119,9 @@ sealed class TRTracePoint(
     val activeLocals: List<ActiveLocal> get() = CodeLocations.activeLocals(context, codeLocationId) ?: emptyList() // used in plugin
     val accessPath: AccessPath? get() = CodeLocations.accessPath(context, codeLocationId)
 
+    fun getAccessPath(useParameterAsReceiver: Boolean): AccessPath? =
+        CodeLocations.accessPath(context, codeLocationId, useParameterAsReceiver)
```
eupp: Why do you need this function here? Why not call `CodeLocations.accessPath` dirctly?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1830,6 +1832,11 @@ internal abstract class ManagedStrategy(
             // we are at normal deterministic method => no need to enforce the deterministic result => do nothing
         }
 
+        val reflectionData = detectReflectedCall(receiver, methodDescriptor.className, methodDescriptor.methodName, params)
+        if (reflectionData != null && reflectionData.owner == null) {
```
eupp: Could ypu please move this statement towards another `ensureClassHierarchyIsTransformed` so they are grouped together?

Also, please add a comment that this is a static method call through reflection.

eupp (changes requested):
Overall, great new feature!

Left several comments to address.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
zhelenskiy: I don't understand what you mean

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
zhelenskiy: What for?

trace/src/main/org/jetbrains/lincheck/trace/TRTracePointPrinters.kt
zhelenskiy: Only by code duplication, which seems worse.

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
zhelenskiy: The access path gets path of the owner. In the case of the target function, I want the owner to be the first parameter, not the actual owner (method handle/reflection method, etc.)

common/src/main/org/jetbrains/lincheck/util/ReflectionCallDetection.kt
zhelenskiy: I would prefer to keep it separate. This file has only one public function. All the rest are just its utils that should not be visible in other files.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/TransformationProfile.kt
zhelenskiy: It was my unsuccessful attempt to mitigate See details in https://github.com/JetBrains/lincheck/issues/717 in this PR.
See https://github.com/JetBrains/lincheck/pull/972#issuecomment-3981936059

zhelenskiy: What do you suggest me doing?

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
zhelenskiy: It is just parametrized `val accessPath` from the two lines to the top.

trace/src/main/org/jetbrains/lincheck/trace/TraceRecorderTracePoints.kt
eupp: I mean that currently you have:
```kotlin
fun detectReflectedCall(
    owner: Any?,
    className: String,
    methodName: String,
    methodParams: Array<Any?>
): ReflectedCallData?

data class ReflectedCallData(
    val owner: Any?,
    val className: String,
    val methodName: String,
    val methodParams: List<Any?>,
)
```

inside `detectReflectedCall` you "parse" what type of reflection method you have (if any) and return `ReflectedCallData`, setting `owner` and `methodParams` directly inside `detectReflectedCall`. 

There are 2 drawbacks of this approach.

1. Code duplication. 

You have:
```kotlin
private data class ReflectionTargetData(
    val className: String,
    val methodName: String,
    val owner: TRValue?,
    val parameters: List<TRValue?>
)
```

which duplicates almost entirely `ReflectedCallData`, except the types of `owner` and `parameters` are different (because in Trace Recorder you do not store objects but rather some serialized representation of it). 

There is also some code duplication in `ManagedStrategy`, where you store very similar data in `MethodCallTracePoint`, except there you store `owner: String` and `params: List<String>`. 

2. Memory overhead. 

The `owner` and `methodParams` stored in `ReflectedCallData` duplicate the original method parameters. 
And because you also store them in `TRMethodCallTracePoint` there could be significant memory overhead if there are a lot of reflective method calls in the trace. 


To tackle this problems, I propose alternative approach.  
You can change the types as follows:
```kotlin
fun detectReflectedCall(
    owner: Any?,
    className: String,
    methodName: String,
    methodParams: Array<Any?>
): ReflectedCallData?

data class ReflectedCallData(
    val className: String,
    val methodName: String,
    val reflectionCallKind: ReflectionCallKind,
)

sealed class ReflectionCallKind { // maybe enum class will be enough
   // encode different kinds of reflection API kinds here
}
```

Then you can define a helper functions:
```kotlin
fun <T> ReflectionCallKind.extractOwner(owner: T, params: List<T>): T
fun <T> ReflectionCallKind.extractParameters(owner: T, params: List<T>): List<T>
```

In essence, you can split the reflection method call parsing and data extraction. 

With this approach:

1. You can avoid copy-paste. Because now you have generic `extractOwner` and `extractParameters` that can work uniformly with different representations of method call arguments (actual objects, serialized values in Trace Recorder, strings in Lincheck, etc). 

2. You have a more compact memory footprint for `TRMethodCallTracePoint`. You only need to store there `ReflectionCallKind`. And you can extract actual owner/params in-place on demand only when you actually need them (for example, during trace printing).

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
eupp: > The access path gets path of the owner. In the case of the target function, I want the owner to be the first parameter, not the actual owner (method handle/reflection method, etc.)

Yeah, I understand what you achieve to do. 

But can we please avoid changes in `CodeLocations` singleton (and `TraceContext.accessPath`)? 

If you'd like you can define this function (the one extracting "real" owner) as an extension of `CodeLocation` class. 

Or you can just use it in-place when you need to extract the owner, see this as an example:
https://github.com/JetBrains/lincheck/blob/9bb1f109b896c810adbd9133e51dc638673ca0e9/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/OwnerNames.kt#L121

In general, the `useParameterAsReceiver: Boolean` argument name in isolation looks strange and raises many questions:
```kotlin
val ownerName = codeLocation.accessPath(useParameterAsReceiver = true) 
// why "true"? 
// what does it mean? 
// what parameter is used?  1-st one or 2-nd, 3-rd ... ?
// why parameter is used as a receiver?
```

eupp:
> The tests don't pass due to #508 despite my attempts to fix it. Do you want me to mute the test?

Could you try to triage the bug to understand what changes exactly have lead to it?

zhelenskiy:
> > The tests don't pass due to #508 despite my attempts to fix it. Do you want me to mute the test?
> 
> Could you try to triage the bug to understand what changes exactly have lead to it?

I suppose I don't have time for that until the leave.

PR #987, opened 2026-03-25
Make sure that `Descriptor.Key` does not depend on any information from `TraceContext`

dmitrii-artuhov:
This PR fixes a bug:
`Descriptor.Key` field cannot depend on any information from TraceContext, because this field should be available for accessing during context building, so its value must be accessible with incomplete trace context as well.


dmitrii-artuhov:
It seems that after the rebase, the problem was fixed in some other PR. But I think adding a comment is still reasonable

PR #991, opened 2026-03-26
Simplify? litmus test specification for EventStructureStrategy

viliamDokov:
@eupp, draft PR, so we can use LambdaRunner to run litmus tests.

It also includes some fixes to make lambda runner usable. Namely:

1. We need to return some result,from lambda runner so we can collect them to a set of outcomes. I copy pasted Lambda runner to LambdaResultRunner to be able to get that result. Name and implementation to be changed
2. When starting a new thread dynamically (not from the per-determined execution scenario), we did not track thread fork 
   events
3. Change 2 broke the `patchResultsClock` method. I was a bit confused what this method tries to restore, so I tried deleting it. All test cases still passed?

Still just a proof of concept, code needs cleanup, I just want to make writing litmus test as simple as possible.

### DIFF
--- a/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
@@ -18,7 +18,7 @@ import java.lang.invoke.CallSite;
  */
 public interface EventTracker {
 
-    void beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
+    int beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
     void beforeThreadRun(ThreadDescriptor descriptor);
     void afterThreadRunReturn(ThreadDescriptor descriptor);
     void afterThreadRunException(ThreadDescriptor descriptor, Throwable exception);
--- a/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
+++ b/bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
@@ -18,7 +18,7 @@ import java.lang.invoke.CallSite;
  */
 public interface EventTracker {
 
-    int beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
+    void beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
     void beforeThreadRun(ThreadDescriptor descriptor);
     void afterThreadRunReturn(ThreadDescriptor descriptor);
     void afterThreadRunException(ThreadDescriptor descriptor, Throwable exception);
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -27,6 +27,13 @@ internal class LambdaRunner<T>(
     val block: () -> T
 ) : AbstractActiveThreadPoolRunner() {
 
+    companion object {
+        // Ghetto secondary constructor
+        operator fun invoke(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> {
+            return LambdaRunner(timeoutMs, { block.run() })
+        }
+    }
+
     private val testName =
         runCatching { block.javaClass.simpleName }.getOrElse { "lambda" }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -22,9 +22,9 @@ import sun.nio.ch.lincheck.ThreadDescriptor
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
 
-internal class LambdaRunner(
+internal class LambdaRunner<T>(
     private val timeoutMs: Long, // for deadlock or livelock detection
-    val block: Runnable
+    val block: () -> T
 ) : AbstractActiveThreadPoolRunner() {
 
     private val testName =
@@ -53,14 +53,14 @@ internal class LambdaRunner(
         }
     }
 
-    private class LambdaWrapper(val strategy: Strategy, val block: Runnable) : Runnable {
-        var result: Result<Unit>? = null
+    private class LambdaWrapper<T>(val strategy: Strategy, val block: () -> T) : Runnable {
+        var result: Result<T>? = null
 
         override fun run() {
             result = kotlin.runCatching {
                 onStart()
                 try {
-                    block.run()
+                    block()
                 } finally {
                     onFinish()
                 }
@@ -87,14 +87,14 @@ internal class LambdaRunner(
 
     // TODO: currently we have to use `ExecutionResult`,
     //   even though in case of `LambdaRunner` the result can be simplified
-    private fun collectExecutionResults(wrapper: LambdaWrapper) =
+    private fun collectExecutionResults(wrapper: LambdaWrapper<T>) =
         emptyExecutionResult().copy(
             parallelResultsWithClock = listOf(listOf(
                 ResultWithClock(wrapper.result?.toLincheckResult() ?: NoResult, emptyClock(1))
             ))
         )
 
-    private fun RunnerTimeoutInvocationResult(wrapper: LambdaWrapper): RunnerTimeoutInvocationResult {
+    private fun RunnerTimeoutInvocationResult(wrapper: LambdaWrapper<T>): RunnerTimeoutInvocationResult {
         val threadDump = collectThreadDump()
         return RunnerTimeoutInvocationResult(threadDump, collectExecutionResults(wrapper))
     }
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -22,17 +22,11 @@ import sun.nio.ch.lincheck.ThreadDescriptor
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
 
+
 internal class LambdaRunner<T>(
     private val timeoutMs: Long, // for deadlock or livelock detection
     val block: () -> T
 ) : AbstractActiveThreadPoolRunner() {
-
-    companion object {
-        // Ghetto secondary constructor
-        operator fun invoke(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> {
-            return LambdaRunner(timeoutMs, { block.run() })
-        }
-    }
 
     private val testName =
         runCatching { block.javaClass.simpleName }.getOrElse { "lambda" }
@@ -106,3 +100,6 @@ internal class LambdaRunner<T>(
         return RunnerTimeoutInvocationResult(threadDump, collectExecutionResults(wrapper))
     }
 }
+
+internal fun LambdaRunner(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> =
+    LambdaRunner(timeoutMs, { block.run() } )
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -29,9 +29,12 @@ internal class LambdaRunner<T>(
     val block: LambdaBlock<T>
 ) : AbstractActiveThreadPoolRunner() {
 
-    // TODO: Not sure if this is the best place for these or if this is the best way to store this method
+    /**
+     * Represents a block of code that can be executed by the runner.
+     * To properly restore state between execution of the block,
+     * the block should be able to declare all the objects it captures.
+     */
     internal interface LambdaBlock<T>: () -> T {
-        // For the sake of [SnapshotTracker], we need to know the fields captured by the block of the lambda runner
         fun capturedObjects(): Iterable<Any>
     }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -24,32 +24,23 @@ import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
 
 
-internal class LambdaRunner<T>(
+internal class LambdaRunner<T> private constructor(
     private val timeoutMs: Long, // for deadlock or livelock detection
-    val block: LambdaBlock<T>
+    private val block: LambdaBlock<T>,
 ) : AbstractActiveThreadPoolRunner() {
 
     /**
      * Represents a block of code that can be executed by the runner.
-     * To properly restore state between execution of the block,
-     * the block should be able to declare all the objects it captures.
+     *
+     * Used only as a synthetic common interface for Java's [Runnable] and Kotlin's lambdas `() -> T`.
      */
-    internal interface LambdaBlock<T>: () -> T {
-        fun capturedObjects(): Iterable<Any>
-    }
+    private sealed interface LambdaBlock<T>: () -> T
 
-    internal class RunnableLambdaBlock(val runnable: Runnable) : LambdaBlock<Unit> {
-        override fun capturedObjects(): Iterable<Any> =
-            runnable.javaClass.declaredFields.mapNotNull { readFieldSafely(runnable, it).getOrNull() }
-
+    private class RunnableLambdaBlock(val runnable: Runnable) : LambdaBlock<Unit> {
         override fun invoke() = runnable.run()
     }
 
-    internal class FunctionLambdaBlock<T>(val lambda: () -> T) : LambdaBlock<T> {
-
-        override fun capturedObjects(): Iterable<Any> =
-            lambda.javaClass.declaredFields.mapNotNull { readFieldSafely(lambda, it).getOrNull() }
-
+    private class FunctionLambdaBlock<T>(val lambda: () -> T) : LambdaBlock<T> {
         override fun invoke(): T = lambda()
     }
 
@@ -57,6 +48,14 @@ internal class LambdaRunner<T>(
         runCatching { block.javaClass.simpleName }.getOrElse { "lambda" }
 
     override val executor = ActiveThreadPoolExecutor(testName, 1)
+
+    companion object {
+        fun fromRunnable(timeoutMs: Long, runnable: Runnable): LambdaRunner<Unit> =
+            LambdaRunner(timeoutMs, RunnableLambdaBlock(runnable))
+
+        fun <T> fromFunction(timeoutMs: Long, lambda: () -> T): LambdaRunner<T> =
+            LambdaRunner(timeoutMs, FunctionLambdaBlock(lambda))
+    }
 
     override fun runInvocation(): InvocationResult {
         var timeout = timeoutMs * 1_000_000
@@ -76,6 +75,19 @@ internal class LambdaRunner<T>(
             return UnexpectedExceptionInvocationResult(e.cause!!, collectExecutionResults(wrapper))
         } finally {
             resetEventTracker()
+        }
+    }
+
+    /**
+     * Returns a list of objects captured by the lambda block of this [LambdaRunner].
+     */
+    fun capturedObjects(): List<Any> {
+        val root = when (block) {
+            is RunnableLambdaBlock -> block.runnable
+            is FunctionLambdaBlock<*> -> block.lambda
+        }
+        return root.javaClass.declaredFields.mapNotNull {
+            readFieldSafely(root, it).getOrNull()
         }
     }
 
@@ -127,7 +139,7 @@ internal class LambdaRunner<T>(
 }
 
 internal fun LambdaRunner(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> =
-    LambdaRunner(timeoutMs, LambdaRunner.RunnableLambdaBlock(block))
+    LambdaRunner.fromRunnable(timeoutMs, block)
 
 internal fun<T> LambdaRunner(timeoutMs: Long, block: () -> T): LambdaRunner<T> =
-    LambdaRunner(timeoutMs, LambdaRunner.FunctionLambdaBlock(block))
+    LambdaRunner.fromFunction(timeoutMs, block)
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
@@ -18,6 +18,7 @@ import org.jetbrains.kotlinx.lincheck.strategy.Strategy
 import org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy
 import org.jetbrains.kotlinx.lincheck.util.toLincheckResult
 import org.jetbrains.kotlinx.lincheck.util.threadMapOf
+import org.jetbrains.lincheck.util.readFieldSafely
 import sun.nio.ch.lincheck.ThreadDescriptor
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
@@ -25,8 +26,29 @@ import java.util.concurrent.TimeoutException
 
 internal class LambdaRunner<T>(
     private val timeoutMs: Long, // for deadlock or livelock detection
-    val block: () -> T
+    val block: LambdaBlock<T>
 ) : AbstractActiveThreadPoolRunner() {
+
+    // TODO: Not sure if this is the best place for these or if this is the best way to store this method
+    internal interface LambdaBlock<T>: () -> T {
+        // For the sake of [SnapshotTracker], we need to know the fields captured by the block of the lambda runner
+        fun capturedObjects(): Iterable<Any>
+    }
+
+    internal class RunnableLambdaBlock(val runnable: Runnable) : LambdaBlock<Unit> {
+        override fun capturedObjects(): Iterable<Any> =
+            runnable.javaClass.declaredFields.mapNotNull { readFieldSafely(runnable, it).getOrNull() }
+
+        override fun invoke() = runnable.run()
+    }
+
+    internal class FunctionLambdaBlock<T>(val lambda: () -> T) : LambdaBlock<T> {
+
+        override fun capturedObjects(): Iterable<Any> =
+            lambda.javaClass.declaredFields.mapNotNull { readFieldSafely(lambda, it).getOrNull() }
+
+        override fun invoke(): T = lambda()
+    }
 
     private val testName =
         runCatching { block.javaClass.simpleName }.getOrElse { "lambda" }
@@ -102,4 +124,7 @@ internal class LambdaRunner<T>(
 }
 
 internal fun LambdaRunner(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> =
-    LambdaRunner(timeoutMs, { block.run() } )
+    LambdaRunner(timeoutMs, LambdaRunner.RunnableLambdaBlock(block))
+
+internal fun<T> LambdaRunner(timeoutMs: Long, block: () -> T): LambdaRunner<T> =
+    LambdaRunner(timeoutMs, LambdaRunner.FunctionLambdaBlock(block))
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
@@ -230,6 +230,7 @@ internal class EventStructure(
         // add new event to current execution
         _execution.add(event)
         // do the same for blocked requests
+
         for (blockedRequest in backtrackingPoint.blockedRequests) {
             _execution.add(blockedRequest)
             // additionally, pin blocked requests if all their predecessors are also blocked ...
@@ -299,9 +300,24 @@ internal class EventStructure(
             addUnblockingResponses(conflicts)
         }
         val danglingRequests = frontier.getDanglingRequests()
+
         val blockedRequests = danglingRequests
-            // TODO: perhaps, we should change this to the list of requests to conflicting response events?
-            .filter { it.label.isBlocking && it != event.parent && (it.label !is CoroutineSuspendLabel) }
+            .filter {
+                check(it.label.isRequest) // Dangling requests should probably be requests
+                val childEvent = execution.get(it.threadId, it.threadPosition + 1) ?: return@filter false
+                if(childEvent.parent != it)  return@filter false
+                // Maybe it would be nice to somehow keep track of conflicts as they are added in the event structure?
+                // We already compute the conflicting events when they are added.
+                // This way we do not have to compute them here every time
+                val conflicts = getConflictingEvents(
+                    it.threadId,
+                    childEvent.label,
+                    it,
+                    it.dependencies.mapNotNull { it as? AtomicThreadEvent }
+                ).filter { it != childEvent }
+                return@filter conflicts.size > 0
+            }
+
         frontier.apply {
             cut(danglingRequests)
             set(event.threadId, event.parent)
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
@@ -303,18 +303,20 @@ internal class EventStructure(
 
         val blockedRequests = danglingRequests
             .filter {
+                // TODO: (it.label !is CoroutineSuspendLabel)
                 check(it.label.isRequest) // Dangling requests should probably be requests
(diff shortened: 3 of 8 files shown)


bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -18,7 +18,7 @@
  */
 public interface EventTracker {
 
-    void beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
+    int beforeThreadStart(ThreadDescriptor descriptor, Thread startingThread, ThreadDescriptor startingThreadDescriptor);
```
viliamDokov: We need to somehow get the new thread id after it has been registered in all of the thread trackers.
Right now I added a return value to before thread start but not sure if this is the best way to approach this.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaResultRunner.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.jetbrains.kotlinx.lincheck.strategy.Strategy
+import org.jetbrains.kotlinx.lincheck.strategy.managed.ManagedStrategy
+import org.jetbrains.kotlinx.lincheck.util.toLincheckResult
+import org.jetbrains.kotlinx.lincheck.util.threadMapOf
+import sun.nio.ch.lincheck.ThreadDescriptor
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeoutException
+
+
+internal class LambdaResultRunner<T>(
```
viliamDokov: Almost 1:1 copy paste from lambda runner. Maybe we can make a generic lambda runner which factors out the common parts between a runner with a result and a runner without one.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -354,9 +304,11 @@ internal class EventStructureStrategy(
 
     private fun registerTestInstance() {
         check(!isTestInstanceRegistered)
-        val testInstance = (runner as ExecutionScenarioRunner).testInstance
+        val testInstance = (runner as? ExecutionScenarioRunner)?.testInstance
```
viliamDokov, line 355: In the case of a lambda runner, there is not test instance, so this should be a noop

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -367,6 +319,19 @@ internal class EventStructureStrategy(
         }
     }
 
+    override fun beforeThreadStart(
+        threadDescriptor: ThreadDescriptor,
+        startingThread: Thread,
+        startingThreadDescriptor: ThreadDescriptor
+    ) : ThreadId {
+        val newThreadId = super.beforeThreadStart(threadDescriptor, startingThread, startingThreadDescriptor)
+        if ( newThreadId != -1 || newThreadId != eventStructure.mainThreadId ) {
```
viliamDokov: Cleanup: hardcoded values and awkward control flow, when newThreadId is non-standard

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -698,13 +697,13 @@ internal abstract class ManagedStrategy(
         threadDescriptor: ThreadDescriptor,
         startingThread: Thread,
         startingThreadDescriptor: ThreadDescriptor
-    ): Unit = threadDescriptor.runInsideIgnoredSection {
+    ): ThreadId = threadDescriptor.runInsideIgnoredSection {
         val currentThreadId = threadScheduler.getCurrentThreadId()
         // do not track threads forked from unregistered threads
-        if (currentThreadId < 0) return
+        if (currentThreadId < 0) return -1
```
viliamDokov: Again hardcoded values and weird return values. -1 should mean that we should not touch this thread id.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
```diff
@@ -103,6 +106,65 @@ class MemoryModelTest {
         }
     }
 
+    //OG version
+    @Ignore
+    @Test
```
viliamDokov: This is how litmus tests look when doing execution scenarios

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
```diff
@@ -103,6 +106,65 @@ class MemoryModelTest {
...
+            val r1 = getValue<Pair<Int, Int>>(results.parallelResults[0][0]!!).first
+            val r2 = getValue<Pair<Int, Int>>(results.parallelResults[0][0]!!).first
+            Pair(r1,r2)
+        }
+    }
+
+    // New version
+    @Test
+    fun testRRWOpaque() {
+        val expectedOutcomes: Set<Pair<Int, Int>> = setOf(
```
viliamDokov: This is the same one, but with the new litmus test. Much shorter I think

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
```diff
@@ -103,6 +106,65 @@ class MemoryModelTest {
...
+            (0 to 0),
+            (0 to 1),
+            (1 to 0),
+//            (1 to 1), TODO: fix exploration strat to unlock this outcome
+        )
+        litmustTestv2(expectedOutcomes) {
+            val x = AtomicInteger(0)
+            var r1 = 0;
```
viliamDokov: Note, I do not like how we use shared variables to store what we read from the different threads. It may 
influence the execution of the litmus test, but in practice it should not, since these shared variables are used exclusively by only one thread. If this were rust the borrow checker would be satisfied. But again, it is a bit clunky

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -367,6 +319,19 @@ internal class EventStructureStrategy(
         }
     }
 
+    override fun beforeThreadStart(
+        threadDescriptor: ThreadDescriptor,
+        startingThread: Thread,
+        startingThreadDescriptor: ThreadDescriptor
+    ) : ThreadId {
+        val newThreadId = super.beforeThreadStart(threadDescriptor, startingThread, startingThreadDescriptor)
```
eupp: Please rollback `beforeThreadStart` API changes. 
To obtain the thread id you can simply use `ThreadScheduler::getThreadId(thread)` method.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
eupp: Please add a comment in the code why this `if` is necessary.

eupp: Also, please fix formatting:
```kotlin
if (newThreadId != -1 || newThreadId != eventStructure.mainThreadId) {
```

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -354,9 +304,11 @@ internal class EventStructureStrategy(
 
     private fun registerTestInstance() {
         check(!isTestInstanceRegistered)
-        val testInstance = (runner as ExecutionScenarioRunner).testInstance
+        val testInstance = (runner as? ExecutionScenarioRunner)?.testInstance
         //NOTE: The threadID may be messed up. See how this can be fixed.
-        (objectTracker as EventStructureObjectTracker).registerExternalObject(testInstance)
+        if(testInstance != null) {
```
eupp:
```suggestion
        if (testInstance != null) {
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -220,52 +215,7 @@ internal class EventStructureStrategy(
 
     }
 
-    // a hack to reset happens-before clocks computed by scheduler,
-    // because these clocks can be not in sync with with
-    // happens-before relation constructed by the event structure
-    // TODO: refactor this --- we need a more robust solution;
-    //   for example, we can compute happens before relation induced by
-    //   the event structure and pass it on
-    private fun patchResultsClock(execution: Execution<AtomicThreadEvent>, executionResult: ExecutionResult): ExecutionResult {
```
eupp, line 229: Why removing this?

AFAIR, it was required for the Event Structure strategy to properly interact with Linearizability checker.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -121,11 +121,6 @@ internal class EventStructureStrategy(
             if (result.isAbortedInvocation()) {
                 eventStructure.abortExploration()
             }
-            // patch clocks
-            if (result is CompletedInvocationResult) {
-                val patchedResult = patchResultsClock(eventStructure.execution, result.results)
-                result = CompletedInvocationResult(patchedResult)
-            }
```
eupp, line 128: Why removing this?

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
eupp: Please rollback, see my comment in `EventStructureStrategy`.

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaResultRunner.kt
eupp: You can just add result to `LambdaRunner` and use `Unit` in case when no result is needed.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
eupp: IMO, it's not a big problem. 

In theory, we can add some kind of `async`/`deffered` primitive that launches a lambda in another thread and returns its (deferred) result. But it is too much trouble for our use case.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -54,7 +53,10 @@ internal fun<Outcome> litmusTest(
         val strategy = createStrategy(testClass, testScenario)
         val failure = strategy.runIteration(INVOCATIONS, verifier)
         assert(failure == null) { failure.toString() }
-        Assert.assertEquals(expectedOutcomes, outcomes)
+        val missingOutcomes = expectedOutcomes - outcomes;
+        val extraOutomes = outcomes - expectedOutcomes;
+        val message = "Litmus test outcomes are different:\nMissing $missingOutcomes\nExtra: $extraOutomes\n"
```
eupp, line 60: Printing "Missing" and "Extra" outcomes explicitly is a nice idea. 
I would still leave "Expected" and "Actual" in the output though.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -109,6 +111,65 @@ internal fun getValueSuspended(result: LincheckResult): Any? = when (result) {
...
+    val INVOCATIONS = 10000
+    val options = ModelCheckingOptions().analyzeStdLib(true)
+    val testCfg = options.createTestConfigurations(block::class.java)
+    val outcomes: MutableSet<Outcome> = mutableSetOf()
+    val verifier = getResultsVerifier { result ->
+        val value = getValue<Outcome>(result.parallelResults[0][0]!!)
+        outcomes.add(value)
+        true
+    }
+    withLincheckTestContext( testCfg.instrumentationMode) {
```
eupp:
```suggestion
withLincheckTestContext(testCfg.instrumentationMode) {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -109,6 +111,65 @@ internal fun getValueSuspended(result: LincheckResult): Any? = when (result) {
...
+                if(it is ExceptionResult) {
+                    throw it.throwable
+                }
+            }
+            return verify(results)
+        }
+    }
+
+
+internal fun <T> createStrategy(testCfg: ModelCheckingCTestConfiguration, block: () -> T): EventStructureStrategy {
```
eupp: I think you don't need full `ModelCheckingCTestConfiguration` here. 
It's a bit overblown configuration object that has a lot of stuff you don't need in this context. 

Better just to pass everything you need explicitly as arguments. 

I guess here you need only:
- `timeoutMs`
- `ManagedStrategySettings`
- `inIdeaPluginReplayMode = false`.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -109,6 +111,65 @@ internal fun getValueSuspended(result: LincheckResult): Any? = when (result) {
     else -> throw IllegalArgumentException()
 }
 
+// NOTE: This is exactly the same as createVerifier from above but we do not check that the execution scenario is the same
+// perhaps we should just use this instead of create verifier? the ExecutionScenario is used for one little assertion only
```
eupp: Yes, let's please re-use `createVerifier` and adjust it if needed.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -109,6 +111,65 @@ internal fun getValueSuspended(result: LincheckResult): Any? = when (result) {
...
+
+
+internal fun <T> createStrategy(testCfg: ModelCheckingCTestConfiguration, block: () -> T): EventStructureStrategy {
+    val runner = LambdaResultRunner(timeoutMs = testCfg.timeoutMs, block)
+    return EventStructureStrategy(runner, testCfg.createSettings(), testCfg.inIdeaPluginReplayMode, LincheckInstrumentation.context).also {
+        runner.initializeStrategy(it)
+    }
+}
+
+internal inline fun<reified Outcome> litmustTestv2 (
```
eupp, line 121: Let's keep only one `litmustTest` function.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
eupp: Let's keep only one variant (the new one).

eupp (changes requested):
Please also rebase on recent `develop`.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
viliamDokov: Expected and actual there, assert equals just prints it:
Here is an example:
```
Litmus test outcomes are different:
Missing []
Extra: [(1, 1)]

Expected :[(0, 1), (1, 0)]
Actual   :[(0, 1), (1, 1), (1, 0)]
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: I had a bit of brainfart. The reason I removed it is because I did not really understand what it did and all tests passed without it anyway. Also with the lambda runner the following check failed:

```
check(actorsExecution.threadIDs.size == hbClockSize + 1)
```

Because of threadFork events we can end up with more threadIDs in the `actorsExecution` compared to `hbClockSize` (which is just `val hbClockSize = executionResult.parallelResultsWithClock.size`)

I am not really sure how the Linearizability checker would interact with the lambda runner.

For now I can just remove the check as things seem to be working without it.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
viliamDokov: I would like to do that, but right now litmusTestv2, is quite buggy because of issues with the object tracker, (which are fixed in #1001). 
Either we merge that PR first and then remove the old litmusTest function and rewrite all of the old tests that used it (mainly the tests in PrimitivesTest), or keep the the two functions for now and then and then do the steps in the first option in a separate PR once #1001 is merged

src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaResultRunner.kt
viliamDokov: Applied the changes to LambdaRunner. Had to change a few callsites but it seems ok

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
eupp: > The reason I removed it is because I did not really understand what it did 

I can explain on the next meeting by voice, too long to type on the keyboard : )

> and all tests passed without it anyway.

Because currently there are no tests that check eventstruct model checker together with linearizability checker : )

> I am not really sure how the Linearizability checker would interact with the lambda runner.

They should not interact, because Linearizability checker is only applicable to data structures testing API and thus only to `ExecutionScenarioRunner`. 

> For now I can just remove the check as things seem to be working without it.

Better to put this check under an `if` and perform it only in case of `ExecutionScenarioRunner`.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
eupp: Ok, let's fix object tracker first.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -299,9 +300,24 @@ internal class EventStructure(
             addUnblockingResponses(conflicts)
         }
         val danglingRequests = frontier.getDanglingRequests()
+
         val blockedRequests = danglingRequests
-            // TODO: perhaps, we should change this to the list of requests to conflicting response events?
-            .filter { it.label.isBlocking && it != event.parent && (it.label !is CoroutineSuspendLabel) }
+            .filter {
+                check(it.label.isRequest) // Dangling requests should probably be requests
```
viliamDokov, line 307: Notably we also introduce the change where we "proprely" track blocked reqeuest. Perhaps this commit can go in a seperate PR. But it also kind of fixes a lot of issues with thread join in the new style of litmus tests and it is a bit hard to write isolated test cases in the old style of litmus tests, so maybe they should just go together

viliamDokov (commented):
Some more changes. I am thinking that maybe it would be a good idea to keep both styles of litmus tests as this PR will baloon out of proportion otherwise. Maybe it is also good to have the old style of litmus test around to make sure that the new model checker also works with "parallel scenarios". Unless you plan on deprecating them idk...

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
```diff
@@ -22,9 +22,9 @@ import sun.nio.ch.lincheck.ThreadDescriptor
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
 
-internal class LambdaRunner(
+internal class LambdaRunner<T>(
     private val timeoutMs: Long, // for deadlock or livelock detection
-    val block: Runnable
+    val block: () -> T
```
eupp: Could you please add a secondary constructor that still takes `Runnable` as an argument?
Just for the sake of keeping behavior & API backward-compatible.

https://kotlinlang.org/docs/classes.html#secondary-constructors

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -299,9 +300,24 @@ internal class EventStructure(
...
+                val childEvent = execution.get(it.threadId, it.threadPosition + 1) ?: return@filter false
+                if(childEvent.parent != it)  return@filter false
```
eupp: I've noticed you often write `if(...)` in your code (without whitespace between `if` and brackets). 
Could you please fix it here and everywhere else ?) 

```suggestion
                if (childEvent.parent != it) return@filter false
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -299,9 +300,24 @@ internal class EventStructure(
             addUnblockingResponses(conflicts)
         }
         val danglingRequests = frontier.getDanglingRequests()
+
         val blockedRequests = danglingRequests
-            // TODO: perhaps, we should change this to the list of requests to conflicting response events?
-            .filter { it.label.isBlocking && it != event.parent && (it.label !is CoroutineSuspendLabel) }
+            .filter {
+                check(it.label.isRequest) // Dangling requests should probably be requests
+                val childEvent = execution.get(it.threadId, it.threadPosition + 1) ?: return@filter false
```
eupp: Minor nitpick: I would rename to `nextEvent`. 
`childEvent` is a bit ambiguous, since an event can have multiple children, and here you pick only one of them (very specific one --- the one from the same execution).

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -299,9 +300,24 @@ internal class EventStructure(
...
+                if(childEvent.parent != it)  return@filter false
+                // Maybe it would be nice to somehow keep track of conflicts as they are added in the event structure?
+                // We already compute the conflicting events when they are added.
+                // This way we do not have to compute them here every time
+                val conflicts = getConflictingEvents(
+                    it.threadId,
+                    childEvent.label,
+                    it,
+                    it.dependencies.mapNotNull { it as? AtomicThreadEvent }
```
eupp: Why this `mapNotNull` is needed here?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -299,9 +300,24 @@ internal class EventStructure(
             addUnblockingResponses(conflicts)
         }
         val danglingRequests = frontier.getDanglingRequests()
+
         val blockedRequests = danglingRequests
-            // TODO: perhaps, we should change this to the list of requests to conflicting response events?
-            .filter { it.label.isBlocking && it != event.parent && (it.label !is CoroutineSuspendLabel) }
```
eupp, line 304: Why `it.label.isBlocking` condition was removed from `filter` ?

eupp: Also, I would leave `(it.label !is CoroutineSuspendLabel)` under TODO comment: we might need to revert it back once we start debugging the new strategy on coroutine tests.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -137,6 +136,56 @@ internal class EventStructureStrategy(
...
+        }
+        return ExecutionResult(
+            initResults = executionResult.initResults,
+            parallelResultsWithClock = patchedParallelResults,
+            postResults = executionResult.postResults,
+            afterInitStateRepresentation = executionResult.afterInitStateRepresentation,
+            afterParallelStateRepresentation = executionResult.afterParallelStateRepresentation,
+            afterPostStateRepresentation = executionResult.afterPostStateRepresentation,
+        )
+    }
```
eupp: Just for the sake of keeping PR diff concise: could you please revert this change? (you removed `patchResultsClock` and then added it back in another place)

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/MemoryModelTest.kt
```diff
@@ -20,56 +20,36 @@
...
-            val r1 = getValue<Int>(results.parallelResults[0][0]!!)
-            val r2 = getValue<Int>(results.parallelResults[0][1]!!)
+        litmustTestv2(outcomes) {
+            val x = AtomicInteger(0)
+            val y = AtomicInteger(0)
+            var r1 = 0;
+            var r2 = 0;
+            val t1 = thread { r1 = x.get(); r2 = y.get() }
+            val t2 = thread { y.set(1) }
+            val t3 = thread { x.set(1) }
```
eupp: Please keep one statement per-line for better readability:

```suggestion
            val t1 = thread { 
                r1 = x.get() 
                r2 = y.get() 
            }
            val t2 = thread { 
                y.set(1) 
            }
            val t3 = thread { 
                x.set(1) 
            }
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -82,11 +86,20 @@ internal fun createStrategy(testClass: Class<*>, scenario: ExecutionScenario): E
...
+    else -> throw IllegalArgumentException()
+}
+
+internal fun getResultsVerifier(verify: (ExecutionResult) -> Boolean): Verifier =
+    object : Verifier {
         override fun verifyResults(scenario: ExecutionScenario?, results: ExecutionResult?): Boolean {
-            require(testScenario == scenario)
             require(results != null)
             results.parallelResults.flatten().forEach {
                 if(it is ExceptionResult) {
```
eupp:
```suggestion
                if (it is ExceptionResult) {
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
```diff
@@ -95,20 +108,52 @@ internal fun createVerifier(testScenario: ExecutionScenario?, verify: (Execution
...
+internal fun <T> createStrategy(timeoutMs: Long, settings: ManagedStrategySettings, inIdeaPluginReplayMode: Boolean = false,  block: () -> T): EventStructureStrategy {
+    val runner = LambdaRunner(timeoutMs = timeoutMs, block)
+    return EventStructureStrategy(runner, settings, inIdeaPluginReplayMode, LincheckInstrumentation.context).also {
+        runner.initializeStrategy(it)
     }
+}
 
-internal inline fun<reified T> getValue(result: LincheckResult): T =
-    (result as ValueResult).value as T
+internal inline fun<reified Outcome> litmustTestv2 (
```
eupp: Could you keep name `litmustTest` without `v2` suffix?
I think Kotlin's overload resolution will be able to pick the right one, because these 2 versions have different types of arguments.

eupp: Also, could you please write a short KDoc for each one of them (explaining what they do and what arguments they take) and add a TODO comment to keep only the lambda version in the future.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
```diff
@@ -22,9 +22,9 @@ import sun.nio.ch.lincheck.ThreadDescriptor
 import java.util.concurrent.ExecutionException
 import java.util.concurrent.TimeoutException
```
eupp, line 25: Please remove empty file `LambdaResultRunner` from the PR.

eupp (changes requested):

eupp:
> I am thinking that maybe it would be a good idea to keep both styles of litmus tests as this PR will baloon out of proportion otherwise. 

Sure, let's keep both for now, to keep the PR size manageable.

> Maybe it is also good to have the old style of litmus test around to make sure that the new model checker also works with "parallel scenarios".

I think for litmus tests it makes sense to keep only one of them. We will later enable all other existing MC tests for new strategy, among them there are a lot of "parallel scenarios" tests. We can also leave a few unit tests specifically for "parallel scenarios" + event structure strategy case.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt
viliamDokov: Does that work when we have to "specialize" the constructor to only return a type of LambdaRunner<Unit>, since Runnable is just, () -> Unit. These 2 declarations do not work:

```
    constructor(timeoutMs: Long, block: Runnable): this(timeoutMs, { block.run() } )
    constructor(timeoutMs: Long, block: Runnable): this<Unit>(timeoutMs, { block.run() } )
 ```

 I have some ideas to preserve backward compatibility, but not sure what you think.

Comment (viliamDokov) [reply lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt]: Here is one option that works:

```
    companion object {
        // Ghetto secondary constructor
        operator fun invoke(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> {
            return LambdaRunner(timeoutMs, { block.run() })
        }
    }
```

File: lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt (hunk 22 above)
Comment (viliamDokov) [reply lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt]: Just so the types work out. The type of `dependencies` is ThreadEvent, while getConfilctingEvents expects AtomicThreadEvents. To be honest I can just change the type signature of getConflicting events

File: lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt (hunk 19 above)
Comment (eupp) [reply lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/runner/LambdaRunner.kt]: We can use factory function patter here:

```kotlin
internal fun LambdaRunner(timeoutMs: Long, block: Runnable): LambdaRunner<Unit> =
    LambdaRunner(timeoutMs, { block.run() } )
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
eupp: This one is still relevant

eupp:
@viliamDokov could you please rebase on `develop` to accommodate changes from https://github.com/JetBrains/lincheck/pull/1018 ?

There is also one comment left about `litmustTestv2` function name.  

Other than that, I don't have any other comments/nitpicks.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/Utils.kt
viliamDokov: I forgot to push these changes :facepalm:

viliamDokov:
All ducks should be in a row now.

viliamDokov:
One of the tests that is failing (`TestClassFieldRestoreTest`) is related to memory snapshot.

```
       protected val memorySnapshot = SnapshotTracker().apply {
        if (runner is LambdaRunner<*>) {
            // save fields referenced by lambda for restoring by the snapshot tracker
            val lambdaBlock = runner.block
            // There is 
            lambdaBlock.javaClass.declaredFields
                ?.mapNotNull { readFieldSafely(lambdaBlock, it).getOrNull() }
                ?.forEach(::trackObjectAsRoot)
        }
    }
```
 
Probably fails now because we wrap the runnable block in lambda, which we did not do previously.
Why that breaks the test -> I do not know

viliamDokov:
Pushed a fix. Previously failing tests are passing now.  Not a big fan but I think it is good enough.

eupp (approved):

PR #997, opened 2026-04-01
Bugfix loop iterations saving

eupp:
- partly revert changes from #978

In #978 loop iteration management was moved to `addChild` function. But streaming trace collecting does not use this function, as it does not need to keep the tree in-memory. As a result of this, loop iteration was left `0`. In  combination with trace compression rule that removes 0-iteration loop nodes, it resulted in all loop nodes being removed from the trace in Trace Recorder plugin UI.


lev-serebryakov-jetbrains:
LGTM!

lev-serebryakov-jetbrains (approved):
Good middle ground.

PR #999, opened 2026-04-02
[JBRes-7900] Ensure trace context elements are always stored/loaded before trace points that reference them

dmitrii-artuhov:


common/src/main/org/jetbrains/lincheck/descriptors/CodeLocations.kt
```diff
@@ -1,85 +0,0 @@
...
-    val argumentNames: List<AccessPath?>? = null,
-    val activeLocals: List<ActiveLocal>? = null,
-) {
-    override fun equals(other: Any?): Boolean {
-        if (this === other) return true
-        if (other !is CodeLocation) return false
-
-        // TODO: argumentNames are not considered here,
-        //   because of some weird bug that occurs if we consider them;
-        //   likely, this is related to the fact that we currently do not (de)serialize them
```
dmitrii-artuhov, line 31: Is there a reproducer for this bug? Since now code locations are descriptors and their keys are code locations themselves, so I need to account for `argumentNames` in hashCode/equals.

PS. Actually `activeLocals` should also be added here. Have they had any problems discussed in this TODO?

eupp: > Is there a reproducer for this bug?

If I remember correctly, this bug lead to any integration test producing empty trace. So if integration tests are passing, it's likely fine.

common/src/main/org/jetbrains/lincheck/descriptors/CodeLocation.kt
```diff
@@ -0,0 +1,34 @@
...
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is CodeLocation) return false
+
+        return (stackTraceElement == other.stackTraceElement) && (accessPath == other.accessPath) && (argumentNames == other.argumentNames)
+    }
+
+    override fun hashCode(): Int {
+        return Objects.hash(stackTraceElement, accessPath, argumentNames)
```
eupp: Let's rewrite to avoid array allocation.

trace/src/main/org/jetbrains/lincheck/trace/SerializationBase.kt
```diff
@@ -184,6 +170,27 @@ internal sealed class TraceWriterBase(
     protected abstract val currentDataPosition: Long
     abstract val writerId: Int
 
+    private inline fun <reified T> isDescriptorSaved(id: Int) = isDescriptorSavedByWriter(T::class, id)
+    private inline fun <reified T> markDescriptorSaved(id: Int) = markDescriptorSavedByWriter(T::class, id)
```
eupp: Let's instead declare `abstract val contextState` and use it. 
All descendants will have to provide an implementation of this interface to manage their context state. 

`BufferedTraceWriter` will manage two context states --- local one (state of its block) and global one.

trace/src/main/org/jetbrains/lincheck/trace/SerializationStream.kt
```diff
@@ -35,8 +34,23 @@ internal data class IndexCell(
...
+ */
+internal data class BlockContextSnapshot(
+    val classDescriptorIds: Set<Int> = emptySet(),
+    val methodDescriptorIds: Set<Int> = emptySet(),
+    val fieldDescriptorIds: Set<Int> = emptySet(),
+    val variableDescriptorIds: Set<Int> = emptySet(),
+    val stringIds: Set<Int> = emptySet(),
+    val codeLocationIds: Set<Int> = emptySet(),
+    val accessPathIds: Set<Int> = emptySet()
+)
```
eupp: Let's use `Bitmap` instead of `Set<Int>` to avoid boxing/unboxing.

trace/src/main/org/jetbrains/lincheck/trace/SerializationStream.kt
```diff
@@ -399,7 +509,7 @@ class FileStreamingTraceCollecting(
     }
 
     private fun getDescriptorBitmap(descriptorClass: KClass<*>): AtomicBitmap? {
-        return when (descriptorClass::class) {
+        return when (descriptorClass) {
             ClassDescriptor::class -> seenClassDescriptors
```
eupp, line 433: Let's extract this bugfix into separate PR.

eupp (changes requested):

eupp (approved):

PR #1001, opened 2026-04-04
Fix EventStructure object tracker

viliamDokov:
The EventStrucutreObjectTracker was broken.
* When replaying events it did not assign ObjectIDs, consistently. In one invocation an object would have ObjectID 6, while in another it would have ObjectID 2, because the replay order changed.  
* It wrongly marked the test instance as an external object, while it should have been internal. This again messed up the ObjectIDs.
* External Object could get garbage collected between invocations, which could again break the ObjectTracker

Fixes:
* External objects are not removed between invocation from the object tracker. This way their ObjectIDs will be consistent
* Now when replaying events, if we allocate an internal object, we now give it an ObjectID from the replayed event instead of giving it a new one.
* We now hold a "hard" reference to external objects to prevent GC

### DIFF
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -1371,9 +1371,7 @@ internal abstract class ManagedStrategy(
         if (fieldDescriptor.isStatic && value !== null && !value.isImmutable) {
             LincheckInstrumentation.ensureClassHierarchyIsTransformed(value.javaClass)
         }
-        if (value !== null && objectTracker.shouldTrackObject(value)) {
-            objectTracker.registerObjectIfAbsent(value)
-        }
+        // TODO: add assertion that the value is registered in the object tracker if it needs toj
         if (collectTrace) {
             val valueRepresentation = objectTracker.getObjectRepresentation(value)
             val typeRepresentation = objectFqTypeName(value)
@@ -1402,9 +1400,7 @@ internal abstract class ManagedStrategy(
         index: Int,
         value: Any?
     ) = threadDescriptor.runInsideIgnoredSection {
-        if (value != null && objectTracker.shouldTrackObject(value)) {
-            objectTracker.registerObjectIfAbsent(value)
-        }
+        // TODO: add assertion that the value is registered in the object tracker if it needs to (but I guess that it could break some other object tracker stuff)
         if (collectTrace) {
             val eventId = getNextEventId()
             val threadId = threadScheduler.getCurrentThreadId()
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
@@ -1371,7 +1371,10 @@ internal abstract class ManagedStrategy(
         if (fieldDescriptor.isStatic && value !== null && !value.isImmutable) {
             LincheckInstrumentation.ensureClassHierarchyIsTransformed(value.javaClass)
         }
-        // TODO: add assertion that the value is registered in the object tracker if it needs toj
+        if (value !== null && objectTracker.shouldTrackObject(value)) {
+            objectTracker.registerObjectIfAbsent(value)
+        }
+
         if (collectTrace) {
             val valueRepresentation = objectTracker.getObjectRepresentation(value)
             val typeRepresentation = objectFqTypeName(value)
@@ -1400,7 +1403,9 @@ internal abstract class ManagedStrategy(
         index: Int,
         value: Any?
     ) = threadDescriptor.runInsideIgnoredSection {
-        // TODO: add assertion that the value is registered in the object tracker if it needs to (but I guess that it could break some other object tracker stuff)
+        if (value !== null && objectTracker.shouldTrackObject(value)) {
+            objectTracker.registerObjectIfAbsent(value)
+        }
         if (collectTrace) {
             val eventId = getNextEventId()
             val threadId = threadScheduler.getCurrentThreadId()
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -174,13 +174,16 @@ typealias ObjectNumber = Int
  * @property objectNumber A unique serial number for the object.
  * @property objectHashCode The identity hash code of the object.
  * @property objectDisplayNumber The number used in string representation of the object.
- * @property objectReference A weak reference to the associated object.
+ * @property objectWeakReference A weak reference to the associated object.
+ * @property objectStrongReference An optional strong reference to the associated object.
+ *      It exists solely to prevent garbage collection of the tracked object
  */
 open class ObjectEntry(
     val objectNumber: Int,
     val objectHashCode: Int,
     val objectDisplayNumber: Int,
-    val objectReference: WeakReference<Any>,
+    val objectWeakReference: WeakReference<Any>,
+    val objectStrongReference: Any? = null,
 )
 
 /**
@@ -239,7 +242,7 @@ fun ObjectTracker.getObjectDisplayNumber(obj: Any): Int =
 internal fun ObjectTracker.enumerateAllObjects(): Map<Any, Int> {
     val objectNumberMap = hashMapOf<Any, Int>()
     for (objectEntry in enumerateObjectEntries()) {
-        val obj = objectEntry.objectReference.get() ?: continue
+        val obj = objectEntry.objectWeakReference.get() ?: continue
         objectNumberMap[obj] = objectEntry.objectDisplayNumber
     }
     return objectNumberMap
@@ -447,11 +450,13 @@ open class BaseObjectTracker(
         objNumber: Int,
         objHashCode: Int,
         objDisplayNumber: Int,
-        objReference: WeakReference<Any>,
+        obj: Any,
         kind: ObjectKind = ObjectKind.NEW,
     ): ObjectEntry {
-        return ObjectEntry(objNumber, objHashCode, objDisplayNumber, objReference)
+        return ObjectEntry(objNumber, objHashCode, objDisplayNumber, createWeakReference(obj))
     }
+
+    protected fun createWeakReference(obj: Any): WeakReference<Any> = WeakReference(obj, referenceQueue)
 
     protected fun computeObjectDisplayNumber(obj: Any): Int {
         val classRepr = obj.getSpecialClassNameRepresentation()
@@ -482,7 +487,7 @@ open class BaseObjectTracker(
             objNumber = ++objectCounter,
             objHashCode = System.identityHashCode(obj),
             objDisplayNumber = computeObjectDisplayNumber(obj),
-            objReference = WeakIdentityReference(obj, referenceQueue),
+            obj = obj,
             kind = kind,
         )
         objectIndex.updateInplace(entry.objectHashCode, default = mutableListOf()) {
@@ -525,7 +530,7 @@ open class BaseObjectTracker(
     override operator fun get(obj: Any?): ObjectEntry? {
         val objHashCode = System.identityHashCode(obj)
         val entries = getEntries(objHashCode) ?: return null
-        return entries.find { it.objectReference.get() === obj }
+        return entries.find { it.objectWeakReference.get() === obj }
     }
 
     override fun enumerateObjectEntries(): Sequence<ObjectEntry> =
@@ -568,7 +573,7 @@ open class BaseObjectTracker(
      */
     private fun MutableList<ObjectEntry>.cleanup() {
         retainAll {
-            val isAlive = (it.objectReference.get() != null)
+            val isAlive = (it.objectWeakReference.get() != null)
             if (!isAlive) {
                 objectNumberIndex.remove(it.objectNumber)
             }
@@ -579,7 +584,7 @@ open class BaseObjectTracker(
     // For debugging
     override fun toString(): String {
         return "${objectIndex.entries.map { idToEntries -> 
-            "#${idToEntries.key} -> [${idToEntries.value.map { objectEntry -> objectEntry.objectReference.get() }.joinToString()}]" 
+            "#${idToEntries.key} -> [${idToEntries.value.map { objectEntry -> objectEntry.objectWeakReference.get() }.joinToString()}]" 
         }.joinToString("\n")}"
     }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -44,6 +44,15 @@ import kotlin.reflect.KClass
 interface ObjectTracker {
 
     /**
+     * Represents the kind of object being tracked.
+     *
+     * Can be either:
+     *   - [NEW] - Newly created object registered in the tracker.
+     *   - [EXTERNAL] - External object registered in the tracker.
+     */
+    enum class ObjectKind { NEW, EXTERNAL }
+
+    /**
      * Registers a thread with the given id in the object tracker.
      *
      * @param threadId the id of the thread to register.
@@ -175,6 +184,7 @@ typealias ObjectNumber = Int
  * @property objectHashCode The identity hash code of the object.
  * @property objectDisplayNumber The number used in string representation of the object.
  * @property objectWeakReference A weak reference to the associated object.
+ * @property objectKind Whether the object is an external or a new object
  * @property objectStrongReference An optional strong reference to the associated object.
  *      It exists solely to prevent garbage collection of the tracked object
  */
@@ -182,6 +192,7 @@ open class ObjectEntry(
     val objectNumber: Int,
     val objectHashCode: Int,
     val objectDisplayNumber: Int,
+    val objectKind: ObjectTracker.ObjectKind,
     val objectWeakReference: WeakReference<Any>,
     val objectStrongReference: Any? = null,
 )
@@ -433,30 +444,6 @@ open class BaseObjectTracker(
     private val perClassObjectNumeration = WeakHashMap<Class<*>, Int>()
 
     /**
-     * Represents the kind of object being tracked.
-     *
-     * Can be either:
-     *   - [NEW] - Newly created object registered in the tracker.
-     *   - [EXTERNAL] - External object registered in the tracker.
-     */
-    protected enum class ObjectKind { NEW, EXTERNAL }
-
-    protected open class BaseObjectEntry(
-        objNumber: Int,
-        objHashCode: Int,
-        objDisplayNumber: Int,
-        objWeakReference: WeakReference<Any>,
-        objStrongReference: Any?,
-        val objKind: ObjectKind
-    ) : ObjectEntry(
-        objNumber,
-        objHashCode,
-        objDisplayNumber,
-        objWeakReference,
-        objStrongReference
-    ) {}
-
-    /**
      * Method responsible for creating an instance of [ObjectEntry] for tracking an object in the system.
      * Derived classes may override this method to create their own customized instances of [ObjectEntry]
      * with additional meta-data.
@@ -466,15 +453,15 @@ open class BaseObjectTracker(
         objHashCode: Int,
         objDisplayNumber: Int,
         obj: Any,
-        kind: ObjectKind = ObjectKind.NEW,
+        kind: ObjectTracker.ObjectKind = ObjectTracker.ObjectKind.NEW,
     ): ObjectEntry {
-        return BaseObjectEntry(
+        return ObjectEntry(
             objNumber,
             objHashCode,
             objDisplayNumber,
-            objWeakReference = createWeakReference(obj),
-            objStrongReference = null,
-            objKind = kind,
+            objectKind = kind,
+            objectWeakReference = createWeakReference(obj),
+            objectStrongReference = null,
         )
     }
 
@@ -495,13 +482,13 @@ open class BaseObjectTracker(
         if (shouldOverwriteNewObjectsIdentityHashCode) {
             overwriteIdentityHashCode(obj)
         }
-        return registerObject(ObjectKind.NEW, obj)
+        return registerObject(ObjectTracker.ObjectKind.NEW, obj)
     }
 
     override fun registerExternalObject(obj: Any): ObjectEntry =
-        registerObject(ObjectKind.EXTERNAL, obj)
+        registerObject(ObjectTracker.ObjectKind.EXTERNAL, obj)
 
-    private fun registerObject(kind: ObjectKind, obj: Any): ObjectEntry {
+    private fun registerObject(kind: ObjectTracker.ObjectKind, obj: Any): ObjectEntry {
         check(!obj.isPrimitive)
         check(obj.isImmutable implies shouldTrackImmutableValues)
         cleanup()
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -575,6 +575,10 @@ open class BaseObjectTracker(
             return@retainAll isAlive
         }
     }
+    // For debugging
+    override fun toString(): String {
+        return "${objectIndex.entries.map { entry -> "#${entry.key} -> [${entry.value.map { foo -> foo.objectReference.get() }.joinToString()}]" }.joinToString("\n")}"
+    }
 
     private fun overwriteIdentityHashCode(obj: Any) {
         // Zero out the identity hash code in the object header to ensure deterministic behavior.
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -575,9 +575,12 @@ open class BaseObjectTracker(
             return@retainAll isAlive
         }
     }
+
     // For debugging
     override fun toString(): String {
-        return "${objectIndex.entries.map { entry -> "#${entry.key} -> [${entry.value.map { foo -> foo.objectReference.get() }.joinToString()}]" }.joinToString("\n")}"
+        return "${objectIndex.entries.map { idToEntries -> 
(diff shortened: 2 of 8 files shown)


src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -787,9 +787,10 @@ internal class EventStructure(
     private fun addResponseEvents(requestEvent: AtomicThreadEvent): Pair<AtomicThreadEvent?, List<AtomicThreadEvent>> {
         require(requestEvent.label.isRequest)
         tryReplayEvent(requestEvent.threadId)?.let { event ->
+            val resync = event.resynchronize(syncAlgebra)
```
eupp: Let's rename `resync` -> `resyncLabel`.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -947,11 +948,7 @@ internal class EventStructure(
     fun addObjectAllocationEvent(iThread: Int, value: OpaqueValue, objectID: ObjectNumber): AtomicThreadEvent {
         tryReplayEvent(iThread)?.let { event ->
             check(event.label is ObjectAllocationLabel)
-            // TODO: Is this a reasonable check?
-            check(event.label.objectID == objectID)
-            val id = event.label.objectID
-//            val entry = RegistryObjectEntry(id, value, event)
-//            objectRegistry.register(entry)
+            check(event.label.objectID <= objectID)
```
eupp, line 956: Why the check was relaxed from `==`  to `<=`. 
This deserves a comment at least : )

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -32,14 +32,42 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
 
     override val shouldTrackImmutableValues: Boolean = true
 
-    private class EventStructureObjectEntry(
+    // We have two types of object entires:
+    // * Internal which belong to object that are allocated during an invocation, and the test instance itself.
```
eupp:
```suggestion
   // * Internal entries which correspond to objects that are allocated during a test invocation, as well as the test instance itself.
```

eupp: Also, please apply few minor stylistic changes. 

IPrevious sentence ends with `:`, but new sentence starts from capital letter `Internal ...`. Please change `:` to `.` to use correcty punctuation rules. 

Also please add an empty comment line between paragraphs, it makes it a bit easier to read.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -356,7 +356,7 @@ internal class EventStructureStrategy(
         check(!isTestInstanceRegistered)
         val testInstance = (runner as ExecutionScenarioRunner).testInstance
         //NOTE: The threadID may be messed up. See how this can be fixed.
-        (objectTracker as EventStructureObjectTracker).registerExternalObject(testInstance)
+        (objectTracker as EventStructureObjectTracker).registerNewObject(testInstance)
```
eupp, line 361: Why test instance is registered as new object, and not as external object?
Let's at least document this in a comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -58,11 +93,16 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
         val obj = objReference.get()!!
```
eupp: Let's change `createObjectEntry` signature in the `BaseObjectTracker` to accept `objReference` as strong reference `: Any`. 

This would be more generic API (and it is fine, because at call sites we pass have a strong reference to pass to `createObjectEntry`). 
The implementation of this method itself internally can create weak reference.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -32,14 +32,42 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
+        allocation: AtomicThreadEvent,
+    ) : EventStructureObjectEntry(objNumber, objHashCode, objDisplayNumber, objReference, allocation)
+
+    private class ExternalEventStructureObjectEntry(
+        objNumber: Int,
+        objHashCode: Int,
+        objDisplayNumber: Int,
+        objReference: WeakReference<Any>,
+        @Suppress("unused")
+        val hardRef: Any, // Prevent the gc from gc
```
eupp: Let's change the  parent `ObjectEntry` API itself. 

Let's add `objectStrongReference` field there, and make it `null` by default. The client code can initialize it to non-null if it wishes for some purpose.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -32,14 +32,42 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
 
+    private class InternalEventStructureObjectEntry(
+        objNumber: Int,
+        objHashCode: Int,
+        objDisplayNumber: Int,
+        objReference: WeakReference<Any>,
+        allocation: AtomicThreadEvent,
+    ) : EventStructureObjectEntry(objNumber, objHashCode, objDisplayNumber, objReference, allocation)
+
+    private class ExternalEventStructureObjectEntry(
```
eupp: The changes between `ExternalEventStructureObjectEntry` and `InternalEventStructureObjectEntry` are minimal. Let's instead encode external/internal as boolean flag or `enum class`. Also let's consider to store this property in the base `ObjectEntry` class.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -78,6 +118,9 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
     fun getObject(id: ObjectNumber): OpaqueValue? {
         return lookupByNumber(id)?.objectReference?.get()?.opaque()
     }
+    override fun reset() {
```
eupp, line 129: Please add an empty line before this method declaration.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1373,9 +1373,7 @@ internal abstract class ManagedStrategy(
         if (fieldDescriptor.isStatic && value !== null && !value.isImmutable) {
             LincheckInstrumentation.ensureClassHierarchyIsTransformed(value.javaClass)
         }
-        if (value !== null && objectTracker.shouldTrackObject(value)) {
-            objectTracker.registerObjectIfAbsent(value)
-        }
```
eupp, line 1376: Why this was removed?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1405,9 +1403,7 @@ internal abstract class ManagedStrategy(
         index: Int,
         value: Any?
     ) = threadDescriptor.runInsideIgnoredSection {
-        if (value != null && objectTracker.shouldTrackObject(value)) {
-            objectTracker.registerObjectIfAbsent(value)
-        }
```
eupp, line 1407: Same question.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1405,9 +1403,7 @@ internal abstract class ManagedStrategy(
         index: Int,
         value: Any?
     ) = threadDescriptor.runInsideIgnoredSection {
-        if (value != null && objectTracker.shouldTrackObject(value)) {
-            objectTracker.registerObjectIfAbsent(value)
-        }
+        // TODO: add assertion that the value is registered in the object tracker if it needs to (but I guess that it could break some other object tracker stuff)
```
eupp: What about these TODOs.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -575,6 +575,10 @@ open class BaseObjectTracker(
             return@retainAll isAlive
         }
     }
+    // For debugging
```
eupp, line 584: Please add an empty line before.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -575,6 +575,10 @@ open class BaseObjectTracker(
             return@retainAll isAlive
         }
     }
+    // For debugging
+    override fun toString(): String {
+        return "${objectIndex.entries.map { entry -> "#${entry.key} -> [${entry.value.map { foo -> foo.objectReference.get() }.joinToString()}]" }.joinToString("\n")}"
```
eupp: For better compatibility, it's worth adding `toString()` to `ObjectEntry` itself, and re-use it here.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1134,4 +1134,173 @@ class PrimitivesTest {
         }
     }
 
+
+    @Test
+    fun testObjectsObjectsIDsDoNotBreakOnBackwardRevisit() {
```
eupp:
```suggestion
    fun testObjectIdsAreNotBrokenOnBackwardRevisit() {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1134,4 +1134,173 @@ class PrimitivesTest {
...
+            }
+        }
+        val outcomes: Set<Unit> = setOf(Unit)
+        litmusTest(Foo::class.java, testScenario, outcomes, UNKNOWN) { results ->
+            Unit
+        }
+    }
+
+    @Test
+    fun testDoNotGCExternalObjects() {
```
eupp:
```suggestion
    fun testExternalObjectIdsAreNotBrokenOnGarbageCollection() {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1134,4 +1134,173 @@ class PrimitivesTest {
...
+                thread {
+                    actor(Foo::two)
+                }
+            }
+        }
+        val outcomes: Set<Int> = setOf(0)
+        litmusTest(Foo::class.java, testScenario, outcomes, UNKNOWN) { results ->
+            val b1 = getValue<Int>(results.parallelResults[0][0]!!)
+            System.gc() // Kindly suggest the GC to do its thing. Should make the test fail more consistently.
+            System.gc()
```
eupp: Why calling GC twice?

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1134,4 +1134,173 @@ class PrimitivesTest {
...
+            return@litmusTest b1
+        }
+    }
+
+    @Test
+    fun testObjectIdsStableCrazy() {
```
eupp:
```suggestion
    fun testObjectIdsAreNotBrokenOnReplays() {
```

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1134,4 +1134,173 @@ class PrimitivesTest {
...
+        litmusTest(Foo::class.java, testScenario, outcomes, UNKNOWN) { results ->
+            val b1 = getValue<Int>(results.parallelResults[0][0]!!)
+            return@litmusTest b1
+        }
+    }
+
+
+
+    @Test
+    fun testStringsObjectsIDsDoNotBreakOnBackwardRevisit() {
```
eupp:
```suggestion
    fun testStringsIdsAreNotBrokenOnBackwardRevisit() {
```

eupp: I would also move this test closer to `testObjectIdsAreNotBrokenOnBackwardRevisit`, since these two tests are similar and related to each other.

eupp (changes requested):

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
viliamDokov: Actually the `entry` variable here is referring to a mutable map entry. 
I renamed is so it is a bit more clear.

src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
viliamDokov: Was messing around to see if the test was still flaky, when the code was broken. 
Maybe calling it twice would make a difference (it did not)

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
viliamDokov: Ok, but we need to somehow create a weak reference that also has the `referenceQueue`, which is a private field. We have 3 options: 

1. Make a special method to get weak references  
```
  protected fun createWeakReference(obj: Any): WeakReference<Any> = WeakReference(obj, referenceQueue)
```
2. Make `referenceQueue` protected
3. Use `super.createObjectEntry` and use the objectReference from there. (this is a bad option) 

I prefer option 1.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
viliamDokov: The todo refers to the fact that the removed code was redundant. I wanted to add a check to assert that it is indeed redundant. See comment above for more context

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
viliamDokov: This registering was redundant in the the case of EventStrucutureStrategy, as the value was always registered in `beforeReadField`/`beforeReadArrayElement` methods. I wanted to remove it and to replace it with an assertion that this object is always registered, since I was afraid that in the future we can have a bug where, while we replay an execution we forget to register some object with its original object id, which would then be registered **here** with the wrong id.

This check wanted to catch such bugs early. In fact I am very afraid of using `registerObjectIfAbsent` in the context of EventStrcutreStrategy, as it could easily break the assignment of object ids during replay.  

However I now realize, that when we are not in the context of EventStructreStrategy, the assertion that the object will always be registered does not hold. So I suppose, we cannot remove this.

Perhaps only in the case of EventStrcutureStrategy, we can substitute this with an assertion that checks if the object is already registered.

viliamDokov: Not sure what the cleanest way to do this specific check would be though.
1. We can put it in an if case but that is very ugly
```
           if (value !== null && objectTracker.shouldTrackObject(value)) {
            // In the case of eventstrcutre this object should already be registered
            if(this is EventStructureStrategy) {
                check(objectTracker.get(value) != null)
            } else {
                objectTracker.registerObjectIfAbsent(value)
            }
        }
```

2. We can override the afterReadField and keep the original code
```
    override fun afterReadField(
        codeLocation: Int,
        obj: Any?,
        fieldId: Int,
        value: Any?
    ) {
        // In the case of eventstrcutre this object should already be registered
        if (value !== null && objectTracker.shouldTrackObject(value)) {
            check(objectTracker.get(value) != null)
        }
       // The super call registers the value again, but it does not matter
        super.afterReadField(threadDescriptor, codeLocation, obj, fieldId, value)
    }
```

viliamDokov: I prefer option 2 but perhaps there is a better option? :thinking:

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: Dumb reason: Because I checked the original code and it actually registered it as a new object, not an external one.

Actual reason: Because the identity hash code of the test instance object changes between invocations. (since we create it multiple times). When we register it as an external object we assume that its object identity hash code will be the same across invocations, which is not the case.

Sidenote: Interesting that the IdentityHashCode of the threads (which we register as external objects) does not change. I would assume that after each invocation we make new threads, which means that their identity hash code changes, but it is not the case. Is this a bug waiting to happen?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
eupp: > This registering was redundant in the the case of EventStrucutureStrategy, as the value was always registered in beforeReadField/beforeReadArrayElement methods.

> However I now realize, that when we are not in the context of EventStructreStrategy, the assertion that the object will always be registered does not hold. So I suppose, we cannot remove this.

If I remember correctly, we've added this for `EventStrucutureStrategy` specifically in the first place, to ensure that the read object is registered. If I remember correctly, there was a specific test involving arrays, where we did array access to an unregistered array, and so by inserting this registration here we had an opportunity to register array first (when it was read and push into the JVM stack), before an access to its element was performed. 

But if you are saying that we don't need it anymore for `EventStrucutureStrategy`, we can drop it. 

The other model checking strategy it less dependent on property of "all objects should be registered". It uses object registration mainly to track thread-local objects and optimize switch points placement (not inject thread switches on accesses to thread-local objects).

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
eupp: Yeah, option (1) is good.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
eupp: >  Interesting that the IdentityHashCode of the threads (which we register as external objects) does not change. 

Actually, Lincheck's runners (`LambdaRunner` and `ExecutionScenarioRunner`) re-use the same "physical" threads between invocations --- these threads just leave in the pool `ActiveThreadPoolExecutor` and are not re-created on each invocation. 

That is why their identity hash codes remain the same.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -947,11 +948,12 @@ internal class EventStructure(
     fun addObjectAllocationEvent(iThread: Int, value: OpaqueValue, objectID: ObjectNumber): AtomicThreadEvent {
         tryReplayEvent(iThread)?.let { event ->
             check(event.label is ObjectAllocationLabel)
-            // TODO: Is this a reasonable check?
-            check(event.label.objectID == objectID)
-            val id = event.label.objectID
-//            val entry = RegistryObjectEntry(id, value, event)
-//            objectRegistry.register(entry)
+            //NOTE: Currently the objectID value represents the "suggested" objectID from the object tracker.
```
eupp, line 951: Minor nitpick: please use the following formatting style (2 spaces indentation, and space after `//` on each line):

```kotlin
// NOTE:  currently the objectID ...
//   The objectID is incremented each ...
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -32,13 +32,31 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
 
     override val shouldTrackImmutableValues: Boolean = true
 
+    // We have two types of object entries.
+    // * Internal entries which correspond to objects that are allocated during a test invocation, as well as the test instance itself. and the test instance itself.
```
eupp: Line is too long.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -32,13 +32,31 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
         objNumber: Int,
         objHashCode: Int,
         objDisplayNumber: Int,
-        objReference: WeakReference<Any>,
-        val allocation: AtomicThreadEvent
-    ) : ObjectEntry(objNumber, objHashCode, objDisplayNumber, objReference)
+        objWeakReference: WeakReference<Any>,
+        objStrongReference: Any?,
+        val allocation: AtomicThreadEvent,
+        val isExternal: Boolean,
```
eupp: Can we just re-use `ObjectKind` enum instead of `isExternal` flag? Can it be moved to parent `ObjectEntry` class?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -48,21 +66,50 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
             (initEvent!!.label as InitializationLabel).trackExternalObject(obj.javaClass.simpleName, objNumber)
-            return EventStructureObjectEntry(objNumber, objHashCode, objDisplayNumber, objReference, initEvent!!)
+            return EventStructureObjectEntry(
+                objNumber,
+                objHashCode,
+                objDisplayNumber,
+                objWeakReference,
+                obj,
+                initEvent!!,
+                true,
```
eupp: Please use named argument syntax `isExternal = true` here. It helps a lot when reading the code.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -48,21 +66,50 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
+            val actualObjectNumber = (allocationEvent.label as ObjectAllocationLabel).objectID
+            check(actualObjectNumber <= objNumber)
+            return EventStructureObjectEntry(
+                actualObjectNumber,
+                objHashCode,
+                objDisplayNumber,
+                objWeakReference,
+                null,
+                allocationEvent,
+                false,
```
eupp: Please use named argument syntax `isExternal = true` here. It helps a lot when reading the code.

eupp (commented):
Left just a few minor nitpicks. 
Please resolve them and then it's ready to be merged.

eupp:
Also, there are a lot of failing tests on CI ...

viliamDokov:
Tests should be fixed (did not use correct ), ran the test under `lincheck_test` locally and they pass. 
Regarding the ObjectEntry `kind`, the reason we cannot add objKind to the base object entry is because ObjectKind is defined BaseObjectTracker. The ObjectTracker interface does not concern itself with Internal/External objects.

Here are our options.

1. Add a new BaseObjectEntry class in BaseObjectTracker that extends ObjectEntry and has objKind field. 
    EventStructureObjectEntry extends from BaseObjectEntry. (names are a bit weird to be honest)
2. Do not change anything
3. Extend ObjectTracker interface with `createObjectEntry` method, and make ObjectKind public. Add objKind field to 
    ObjectEntry

I went with option 1 for now but I am not too sure.

eupp:
> Regarding the ObjectEntry kind, the reason we cannot add objKind to the base object entry is because ObjectKind is defined BaseObjectTracker. The ObjectTracker interface does not concern itself with Internal/External objects.

The `ObjectTracker` interface is implicitly aware of new/external distinctions, since it has `registerNewObject` and `registerExternalObject` methods. And `ObjectEntry` already fulfills purpose similar to the one of `BaseObjectEntry` class you introduced. 

So I would propose instead to:
1. Move `protected enum class ObjectKind { NEW, EXTERNAL }` to `ObjectTracker` interface and drop `protected` (make it public). 
2. Move `objKind` field to `ObjectEntry` and drop `BaseObjectTracker`.

eupp:
@viliamDokov I've run CI. 

Also left one last nitpick about `ObjectKind`.

viliamDokov:
Done. However now we have this annoying `ObjectTracker.ObjectKind` prefix everywhere. Not sure what the guidelines are and if we should keep it or not

eupp (approved):

PR #1002, opened 2026-04-06
Volatile field tracking

viliamDokov:
TODO: 
1. Clean some things up around `FinalFields.kt`
2. Manually testing this seems to give ok results. But we need some automated tests for this
3. Double check that I did not miss some place, which should be marked as volatile

### DIFF
--- a/common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
+++ b/common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
@@ -64,6 +64,7 @@ data class ObjectFieldAccessLocation(
                 type = Types.INT_TYPE,
                 fieldKind = FieldKind.INSTANCE,
                 isFinal = true,
+                isVolatile = false // TODO: is this really the case?
             )
         )
     }
@@ -156,10 +157,12 @@ fun Field.toAccessLocation(context: TraceContext): FieldAccessLocation {
     val fieldName = name
     val isStatic = Modifier.isStatic(modifiers)
     val isFinal = Modifier.isFinal(modifiers)
+    val isVolatile = Modifier.isVolatile(modifiers)
     val descriptor = context.createAndRegisterFieldDescriptor(className, fieldName,
         type = type.kotlin.getType(),
         fieldKind = FieldKind.fromIsStatic(isStatic),
         isFinal = isFinal,
+        isVolatile = isVolatile
     )
     return if (isStatic) {
         StaticFieldAccessLocation(descriptor)
--- a/common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
+++ b/common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
@@ -64,7 +64,7 @@ data class ObjectFieldAccessLocation(
                 type = Types.INT_TYPE,
                 fieldKind = FieldKind.INSTANCE,
                 isFinal = true,
-                isVolatile = false // TODO: is this really the case?
+                isVolatile = false
             )
         )
     }
--- a/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FieldsInfo.kt
+++ b/jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FieldsInfo.kt
@@ -10,42 +10,49 @@
 
 package org.jetbrains.lincheck.jvm.agent
 
-import org.jetbrains.lincheck.jvm.agent.FinalFields.addField
-import org.jetbrains.lincheck.jvm.agent.FinalFields.collectFieldInformation
-import org.jetbrains.lincheck.jvm.agent.FinalFields.isFinalField
+import jdk.internal.org.objectweb.asm.Opcodes.ACC_FINAL
+import jdk.internal.org.objectweb.asm.Opcodes.ACC_VOLATILE
+import org.jetbrains.lincheck.jvm.agent.FieldsInfo.addField
+import org.jetbrains.lincheck.jvm.agent.FieldsInfo.collectFieldInformation
+import org.jetbrains.lincheck.jvm.agent.FieldsInfo.isFinalField
 import java.util.concurrent.ConcurrentHashMap
 import org.objectweb.asm.*
 
 /**
- * [FinalFields] object is used to track final fields across different classes.
+ * [FieldsInfo] object is used to track final/volatile fields across different classes.
  * It is used only during byte-code transformation to get information about fields
  * and decide should we track reads of a field or not.
  *
  * During transformation [addField] and [addMutableField] methods are called when
- * we meet a field declaration. Then, when we are faced with field read instruction, [isFinalField] method
- * is called.
+ * we meet a field declaration. Then, when we are faced with field read instruction, [isFinalField] and [isVolatileField] methods
+ * are called.
  *
  * However, sometimes due to the order of class processing, we may not have information about this field yet,
  * as its class it is not loaded for now.
- * Then, we fall back to the slow-path algorithm [collectFieldInformation]: we read bytecode and analyze it using [FinalFieldsVisitor].
+ * Then, we fall back to the slow-path algorithm [collectFieldInformation]: we read bytecode and analyze it using [FieldInfoVisitor].
  * If field is found, we record this information and return the result, otherwise we scan superclass and all implemented
  * interfaces in the same way.
  */
 // TODO or create a ticket to refactor this and use FieldDescriptor and ClassNode visitor instead.
-internal object FinalFields {
+internal object FieldsInfo {
 
     /**
      * Stores a map INTERNAL_CLASS_NAME -> { FIELD_NAME -> IS FINAL } for each processed class.
      */
     private val classToFieldsMap = ConcurrentHashMap<String, HashMap<String, FieldInfo>>()
 
-     //TODO: comments
-    fun addField(internalClassName: String, fieldName: String, isFinal: Boolean, isVolatile: Boolean) {
+     /**
+     * Registers the field [fieldName] as a final field of the class [internalClassName].
+     */
+    fun addField(internalClassName: String, fieldName: String, access: Int) {
         val fields = classToFieldsMap.computeIfAbsent(internalClassName) { HashMap() }
-        fields[fieldName] = FinalFields.FieldInfo(isFinal, isVolatile)
+        fields[fieldName] = FieldsInfo.FieldInfo(access and ACC_FINAL != 0, access and ACC_VOLATILE != 0)
     }
 
-    // TODO: probably better to just expose this method
+    /**
+     * Determines the FieldInfo for the specified [fieldName]. First checks if we have already recorded the information.
+     * If not it falls back to the slow path of analyzing the bytecode and stores the results
+     */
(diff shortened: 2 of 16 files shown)


common/src/main/org/jetbrains/lincheck/descriptors/AccessLocation.kt
```diff
@@ -64,6 +64,7 @@ data class ObjectFieldAccessLocation(
                 type = Types.INT_TYPE,
                 fieldKind = FieldKind.INSTANCE,
                 isFinal = true,
+                isVolatile = false // TODO: is this really the case?
```
eupp: Yes, this is a correct interpretation, so you can remove the TODO.  

`.length` access has semantics similar to `final` fields in JMM.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FieldsInfo.kt
```diff
@@ -22,7 +21,7 @@ import org.objectweb.asm.*
  * It is used only during byte-code transformation to get information about fields
  * and decide should we track reads of a field or not.
  *
- * During transformation [addFinalField] and [addMutableField] methods are called when
+ * During transformation [addField] and [addMutableField] methods are called when
```
eupp, line 26: `addMutableField` was removed, please update the documentation accordingly

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FinalFields.kt
```diff
@@ -40,37 +39,37 @@ internal object FinalFields {
      */
     private val classToFieldsMap = ConcurrentHashMap<String, HashMap<String, FieldInfo>>()
 
-    /**
-     * Registers the field [fieldName] as a final field of the class [internalClassName].
-     */
-    fun addFinalField(internalClassName: String, fieldName: String) {
+     //TODO: comments
```
eupp: don't forget to remove TODOs

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FinalFields.kt
```diff
@@ -40,37 +39,37 @@ internal object FinalFields {
...
-    fun addMutableField(internalClassName: String, fieldName: String) {
-        val fields = classToFieldsMap.computeIfAbsent(internalClassName) { HashMap() }
-        fields[fieldName] = FinalFields.FieldInfo.MUTABLE
-    }
-
-    /**
-     * Determines if this field is final or not.
-     */
-    fun isFinalField(internalClassName: String, fieldName: String): Boolean {
+    // TODO: probably better to just expose this method
```
eupp: Let's not expose it for now. 

I am planning to refactor this class at some point in the future anyway (since we currently have several places where we collect info about classes/methods/fields, and I want to unify this at some point).

eupp: By the way, let's rename `FinalFields` class to `FieldsInfo` or something like this, since now this class keeps information not only about field being `final`, but also `volatile`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckClassVisitor.kt
```diff
@@ -40,11 +40,7 @@ internal class LincheckClassVisitor(
         signature: String?,
         value: Any?
     ): FieldVisitor {
-        if (access and ACC_FINAL != 0) {
-            FinalFields.addFinalField(className, fieldName)
-        } else {
-            FinalFields.addMutableField(className, fieldName)
-        }
+        FinalFields.addField(className, fieldName, access and ACC_FINAL != 0, access and ACC_VOLATILE != 0)
```
eupp: By the way, it's better to change API here so `addField` accepts just `access` as an argument, and initilizes `isFinal` and `isVolatile` internally itself.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -534,51 +535,89 @@ private class EventStructureMemoryTracker(
...
+        location: MemoryLocation,
+        memoryOrder: MemoryOrdering,
+        value: Any?
+    ) {
+        addWriteEvent(iThread, codeLocation, location, memoryOrder, value?.opaque())
     }
 
-    override fun beforeRead(iThread: Int, codeLocation: Int, location: MemoryLocation) {
-        addReadRequest(iThread, codeLocation, location)
+    override fun beforeRead(iThread: Int, codeLocation: Int, location: MemoryLocation, memoryOrder: MemoryOrdering) {
```
eupp: Please also make `beforeRead` signature multi-line for uniformity with other methods.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -599,10 +638,11 @@ private class EventStructureMemoryTracker(
             val readLocation  = objectTracker.getArrayAccessMemoryLocation(srcArray, srcPos + i, srcType)
             val writeLocation = objectTracker.getArrayAccessMemoryLocation(dstArray, dstPos + i, dstType)
             val value = run {
-                beforeRead(iThread, codeLocation, readLocation)
+                // TODO: not sure what memory order we should have here. I assume plain?
```
eupp: Yes, let's assume plain and remove TODO.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1327,7 +1327,8 @@ internal abstract class ManagedStrategy(
                 isStatic = fieldDescriptor.isStatic,
                 isFinal = fieldDescriptor.isFinal,
             )
-            memoryTracker!!.beforeRead(threadId, codeLocation, location)
+            val memoryOrder = if (fieldDescriptor.isVolatile) MemoryOrdering.VOLATILE else MemoryOrdering.PLAIN
```
eupp, line 1330: IDEA shows some weird warning on this line, could you check please?

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1348,12 +1349,12 @@ internal abstract class ManagedStrategy(
         }
         val threadId = threadScheduler.getCurrentThreadId()
         newSwitchPoint(threadId, codeLocation)
-        // TODO: this is a mess
         if (memoryTracker != null) {
             val type = array.javaClass.kotlin.getArrayElementType()
             val location = objectTracker.getArrayAccessMemoryLocation(array, index, type)
             // TODO: Should we use threadID or thread Descriptor here?
-            memoryTracker!!.beforeRead(threadId, codeLocation, location)
+            // NOTE: For now the memory ordering is plain. Not sure how we handle AtomicIntegerArray
```
eupp: I think this method is called only for regular array accesses. Atomic array accesses will go through `onMethodCall` injection, since techinically they are method calls on `AtomicXArray` classes.

eupp: So let's remove this comment.

src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/MemoryTracker.kt
```diff
@@ -74,51 +75,53 @@ internal fun MemoryTracker.trackAtomicMethodMemoryAccess(
     argOffset += if (location is ArrayElementMemoryLocation && !isUnsafe(owner)) 1 else 0
     when (methodDescriptor.kind) {
         AtomicMethodKind.SET -> {
-            beforeWrite(iThread, codeLocation, location,
+            // TODO: add memory ordering
```
eupp: Is this TODO still relevant?

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -439,7 +440,8 @@ internal fun DataInput.readFieldDescriptor(context: TraceContext): FieldDescript
         fieldName = readUTF(),
         type = readType(),
         fieldKind = FieldKind.fromIsStatic(isStatic = readBoolean()),
-        isFinal = readBoolean()
+        isFinal = readBoolean(),
+        isVolatile = readBoolean() // TODO this is most definitely wrong
```
eupp: Why is it wrong? Looks legit.

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
```diff
@@ -439,7 +440,8 @@ internal fun DataInput.readFieldDescriptor(context: TraceContext): FieldDescript
...
     )
 }
```
eupp, line 458: By the way, every time you change serialization scheme, you need to bump `TRACE_VERSION` constant. It is used by trace serializer/desirializer to check the saved binary data matches expected schema.

eupp (changes requested):
Please also rebase to resolve conflicts. 

Regarding tests: let's test this via litmus tests. Once we have support for non-SC memory model in the model checker, we can simply add litmus tests verifying that `volatile` fields indeed behave differently compared to plain fields.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
viliamDokov: My idea does not show any warnings, but I assume it is because I made MemoryOrdering not internal, since it is now used pretty much everywhere. Can you tell me what the error is?

Maybe we need to change the APIs to not use MemoryOrdering and to keep it internal.
BTW, my IDEA is acting super weird when it comes to internal types, so there might some other weirdness that I have missed

eupp: Nevermind, I think it was some weird false positive anyway.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/FieldsInfo.kt
eupp: Reference to removed method `[addMutableField]` is still in the doc.

trace/src/main/org/jetbrains/lincheck/trace/SerializationCommon.kt
eupp: Please increment `TRACE_VERSION` in `SerializationCommon.kt`.

eupp (commented):
Hi @viliamDokov !

Looks good, please fix 2 remaining minor comments, and then it is ready to be merged. 

I'll also run CI.

eupp (approved):

PR #1004, opened 2026-04-07
Await loop paths instrumentation and runtime detection

acturcu:
@eupp 

This PR contains the instrumentation for await loops. I will rebase the branch after we merge #921


eupp:
@acturcu sorry, I posted my comments intended for this PR into https://github.com/JetBrains/lincheck/pull/1005 by mistake ...

eupp (changes requested):
See related comments in https://github.com/JetBrains/lincheck/pull/1005

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -251,4 +297,304 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+                visitedState[target] = visitedState[target] or bfsBit
+                queue.add(BfsEntry(target, newHasRead))
+            }
+        }
+    }
+
+    return cleanBackEdges
+}
+
+private fun BasicBlockControlFlowGraph.headerGuard(
```
eupp: Please add a short comment explaining what this function checks.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -251,4 +297,304 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+    queue.add(BfsEntry(header, initialHasRead))
+    visitedState[header] = if (initialHasRead) VISITED_WITH_READ else VISITED_NO_READ
+
+    val cleanBackEdges = mutableSetOf<BasicBlockIndex>()
+
+    while (queue.isNotEmpty()) {
+        val (currentBlock, pathHasRead) = queue.removeFirst()
+
+        // Add current block if it is a back-edge source reached through a path that has a shared read.
+        if (currentBlock in backEdgeSources && pathHasRead) {
```
eupp: I think there is a bug with the current algorithm for finding clean back edges. 

Currently, you mark back-edge as clean if "there exists one path with no side-effects".

Consider a case like this one:
```kotlin
while (i + j > 10) {
    if (i % 2 == 0) {
        Thread.onSpinWait()
    } else {
         j++
    }
    if (j % 2 == 0) continue
    i++
}
```

It looks like currently the `continue` back-edge will be marked as "clean", even though there exist a program execution that reaches it while also producing a side-effect `j++` (Let's add this case to the set of test cases).

It looks like you need to mark back-edge as clean only if ALL loop paths leading to it have no side effects. 

Fortunately, it should not be very hard to modify this function code to achieve this. 

In the `BfsEntry` you need to store both flags: `hasRead` and `hasSideEffect`, and you need to update `hasSideEffect` in the loop body. You also should remove `continue` statements when you classify basic block as having side-effect.
Instead, you should push a `BfsEntry` into queue only if target block changes its flags (you should store some map like `BasicBlock -> BlockClassification`). 

This procedure will basically compute a fixpoint over `BasicBlock` classification, so when the loop exits (`queue` becomes empty), you'll have a map  `BasicBlock -> BlockClassification` that for each basic block tells whether it has reads and side-effects. You then need to iterate through back-edge source blocks and check their flags to decide whether it is "clean" back-edge.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LocalVariablesAccessTransformer.kt
```diff
@@ -20,6 +20,12 @@ import org.objectweb.asm.commons.*
 import org.objectweb.asm.commons.InstructionAdapter.OBJECT_TYPE
 import sun.nio.ch.lincheck.*
 
+internal fun isLoadOpcode(opcode: Int) =
+    opcode == ILOAD || opcode == LLOAD || opcode == FLOAD || opcode == DLOAD || opcode == ALOAD
+
+internal fun isStoreOpcode(opcode: Int) =
+    opcode == ISTORE || opcode == LSTORE || opcode == FSTORE || opcode == DSTORE || opcode == ASTORE
```
eupp: Let's move these util functions to `TransformationUtils.kt`

bootstrap/src/sun/nio/ch/lincheck/EventTracker.java
```diff
@@ -68,6 +68,7 @@ public interface EventTracker {
 
     void onLoopIteration(ThreadDescriptor descriptor, int codeLocation, int loopId);
     void onIrreducibleLoopIteration(ThreadDescriptor descriptor, int codeLocation, int loopId);
+    void onAwaitLoopPathIteration(ThreadDescriptor descriptor, int codeLocation, int loopId);
```
eupp: Should be renamed to `onAwaitLoopPath` for consistency with `Injections.java`.

trace/src/main/org/jetbrains/lincheck/trace/serialization/Common.kt
```diff
@@ -62,6 +62,7 @@ internal enum class CodeLocationKind {
 
 internal val CodeLocation.kind: CodeLocationKind get() = when (this) {
     is LineCodeLocation -> CodeLocationKind.LINE
+    is LoopHeaderCodeLocation -> CodeLocationKind.ACCESS
```
eupp: Please add new `CodeLocationKind` for loop locations and bump `TRACE_VERSION`.

jvm-agent/src/test/resources/analysis/controlflow/AwaitLoopCases.java.txt
```diff
@@ -0,0 +1,290 @@
...
+        outer:
+        while (!flag) {
+            while (state == 0) {
+                if (counter == 0) {
+                    break outer;
+                }
+                Thread.onSpinWait();
+            }
+        }
+    }
```
eupp: Please add few more test cases:
1. nested loop where outer-loop has side-effect after inner loop:

```kotlin
while (!flag) {
    while (state == 0) {
        if (counter == 0) break;
    }
    counter++
}
```

2. nested loop where outer-loop has side-effect before inner loop:

```kotlin
while (!flag) {
    counter++
    while (state == 0) {
        if (counter == 0) break;
    }
}
```

3. nested loop where inner loop has side effect:

```kotlin
while (!flag) {
    while (state == 0) {
        counter++
    }
    if (counter < 10) continue;
    flag = true
}
```

4.  same, but await path is before side-effecting inner loop:

```kotlin
while (!flag) {
    if (counter < 10) continue;
    while (state == 0) {
        counter++
    }
    flag = true
}
```

eupp (changes requested):

trace/src/main/org/jetbrains/lincheck/trace/serialization/Common.kt
eupp: Please note this code has been moved from `Common.kt` to `TraceBinarySerialization.kt`.

eupp (approved):

PR #1005, opened 2026-04-07
Draft: Improved loop detector

acturcu:
Add the code for the adaptive loop detector, in which we try to classify the type of loops. Will rebase to latest develop when we merge #921. Depends on #1004


jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+        val nonPhonyIndex = insnIndexRemapping[idx]
+        if (nonPhonyIndex >= 0) {
+            result.getOrPut(nonPhonyIndex) { mutableListOf() }.add(loop.id)
+        }
+    }
+    return result
+}
+
+/**
+ * A loop is considered an await loop if there exists at least a back edge path such that:
```
eupp: Let's change the terminology a bit to clarify the semantics. 

Because in your approach, the `awaitness` is not the property of the whole loop, but rather a property of a particular path within a loop. 

So I would propose to use `await path` term and update the documentation accordingly.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+ *   - no shared writes (field/array writes), monitor operations, or side-effecting calls (except `Thread.onSpinWait`) are present on that path
+ *   - at least one shared read (field/array read) is present on that path
+ *
+ * By this definition, a loop that contains writes or calls on some paths can still be classified as an await loop,
+ * as long as it has at least one clean back-edge path that satisfies above conditions
+ *
+ * Examples:
+ * ```
+ * // Simple busy-wait — the whole loop body is read-only
+ * while (!flag) { }
```
eupp: Could you please also add an example of the code with the injected method calls?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+ *
+ * // Complex loop with CAS and writes on some paths, but a read-only spin-retry path
+ * _state.loop { state ->
+ *     val element = array[head].value
+ *     if (element == null) return@loop   // <-- read-only spin path
+ *     if (_state.compareAndSet(old, new)) { ... return result }
+ * }
```
eupp, line 366: Could you please also add an example of the code with the injected method calls?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+ * ```
+ */
+
+private val ON_SPIN_WAIT_METHOD = runCatching {
+    Thread::class.java.getDeclaredMethod("onSpinWait")
+}.getOrNull()
+private val THREAD_OWNER: String = Type.getInternalName(Thread::class.java)
+
+private val NO_SIDE_EFFECT_GET_METHODS = setOf(
```
eupp: Let's rename to `ATOMIC_SIDE_EFFECT_FREE_GET_METHODS`. 

Also, could you please add a TODO comment to consider moving this to `ConditionSafetyChecker` too.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+    "java/lang/invoke/VarHandle.getVolatile",
+    "java/lang/invoke/VarHandle.getAcquire",
+    "java/lang/invoke/VarHandle.getOpaque",
+)
+
+/**
+ * Classification object result used for await loop analysis
+ */
+private data class BlockClassification(
+    val hasSharedRead: Boolean,
```
eupp, line 400: How is the information about whether loop has shared reads used?

I mean, what is the point of computing it? How is it used in aglorithm?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+ *
+ * Returns a map between [LoopId] and he set of clean back-edge source [BasicBlockIndex] values.
+ */
+internal fun BasicBlockControlFlowGraph.computeAwaitLoops(
+    loopInfo: MethodLoopsInformation
+): Map<LoopId, Set<BasicBlockIndex>> {
+    if (!loopInfo.hasLoops()) return emptyMap()
+
+    // helpers for classification
+    fun isSharedWriteOpcode(opcode: Int): Boolean = when (opcode) {
```
eupp:
```suggestion
    fun isWriteOpcode(opcode: Int): Boolean = when (opcode) {
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+        Opcodes.PUTFIELD, Opcodes.PUTSTATIC,
+        Opcodes.IASTORE, Opcodes.LASTORE, Opcodes.FASTORE, Opcodes.DASTORE,
+        Opcodes.AASTORE, Opcodes.BASTORE, Opcodes.CASTORE, Opcodes.SASTORE,
+        Opcodes.MONITORENTER, Opcodes.MONITOREXIT -> true
+        else -> false
+    }
+
+    fun isSharedReadOpcode(opcode: Int): Boolean = when (opcode) {
```
eupp:
```suggestion
    fun isReadOpcode(opcode: Int): Boolean = when (opcode) {
```

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+    loopInfo: MethodLoopsInformation
+): Map<LoopId, Set<BasicBlockIndex>> {
+    if (!loopInfo.hasLoops()) return emptyMap()
+
+    // helpers for classification
+    fun isSharedWriteOpcode(opcode: Int): Boolean = when (opcode) {
+        Opcodes.PUTFIELD, Opcodes.PUTSTATIC,
+        Opcodes.IASTORE, Opcodes.LASTORE, Opcodes.FASTORE, Opcodes.DASTORE,
+        Opcodes.AASTORE, Opcodes.BASTORE, Opcodes.CASTORE, Opcodes.SASTORE,
+        Opcodes.MONITORENTER, Opcodes.MONITOREXIT -> true
```
eupp: Let's extract monitor-related instruction into separate helper.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+        else -> false
+    }
```
eupp: Also let's extract all these helper functions to be top-level functions in `LoopTransformer.kt`. Just to be sure no memory allocations happen for lambdas, and to not think about it.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+    if (headerClassification.hasSideEffects) return emptySet()
+
+    val localVariablesInHeader = mutableSetOf<Int>()
+    // Check for side effects on the header block and for variables loaded in the header and written in the body.
+    val headerBlock = basicBlocks.getOrNull(header)
+    if (headerBlock?.executableRange != null) {
+        for (i in headerBlock.executableRange) {
+            val insn = instructions.get(i)
+            if (insn is VarInsnNode &&
+                insn.opcode in listOf(Opcodes.ILOAD, Opcodes.LLOAD, Opcodes.FLOAD, Opcodes.DLOAD, Opcodes.ALOAD)
```
eupp: There already is `isLoadOpcode` helper function exactly for this purpose. 

Btw, `listOf` will result in a memory allocation on each call.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+    fun headerGuard(blockIndex: BasicBlockIndex): Boolean {
+        if (localVariablesInHeader.isEmpty()) return false
+        val block = basicBlocks.getOrNull(blockIndex) ?: return false
+        val range = block.executableRange ?: return false
+        for (i in range) {
+            val insn = instructions.get(i)
+            when (insn) {
+                is VarInsnNode -> {
+                    if (insn.opcode in listOf(
+                            Opcodes.ISTORE, Opcodes.LSTORE, Opcodes.FSTORE, Opcodes.DSTORE, Opcodes.ASTORE
```
eupp: There already is `isStoreOpcode` util function.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+        }
+    }
+
+    fun headerGuard(blockIndex: BasicBlockIndex): Boolean {
+        if (localVariablesInHeader.isEmpty()) return false
+        val block = basicBlocks.getOrNull(blockIndex) ?: return false
+        val range = block.executableRange ?: return false
+        for (i in range) {
+            val insn = instructions.get(i)
+            when (insn) {
```
eupp: Worth extracting the whole check into separate util function.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+        }
+        return false
+    }
+
+    // BFS visited state bitmask
+    // VISITED_NO_READ = visited without a read on the path,
+    // VISITED_WITH_READ = visited with at least one read on the path.
+    val VISITED_NO_READ = 1
+    val VISITED_WITH_READ = 2
+    val visitedState = IntArray(basicBlocks.size)
```
eupp: Same question as one in the comments above: why it is important to track whether path has shared read accesses?

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+ * _state.loop { state ->
+ *     val element = array[head].value
+ *     if (element == null) return@loop   // <-- read-only spin path
+ *     if (_state.compareAndSet(old, new)) { ... return result }
+ * }
+ * ```
+ */
+
+private val ON_SPIN_WAIT_METHOD = runCatching {
+    Thread::class.java.getDeclaredMethod("onSpinWait")
```
eupp: It's better to avoid using reflection here and isntead hardcode required method name & signature check. 

See, for instance, how `isThreadJoinCall` is implemented (`ThreadTransformer.kt`).

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -250,4 +293,283 @@ private fun BasicBlockControlFlowGraph.computeReachabilityFromOutsideLoops(
...
+}.getOrNull()
+private val THREAD_OWNER: String = Type.getInternalName(Thread::class.java)
```
eupp: Please move to `TransformationUtils.kt` (next to `THROWABLE_TYPE`), refactor to compute `THREAD_TYPE` instead and then use `.internalClassName` on usage-site.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -82,9 +89,30 @@ internal class LoopTransformer(
...
+    // Map from a non-phony instruction index (the last opcode of a clean back-edge source block)
+    // to the loopId. These are the sites where `onAwaitLoop` should be injected.
+    private val awaitInjectionLocations: Map<InstructionIndex, List<LoopId>> =
+        methodInfo.basicControlFlowGraph!!.computeAwaitInjectionLocations(insnIndexRemapping, awaitLoopBackEdgeSources)
+
     override fun beforeInsn(index: Int, opcode: Int): Unit = adapter.run {
         val nonPhonyIndex = currentNonPhonyInsnIndex
 
+        loopIdsByHeaderNonPhonyIndex[nonPhonyIndex]?.let { loopIds ->
+            val canonicalId = createAndDiscardCodeLocationId()
```
eupp: Code locations class hierarchy was refactored recently on `develop`. 
Could you please rebase the branch and update the codeloc-related changes?

Also, it would make sense to add new subclass for `CodeLocation` class after rebase, something like `LoopHeaderCodeLocation`.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -82,9 +89,30 @@ internal class LoopTransformer(
     private val opcodesReachableFromOutsideLoops: Map<InstructionIndex, Set<LoopId>> =
         methodInfo.basicControlFlowGraph!!.computeReachabilityFromOutsideLoops(insnIndexRemapping, loopInfo)
 
+    private val awaitLoopBackEdgeSources: Map<LoopId, Set<BasicBlockIndex>> =
+        methodInfo.basicControlFlowGraph!!.computeAwaitLoops(loopInfo)
+
+    private val loopIdsByHeaderNonPhonyIndex: Map<InstructionIndex, List<LoopId>> =
+        methodInfo.basicControlFlowGraph!!.computeLoopIdsByHeaderNonPhonyIndex(insnIndexRemapping, loopInfo)
+
+    private val codeLocationIdByLoopId = mutableMapOf<LoopId, Int>()
```
eupp: Please add comments for all newly introduced fields.

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/transformers/LoopTransformer.kt
```diff
@@ -124,6 +152,21 @@ internal class LoopTransformer(
...
+            for (loopId in loopIds) {
+                val isReducible = loopInfo.getLoopInfo(loopId)?.isReducible ?: false
+                if (!isReducible) continue
+
+                // STACK: <empty>
+                invokeStatic(Injections::getCurrentThreadDescriptorIfInAnalyzedCode)
+                adapter.push(codeLocationIdByLoopId.getValue(loopId))
+                adapter.push(loopId)
+                // STACK: descriptor, codeLocation, loopId
+                adapter.invokeStatic(Injections::onAwaitLoop)
```
eupp: Since `onAwaitLoop` is injected at loop back-edge (not on loop header enter), it would make sense to rename it to clarify the semantics. 

Few options to consider:
- `onAwaitLoopPath`
- `onBlockedLoopPath`
- `onSpinLoopPath`
- `onSpinDetected`

jvm-agent/src/main/org/jetbrains/lincheck/jvm/agent/LincheckMethodVisitor.kt
```diff
@@ -109,6 +109,13 @@ internal open class LincheckMethodVisitor(
         lineNumber = line
         super.visitLineNumber(line, start)
     }
+
+    protected fun createAndDiscardCodeLocationId(
+        accessPath: AccessPath? = null,
+        argumentNames: List<AccessPath?>? = null,
+    ): Int = loadNewCodeLocationId(accessPath, argumentNames).also {
+        adapter.pop()
+    }
```
eupp: Please see the comment about `CodeLocation` refactoring. 
I think after rebase this helper function will not be needed anymore.

jvm-agent/src/test/resources/analysis/controlflow/golden/await_loops/awaitFieldRead.txt
```diff
@@ -0,0 +1,16 @@
...
+  NORMAL EXITS:
+    B0 -> B2 : JUMP(opcode=IFNE)
+  EXCEPTION EXIT HANDLERS:
+    NONE
+  EXCLUSIVE EXITS:
+    B2
+
+
+=== Await Loop Detection ===
+Await loop 0: header=0, body=[0, 1], cleanBackEdgeSources=[1]
```
eupp: Could you please patch the printing code so its output is consistent wrt. to notation used in loop printer?

For instance, here I would expect the following text:
```text
Await loop 1: header=B0, body=[B0, B1], cleanBackEdgeSources=[B1]
```

jvm-agent/src/test/resources/analysis/controlflow/AwaitLoopCases.java.txt
```diff
@@ -0,0 +1,217 @@
...
+    // --- Case 3: Non-await loop with a method call ---
+    // Should NOT be detected as an await loop (method call is a side-effect).
+    public void nonAwaitWithMethodCall() {
+        while (!flag) {
+            doSomething();
+        }
+    }
+
+    private void doSomething() {
+        // some side-effecting call
```
eupp: Let's add some actual side-effect here. 

Because a natural potential improvement for loop analysis would be to try to analyze called method bodies for side effects. In hypothetical future when such an optimization will be implemented, this test will become red.

jvm-agent/src/test/resources/analysis/controlflow/AwaitLoopCases.java.txt
```diff
@@ -0,0 +1,217 @@
...
+            }
+        }
+    }
+
+    // --- Case 5: Await loop with array read only ---
+    // Should be detected as an await loop.
+    public int awaitArrayRead() {
+        int sum = 0;
+        while (sum < 100) {
+            sum = array[0];
```
eupp, line 71: Please add a variant of this test case where different array element is accesses on each iteration:
```text
sum = array[i++];
```

jvm-agent/src/test/resources/analysis/controlflow/AwaitLoopCases.java.txt
```diff
@@ -0,0 +1,217 @@
...
+                // clean spin path
+                continue;
+            }
+            if (s == 1) {
+                counter = s; // dirty spin path
+                continue;
+            }
+            return;
+        }
+    }
```
eupp, line 219: Please also add few cases with nested loops (including nested loops + break/continue from inner/outer loop).

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/LoopDetector.kt
```diff
@@ -58,61 +59,153 @@ interface LoopDetector {
...
+
+    // Called when a CAS operation completes
+    fun onCasResult(threadId: Int, codeLocation: Int, success: Boolean) {}
+
+    /**
+     * Called by managed strategy after a thread switch from a loop.
+     * Records which thread was switched to and the set of threads that were available,
+     * so the detector can determine whether all alternatives have been explored before declaring STUCK
+     */
+    fun onSwitchedFromLoop(threadId: Int, loopId: Int, codeLocation: Int, enabledThreads: Set<Int>) {}
```
eupp, line 80: Can we postpone this changes (and everything related) to the next PR: https://github.com/JetBrains/lincheck/pull/1005 ?

To keep this PR self-contained and centered around await loops?

I believe for await loops we only need to add one new method to handle detected await paths.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -75,12 +75,13 @@ internal abstract class ManagedStrategy(
...
-        iterationsBound = settings.loopBound,
-        recursiveCallsBound = settings.recursionBound,
-    )
+//    internal val loopDetector: LoopDetector = BoundedLoopDetector(
+//        iterationsBeforeThreadSwitch = settings.loopIterationsBeforeThreadSwitch,
+//        iterationsBound = settings.loopBound,
+//        recursiveCallsBound = settings.recursionBound,
+//    )
 
+    internal val loopDetector: LoopDetector = AdaptiveLoopDetector(recursiveCallsBound = settings.recursionBound)
```
eupp: Let's postpone `AdaptiveLoopDetector` to the next PR and keep only `BoundedLoopDetector` for now. 

Let's add new option/flag to enable/disable await loops handling. 
We need:
- API-level surface flag (see `Options` class). 
- One flag in `TransformationConfiguration` to enable/disable away-loop instrumentation. 
- One flag in `BoundedLoopDetector` to eanble/disable await loops handling at runtime. 
- Glue everything together.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2249,6 +2313,14 @@ internal abstract class ManagedStrategy(
...
             }
         }
+        // Check if the method is declared in a different class (inheritance),
+        // and if yes, check the section for the declared class as well.
+        if (ownerName != className) {
+            val declaredSection = analysisProfile.getAnalysisSectionFor(className, methodName)
+            if (declaredSection > section) {
+                return declaredSection
+            }
+        }
```
eupp: This looks like an unrelated problem fix? 
What was its purpose? 
Can you extract this fix into separate small PR?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2205,6 +2199,76 @@ internal abstract class ManagedStrategy(
...
+                    actorId = currentActorId[threadId]!!,
+                    codeLocation = codeLocation,
+                    loopId = loopId,
+                    iteration = iteration
+                )
+            )
+        }
+    }
+
+    private fun getLoopDecision(
```
eupp: I would consider renaming it into `processLoopDetectorDecision`, since it does not return anything.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -2091,61 +2126,8 @@ internal abstract class ManagedStrategy(
...
-                                codeLocation = codeLocation
-                            )
-                        )
-                        traceCollector?.addStateRepresentation()
-                    }
-                    failDueToLivelock()
-                }
-            }
+            collectLoopTrace(started, threadId, codeLocation, loopId)
+            getLoopDecision(decision, threadId, loopId, codeLocation)
```
eupp: This is actually helpful refactoring, thanks!

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
```diff
@@ -1448,6 +1462,11 @@ internal abstract class ManagedStrategy(
         }
         newSwitchPoint(threadId, codeLocation)
 
+        // inform loop detector about a shared write
+        val locationKey = fieldLocationKey(obj, fieldId)
+        val valueHash = value?.hashCode() ?: 0
+        loopDetector.onSharedWrite(threadId, codeLocation, locationKey, valueHash)
```
eupp, line 1472: Can we postpone these changes to the next PR? (same for other new loop detector calls added to `ManagedStrategy`)

eupp (changes requested):

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedStrategy.kt
acturcu: I'll remove it completely. At some point, it was a fix for how I was handling the await instrumentation, but now it's obsolete.

PR #1010, opened 2026-04-23
Fix boxed primitive type `ValueID` representation

viliamDokov:
This is a bit of an experimental fix to how we handle boxed primitive types and null values.
Previously we had a `NULL_OBJECT_NUMBER` that was equal to 0.
This was not optimal for boxed primitive types as they did not go through the object tracker, but instead their object number was derived from the value of the primitive itself. So int the case where we have a variable of type `Int?' we could not distinguish from `null` and `0`.  

The first fix here is to make the `NULL_OBJECT_NUMBER` equal to `Int.MIN_VALUE - 1`. The object number needs to be preferably negative as to not collide with the incrementing object numbers in the regular object tracker. It also needs to be some value that is outside the range of an int, in order to avoid collisions.

However there persists one issue with boxed longs. To resolve that I have a **highly experimental** suggestion, to just track some subset of Longs which includes `NULL_OBJECT_NUMBER`, i.e `<= NULL_OBJECT_NUMBER` inside the object tracker. 

Sadly the object tracker uses `identityHashCode` for to compare object, which is not what we want for Longs. We want the same long values to be identified by value and not by reference in order to have stable objects ids. So I just made a separate tracker just for the above-mentioned subset of longs.

### DIFF
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
@@ -33,7 +33,7 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
     override val shouldTrackImmutableValues: Boolean = true
 
     private class EventStructureObjectEntry(
-        objNumber: Int,
+        objNumber: Long,
         objHashCode: Int,
         objDisplayNumber: Int,
         objReference: WeakReference<Any>,
@@ -49,7 +49,7 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
     }
 
     override fun createObjectEntry(
-        objNumber: Int,
+        objNumber: Long,
         objHashCode: Int,
         objDisplayNumber: Int,
         objReference: WeakReference<Any>,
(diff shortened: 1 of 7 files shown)


lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -80,30 +113,39 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
...
 }
 
+// TODO: can we remove this method and just use getOrRegisterValueID, and make Type optional
 internal fun EventStructureObjectTracker.registerValueIfAbsent(obj: OpaqueValue?): ObjectNumber =
     when {
         obj == null -> NULL_OBJECT_NUMBER
         else -> registerObjectIfAbsent(obj.unwrap()).objectNumber
     }
 
+internal fun<T: Any> boxedValueIdToOpaqueValue(id: ValueID, conv: (Long) -> T): OpaqueValue? {
```
viliamDokov, line 123: This function is a bit ugly

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureObjectTracker.kt
```diff
@@ -43,13 +43,46 @@ internal class EventStructureObjectTracker(private val eventStructure: EventStru
 
     private var initEvent: AtomicThreadEvent? = null
 
+    private var longObjectCounter = NULL_OBJECT_NUMBER - 1
```
viliamDokov, line 46: Bad code, but the idea works

viliamDokov:
Closed because there are more issues with primitive values

PR #1013, opened 2026-04-24
Fix should apply visitor for ObjectCreationTransformerBase

viliamDokov:
The when check does not work for inherited classes.


eupp:
Wow, that one was a real blunder...
@viliamDokov huge thanks for spotting this. 

One thing though: could you please remove `ObjectCreationTransformerBase` completely and leave only `ObjectCreationTransformer`. We used to have several subclasses of this class. but now only one of them is left, so there is no sense to keep `abstract` parent class.

eupp (approved):

PR #992, opened 2026-03-26
Add release-acquire/opaque litmus tests from JAM19  for EventStrcutureStrategy

viliamDokov:
Added all of the interesting litmus test from the JAM19 paper (https://dl.acm.org/doi/10.1145/3360568).
This still uses execution scenario DSL to define the litmus tests. Probably need to merge the other PR about litmus tests
so we can reduce down the code size

PR #1015, opened 2026-04-26
Add jam 21 model

viliamDokov:
NOTE: This merge request depends on #992

We add a pretty direct and inefficient implementation of the RA section of JAM21 paper. 
This also includes a fix in the "base" eventstructure model checker, which ensures completeness.


eupp:
@viliamDokov sorry to be annoying, but would it make sense to split into 2 PRs?
- completness fix;
- JAM21

I believe for completeness problem we had a test that showcased the problem even under SeqCst model? So we don't need to depend on JAM to verify this fix.

viliamDokov:
This PR is extremely unfinished anyway.  I will retract it and I will give the completeness fix only. Initially I combined the two because I though that the completeness fix may break the original SC checker. But at that point in time I did not know how that checker worked. Now that I actually have read the code, this is not an issue

PR #1016, opened 2026-04-27
Using structural equality to support immutable objects in `ObjectTracker`

viliamDokov:
This aims to fix the issue where objectTracker cannot handle boxed primitives for fields of type any. @eupp

### DIFF
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -479,12 +479,12 @@ open class BaseObjectTracker(
     }
 
     fun compareObjects(left: Any?, right: Any?): Boolean {
-        if(left.isImmutable) return left == right
+        if (left.isImmutable && right.isImmutable) return left == right
         else return left === right
     }
 
     private fun getObjectHashCode(obj: Any?): Int {
-        if(obj.isImmutable) return obj.hashCode()
+        if (obj.isImmutable) return obj.hashCode()
         else return System.identityHashCode(obj)
     }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
@@ -475,6 +475,17 @@ open class BaseObjectTracker(
         return perClassObjectNumeration.update(objClassKey, default = offset) { it + 1 }
     }
 
+    fun compareObjects(left: Any?, right: Any?): Boolean {
+        if(left.isImmutable) return left == right
+        else return left === right
+    }
+
+    private fun getObjectHashCode(obj: Any?): Int {
+        if(obj.isImmutable) return obj.hashCode()
+        else return System.identityHashCode(obj)
+    }
+
+
     override fun registerThread(threadId: Int, thread: Thread) {
         registerObjectIfAbsent(thread)
     }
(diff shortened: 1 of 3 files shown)


lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -475,6 +475,17 @@ open class BaseObjectTracker(
         return perClassObjectNumeration.update(objClassKey, default = offset) { it + 1 }
     }
 
+    fun compareObjects(left: Any?, right: Any?): Boolean {
+        if(left.isImmutable) return left == right
```
eupp:
```suggestion
        if (left.isImmutable) return left == right
```

eupp: Also, safer option would be to check structural equality if both are immutable (as to not accidently call `equals(...)` on non-immutable type). 

In case if one is immutable, but another is not, I think we should consider them inequal for our purposes (or just fallback to reference `===` check that should also fail in such case).

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -475,6 +475,17 @@ open class BaseObjectTracker(
...
+        else return left === right
+    }
+
+    private fun getObjectHashCode(obj: Any?): Int {
+        if(obj.isImmutable) return obj.hashCode()
```
eupp:
```suggestion
        if (obj.isImmutable) return obj.hashCode()
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -538,9 +548,9 @@ open class BaseObjectTracker(
     }
 
     override operator fun get(obj: Any?): ObjectEntry? {
-        val objHashCode = System.identityHashCode(obj)
+        val objHashCode = getObjectHashCode(obj)
```
eupp, line 554: Could you please update documentation in `ObjectTracker.kt` to reflect the fact that now structural hashcode/equality is used for immutable types. It is very important change in the contract, and the documentation should mention this (currently identity hash code is mentioned in several places).

eupp (changes requested):
Left few minor nitpicks.

@viliamDokov by the way, am I right that this change should only affect new event-structure-based model checking strategy? And it should not affect the old strategy (because it does not track immutable objects anyway)?

viliamDokov:
Yes you are correct, this will now only work for EventStructreStrategy as the other object tracker does not track immutable values in the first place

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
viliamDokov: Added two sentences in the comment for the ObjectTracker interface and in the comment for ObjectEntry

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -32,10 +32,12 @@ import kotlin.reflect.KClass
  *
  * The unique object id is a 64-bit integer number, constructed from
  * the object's serial number and its identity hash code.
+ * In the case of immutable objects it hashCode is used instead.
```
eupp:
```suggestion
* the object's serial number and its identity hash code 
* (or regular `hashCode` in case of immutable objects).
```

eupp (commented):
LGTM!

@viliamDokov please rebase to resolve conflicts and apply minor suggestion, then it's good to go.

eupp (approved):

PR #1017, opened 2026-04-29
Release-Acquire memory model checker

viliamDokov:
Depends on #992. I would put that as the target branch if github allowed me to.
Currently this is quite a hacky implementation, as there are some TODOs still left.
But it implements the following RA check:

image
image

### DIFF
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -75,6 +75,10 @@ interface ThreadEvent : Event {
      */
     val causalityClock: VectorClock
 
+    // TODO: maybe generalize the notion of vector clocks somehow, to avoid making a new vector clock for each
+    //   freaking relation we want to track
+    val happensBeforeClock: VectorClock
+
     /**
      * Returns n-th predecessor of the given event.
      */
@@ -303,6 +307,18 @@ abstract class AbstractThreadEvent(
         }
     }
 
+    final override val happensBeforeClock: VectorClock = run {
+        dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
+            if(this.label.isAcquire() && event.label.isRelease() ) {
+                clock + event.happensBeforeClock
+            } else {
+                clock
+            }
+        }.apply {
+            set(threadId, threadPosition)
+        }
+    }
+
     override fun validate() {
         super.validate()
         require(threadPosition == parent.calculateNextEventPosition())
@@ -438,6 +454,10 @@ val programOrder = Relation<ThreadEvent> { x, y ->
 
 val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
+}
+
+val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
+    (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
 }
 
 val causalityCovering: Covering<ThreadEvent> = Covering { it.dependencies }
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -308,7 +308,7 @@ abstract class AbstractThreadEvent(
     // TODO: In the future we need also resolve release-acquire fences
     final override val happensBeforeClock: VectorClock = run {
         dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
-            if (this.label.isAcquire() && event.label.isRelease() ) {
+            if (this.label.isAcquire() && event.label.isRelease()) {
                 clock + event.happensBeforeClock
             } else {
                 clock
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -460,11 +460,11 @@ val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
 }
 
-val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
+val releaseAcquireHappensBefore = Relation<ThreadEvent> { x, y ->
     (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
 }
 
-val happensBeforeSameLocationOrder = (happensBeforeOrder intersection sameLocation) union initRelation
+val happensBeforeSameLocationOrder = (releaseAcquireHappensBefore intersection sameLocation) union initRelation
 
 val causalityCovering: Covering<ThreadEvent> = Covering { it.dependencies }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -307,6 +307,7 @@ abstract class AbstractThreadEvent(
         }
     }
 
+    // TODO: In the future we need also resolve release-acquire fences
     final override val happensBeforeClock: VectorClock = run {
         dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
             if(this.label.isAcquire() && event.label.isRelease() ) {
@@ -452,6 +453,11 @@ val programOrder = Relation<ThreadEvent> { x, y ->
     else (x == y.predNth(y.threadPosition - x.threadPosition))
 }
 
+val initRelation = Relation<AtomicThreadEvent> { x, _ -> x.label is InitializationLabel }
+
+val sameLocation = Relation<AtomicThreadEvent> { x, y -> getLocationForSameLocationAccesses(x,y) != null }
+
+
 val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
 }
@@ -459,6 +465,8 @@ val causalityOrder = Relation<ThreadEvent> { x, y ->
 val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
 }
+
+val happensBeforeLocOrder = (happensBeforeOrder intersection sameLocation) union initRelation
 
 val causalityCovering: Covering<ThreadEvent> = Covering { it.dependencies }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -460,11 +460,11 @@ val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
 }
 
-val releaseAcquireHappensBefore = Relation<ThreadEvent> { x, y ->
+val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
 }
 
-val happensBeforeSameLocationOrder = (releaseAcquireHappensBefore intersection sameLocation) union initRelation
+val happensBeforeSameLocationOrder = (happensBeforeOrder intersection sameLocation) union initRelation
 
 val causalityCovering: Covering<ThreadEvent> = Covering { it.dependencies }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
@@ -75,8 +75,6 @@ interface ThreadEvent : Event {
      */
     val causalityClock: VectorClock
 
-    // TODO: maybe generalize the notion of vector clocks somehow, to avoid making a new vector clock for each
-    //   freaking relation we want to track
     val happensBeforeClock: VectorClock
 
     /**
@@ -310,7 +308,7 @@ abstract class AbstractThreadEvent(
     // TODO: In the future we need also resolve release-acquire fences
     final override val happensBeforeClock: VectorClock = run {
         dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
-            if(this.label.isAcquire() && event.label.isRelease() ) {
+            if (this.label.isAcquire() && event.label.isRelease() ) {
                 clock + event.happensBeforeClock
             } else {
                 clock
@@ -462,11 +460,11 @@ val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
 }
 
-val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
+val releaseAcquireHappensBefore = Relation<ThreadEvent> { x, y ->
     (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
 }
 
-val happensBeforeLocOrder = (happensBeforeOrder intersection sameLocation) union initRelation
+val happensBeforeSameLocationOrder = (releaseAcquireHappensBefore intersection sameLocation) union initRelation
 
 val causalityCovering: Covering<ThreadEvent> = Covering { it.dependencies }
 
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
@@ -248,13 +248,13 @@ private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAcces
                     // to detect write-write race,
                     // it is sufficient to check only against the latest write event
                     val lastWrite = getLastWrite(location)!!
-                    info.isWriteWriteRaceFree = causalityOrder(lastWrite, event)
+                    info.isWriteWriteRaceFree = happensBeforeOrder(lastWrite, event)
                 }
                 if (info.isReadWriteRaceFree) {
                     // to detect read-write race,
                     // we need to check against all the read-request events
                     info.isReadWriteRaceFree = getReadRequests(location).all { read ->
-                        causalityOrder(read, event)
+                        happensBeforeOrder(read, event)
                     }
                 }
             }
@@ -263,9 +263,9 @@ private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAcces
             // it is sufficient to check only against the latest write event
             event.label is ReadAccessLabel && info.isRaceFree -> {
                 val lastWrite = getLastWrite(location)!!
-                if (causalityOrder(lastWrite, event))
+                if (happensBeforeOrder(lastWrite, event))
                     return
-                check(causalityOrder.unordered(lastWrite, event))
+                check(happensBeforeOrder.unordered(lastWrite, event))
                 info.isReadWriteRaceFree = false
             }
 
@@ -273,7 +273,7 @@ private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAcces
             // we need to check against all the write events
             event.label is ReadAccessLabel && info.isReadWriteRaceFree -> {
                 info.isReadWriteRaceFree = getWrites(location).all { write ->
-                    causalityOrder(write, event)
+                    happensBeforeOrder(write, event)
                 }
             }
         }
--- a/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
+++ b/lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
@@ -174,16 +174,16 @@ fun AtomicMemoryAccessEventIndex.getLastWrite(location: MemoryLocation): AtomicT
     getWrites(location).lastOrNull()
 
 
-fun AtomicMemoryAccessEventIndex(happensBefore: Relation<AtomicThreadEvent>): AtomicMemoryAccessEventIndex =
-    MutableAtomicMemoryAccessEventIndex(happensBefore)
+fun AtomicMemoryAccessEventIndex(): AtomicMemoryAccessEventIndex =
+    MutableAtomicMemoryAccessEventIndex()
 
-fun MutableAtomicMemoryAccessEventIndex(happensBefore: Relation<AtomicThreadEvent>): MutableAtomicMemoryAccessEventIndex =
-    MutableAtomicMemoryAccessEventIndexImpl(happensBefore)
+fun MutableAtomicMemoryAccessEventIndex(): MutableAtomicMemoryAccessEventIndex =
+    MutableAtomicMemoryAccessEventIndexImpl()
 
 typealias AtomicMemoryAccessEventClassifier =
         EventIndexClassifier<AtomicThreadEvent, AtomicMemoryAccessCategory, MemoryLocation>
 
-private class MutableAtomicMemoryAccessEventIndexImpl(val happensBefore: Relation<AtomicThreadEvent>) : MutableAtomicMemoryAccessEventIndex {
+private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAccessEventIndex {
 
     private data class LocationInfoData(
         override var isReadWriteRaceFree: Boolean,
@@ -248,13 +248,13 @@ private class MutableAtomicMemoryAccessEventIndexImpl(val happensBefore: Relatio
                     // to detect write-write race,
                     // it is sufficient to check only against the latest write event
                     val lastWrite = getLastWrite(location)!!
-                    info.isWriteWriteRaceFree = happensBefore(lastWrite, event)
+                    info.isWriteWriteRaceFree = happensBeforeOrder(lastWrite, event)
                 }
                 if (info.isReadWriteRaceFree) {
                     // to detect read-write race,
                     // we need to check against all the read-request events
                     info.isReadWriteRaceFree = getReadRequests(location).all { read ->
-                        happensBefore(read, event)
+                        happensBeforeOrder(read, event)
                     }
                 }
             }
@@ -263,9 +263,9 @@ private class MutableAtomicMemoryAccessEventIndexImpl(val happensBefore: Relatio
             // it is sufficient to check only against the latest write event
             event.label is ReadAccessLabel && info.isRaceFree -> {
                 val lastWrite = getLastWrite(location)!!
-                if (happensBefore(lastWrite, event))
+                if (happensBeforeOrder(lastWrite, event))
                     return
-                check(happensBefore.unordered(lastWrite, event))
+                check(happensBeforeOrder.unordered(lastWrite, event))
                 info.isReadWriteRaceFree = false
             }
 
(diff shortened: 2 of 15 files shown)


lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/CoherenceChecker.kt
```diff
@@ -87,12 +87,19 @@ class CoherenceOrder(
     override fun compute() {
         check(map.isEmpty())
         generate(execution, memoryAccessEventIndex, rmwChainsStorage, writesOrder).forEach { coherence ->
+            // TODO: Also maybe lets keep coherence checker for SC and make a RA coherence checker somewhere else
+            // TODO: there is a probably a better place for these two relations
+            val initRelation = Relation<AtomicThreadEvent> { x, _ -> x.label is InitializationLabel }
+            val sameLocation = Relation<AtomicThreadEvent> { x, y -> getLocationForSameLocationAccesses(x,y) != null }
```
eupp: Let's either move to `Event.kt` next to `programOrder` relation, or move them and other similar relations from `Event.kt` into a separate file.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/CoherenceChecker.kt
```diff
@@ -87,12 +87,19 @@ class CoherenceOrder(
...
+            val hbLoc = (happensBeforeOrder intersection sameLocation) union initRelation
```
eupp: This should be parametrized. 
So `CoherenceOrder` should take some relation argument. 
For `SeqCst` you pass `causality order as before.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/SequentialConsistencyChecker.kt
```diff
@@ -53,9 +53,10 @@ class SequentialConsistencyChecker(
         check(execution.executionOrderComputable.computed)
         val executionOrder = execution.executionOrderComputable.value
             .ensure { it.isConsistent() }
-        SequentialConsistencyReplayer().ensure {
-            it.replay(executionOrder.ordering) != null
-        }
+        // TODO: The current coherence checker is just for RA, we need to seperate the SC coherence from RA coherence
```
eupp, line 57: Probably we need to extract common part into a new class. Then have two subclasses for SeqCst & RelAcq.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
```diff
@@ -75,6 +75,10 @@ interface ThreadEvent : Event {
      */
     val causalityClock: VectorClock
 
+    // TODO: maybe generalize the notion of vector clocks somehow, to avoid making a new vector clock for each
+    //   freaking relation we want to track
+    val happensBeforeClock: VectorClock
```
eupp, line 78: This is fine actually, no point to generalize to other relations at this point.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
```diff
@@ -303,6 +307,18 @@ abstract class AbstractThreadEvent(
         }
     }
 
+    final override val happensBeforeClock: VectorClock = run {
+        dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
+            if(this.label.isAcquire() && event.label.isRelease() ) {
```
eupp: Note for the future: for fences we will need to adjust.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
```diff
@@ -248,13 +248,13 @@ private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAcces
...
                     val lastWrite = getLastWrite(location)!!
-                    info.isWriteWriteRaceFree = causalityOrder(lastWrite, event)
+                    info.isWriteWriteRaceFree = happensBeforeOrder(lastWrite, event)
                 }
                 if (info.isReadWriteRaceFree) {
                     // to detect read-write race,
                     // we need to check against all the read-request events
                     info.isReadWriteRaceFree = getReadRequests(location).all { read ->
-                        causalityOrder(read, event)
+                        happensBeforeOrder(read, event)
```
eupp, line 257: Just in case, we need to check that in case of `SeqCst`: `happensBeforeOrder == causalityOrder`.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventLabel.kt
```diff
@@ -698,6 +698,17 @@ fun EventLabel.isExclusiveWriteAccess(): Boolean =
 fun EventLabel.isInitializingWriteAccess(): Boolean =
     this is InitializationLabel || this is ObjectAllocationLabel
 
+fun EventLabel.isRelease(): Boolean =
+    (this is WriteAccessLabel && (this.memoryOrdering == MemoryOrdering.RELEASE || this.memoryOrdering == MemoryOrdering.VOLATILE)) ||
+    (this is ThreadForkLabel) ||
+    (this is ThreadFinishLabel)
```
eupp: TODO: add locks and stuff

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Execution.kt
```diff
@@ -418,6 +420,32 @@ fun<E : ThreadEvent> Execution<E>.buildGraph(
 
 }
 
+// NOTE: I am not sure exactly about the assumptions needed in the orgininal buildGraph
+// But it seems that it is not correct for arbitary relations
+// TODO: figure them out and get rid of this alt version
+fun<E : ThreadEvent> Execution<E>.buildGraphAlt(
```
eupp: I believe we should merge two `buildGraph` implementations. 
And for relation that respect program order, use the other impl., for those that do not repsect, use this impl.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExtendedExecution.kt
```diff
@@ -153,7 +153,7 @@ fun MutableExtendedExecution(): MutableExtendedExecution =
             execution,
             memoryAccessEventIndex,
             readModifyWriteOrderComputable.value,
-            causalityOrder
+            happensBeforeOrder // TODO: maybe have a variable for "the order" instead of changing it everywhere
```
eupp: Maybe pass it as a parameter (either causality order or happens before order depending on the memory model).

lincheck/src/jvm/test-jdk11/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/JamMemoryModelTests.kt
```diff
@@ -0,0 +1,1493 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Lesser Public License for more details.
+ *
+ * You should have received a copy of the GNU General Lesser Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/lgpl-3.0.html>
+ */
+
+package org.jetbrains.kotlinx.lincheck_test.strategy.eventstructure
```
eupp, line 21: Let's rebase, this should already be in `develop`.

eupp (changes requested):

viliamDokov:
Added the plumbing to have the memory as a parameter. However I do not like the current setup that we have now. It is just a coincidence that RA is so similar to the SC and we can reuse so many of parts that were part of the SC checker.

Perhaps we can make the whole ExtendedExecution parametric on the memory model. The relations and computable values we can keep, in case different checkers want to reuse them. 

So instead of the current setup where we have one mega ExtendedExecution with many checkers, and each checker/relation has special logic per memory model, instead we can have many extended execution which different but maybe the same checkers/relation that are memory model agnostic.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
```diff
@@ -75,6 +75,10 @@ interface ThreadEvent : Event {
      */
     val causalityClock: VectorClock
 
+    // TODO: maybe generalize the notion of vector clocks somehow, to avoid making a new vector clock for each
+    //   freaking relation we want to track
```
eupp: Let's remove this comment. 
Causality/happens-before should cover all of what we will need for JMM support : )

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
```diff
@@ -303,6 +307,19 @@ abstract class AbstractThreadEvent(
         }
     }
 
+    // TODO: In the future we need also resolve release-acquire fences
+    final override val happensBeforeClock: VectorClock = run {
+        dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
+            if(this.label.isAcquire() && event.label.isRelease() ) {
```
eupp:
````suggestion
            if (this.label.isAcquire() && event.label.isRelease()) {
````

File: lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt [hunk 13]
```diff
@@ -436,10 +453,21 @@ val programOrder = Relation<ThreadEvent> { x, y ->
...
+
 val causalityOrder = Relation<ThreadEvent> { x, y ->
     (x != y) && y.causalityClock.observes(x.threadId, x.threadPosition)
 }
 
+val happensBeforeOrder = Relation<ThreadEvent> { x, y ->
+    (x != y) && y.happensBeforeClock.observes(x.threadId, x.threadPosition)
+}
+
+val happensBeforeLocOrder = (happensBeforeOrder intersection sameLocation) union initRelation
```
eupp: `happensBeforeSameLocationOrder` ?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
```diff
@@ -174,16 +174,16 @@ fun AtomicMemoryAccessEventIndex.getLastWrite(location: MemoryLocation): AtomicT
...
-fun MutableAtomicMemoryAccessEventIndex(): MutableAtomicMemoryAccessEventIndex =
-    MutableAtomicMemoryAccessEventIndexImpl()
+fun MutableAtomicMemoryAccessEventIndex(causalOrder: Relation<AtomicThreadEvent>): MutableAtomicMemoryAccessEventIndex =
+    MutableAtomicMemoryAccessEventIndexImpl(causalOrder)
 
 typealias AtomicMemoryAccessEventClassifier =
         EventIndexClassifier<AtomicThreadEvent, AtomicMemoryAccessCategory, MemoryLocation>
 
-private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAccessEventIndex {
+private class MutableAtomicMemoryAccessEventIndexImpl(val causalOrder: Relation<AtomicThreadEvent>) : MutableAtomicMemoryAccessEventIndex {
```
eupp: In the context of `MutableAtomicMemoryAccessEventIndexImpl` this `causalOrder` relation is used only for race detection; and data races are usually defined in terms of happens-before order, so I would suggest renaming it to `happensBefore` order. Or, perhaps, we should not even parametrize this by relation, and just use "global" `happensBefore` order here. 

The reason why `causalityOrder` was used here here before is that in the context of SeqCst model `happensBefore == causalityOrder`, because effectively every read is acquire read, and every write is release write.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
eupp: I think that in case of SeqCst memory model we wanna ensure that `happensBefore == causalityOrder`. I think one way to achieve that could be to change how event `label`-s are created for events --- instead of actual `memoryOrder` of access we can always pass `Volatile` if we are checking under SeqCst.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventLabel.kt
```diff
@@ -698,6 +698,21 @@ fun EventLabel.isExclusiveWriteAccess(): Boolean =
 fun EventLabel.isInitializingWriteAccess(): Boolean =
     this is InitializationLabel || this is ObjectAllocationLabel
 
+fun EventLabel.isRelease(): Boolean =
+    (this is WriteAccessLabel && (this.memoryOrdering == MemoryOrdering.RELEASE || this.memoryOrdering == MemoryOrdering.VOLATILE)) ||
```
eupp: Just to respect 120-chars per line limit:

```kotlin
(this is WriteAccessLabel && (
        this.memoryOrdering == MemoryOrdering.RELEASE || 
        this.memoryOrdering == MemoryOrdering.VOLATILE
)) ||
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventLabel.kt
```diff
@@ -698,6 +698,21 @@ fun EventLabel.isExclusiveWriteAccess(): Boolean =
...
+    (this is ThreadForkLabel) ||
+    (this is ThreadFinishLabel) ||
+    (this is UnlockLabel) ||
+    (this is NotifyLabel)
+
+
+fun EventLabel.isAcquire(): Boolean =
+    (this is ReadAccessLabel && (this.memoryOrdering == MemoryOrdering.ACQUIRE || this.memoryOrdering == MemoryOrdering.VOLATILE)) ||
```
eupp: Same here:

```kotlin
(this is ReadAccessLabel && (
        this.memoryOrdering == MemoryOrdering.ACQUIRE || 
        this.memoryOrdering == MemoryOrdering.VOLATILE
)) ||
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -381,25 +381,29 @@ internal class EventStructureStrategy(
         startingThreadDescriptor: ThreadDescriptor
     ) {
         super.beforeThreadStart(threadDescriptor, startingThread, startingThreadDescriptor)
-        val newThreadId = threadScheduler.getThreadId(startingThread)
-        //NOTE: Main thread is special cased, since it is forked from the "initial" thread
-        //      which is not tracked in the eventstrcuture. Therefore, we want to skip tracking it
-        if (newThreadId != -1 || newThreadId != eventStructure.mainThreadId) {
-            val currentThreadId = threadScheduler.getCurrentThreadId()
-            eventStructure.addThreadForkEvent(currentThreadId, setOf(newThreadId))
+        runInsideIgnoredSection {
```
eupp: You can as well wrap whole function into `runInsideIgnoredSection` :
```kotlin
override fun beforeThreadStart(
        startingThread: Thread,
        startingThreadDescriptor: ThreadDescriptor
) = threadDescriptor. runInsideIgnoredSection {
```

Ignored sections are re-entrant, so it should be the same. 

At least, please add `threadDescriptor.runInsideIgnoredSection` --- it saves one additional thread-local lookup.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -381,25 +381,29 @@ internal class EventStructureStrategy(
...
         //   and the `currentThread` lock is held
-        threadScheduler.awaitTurn(threadId)
-        // TODO: extract this check into a method ?
-        while (eventStructure.inReplayPhase() && !eventStructure.canReplayNextEvent(threadId)) {
-            switchCurrentThread(threadId, null) // TODO: Should not be null, since we would not switch we need another blocking reason
+        runInsideIgnoredSection {
+            threadScheduler.awaitTurn(threadId)
+            // TODO: extract this check into a method ?
+            while (eventStructure.inReplayPhase() && !eventStructure.canReplayNextEvent(threadId)) {
+                switchCurrentThread(threadId, null) // TODO: Should not be null, since we would not switch we need another blocking reason
```
eupp, line 403: We can add `ThreadFinish` blocking reason --- it would actually also simplify few things in `ManagedStrategy`. But it's probably better to do in a separate PR.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -381,25 +381,29 @@ internal class EventStructureStrategy(
...
 
     override fun onThreadFinish(threadId: Int) {
         // TODO: refactor, make `switchCurrentThread` private again in ManagedStrategy,
         //   call overridden `onStart` and `onFinish` methods only when thread is active
         //   and the `currentThread` lock is held
-        threadScheduler.awaitTurn(threadId)
-        // TODO: extract this check into a method ?
-        while (eventStructure.inReplayPhase() && !eventStructure.canReplayNextEvent(threadId)) {
-            switchCurrentThread(threadId, null) // TODO: Should not be null, since we would not switch we need another blocking reason
+        runInsideIgnoredSection {
```
eupp, line 399: Same comment here about `threadDescriptor.runInsideIgnoredSection`

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExtendedExecution.kt
```diff
@@ -130,19 +130,39 @@ interface MutableExtendedExecution : ExtendedExecution, MutableExecution<AtomicT
...
+    ExtendedExecutionImpl(ResettableExecution(), memoryModel)
 
 
 /* private */ class ExtendedExecutionImpl(
-    val execution: ResettableExecution
+    val execution: ResettableExecution,
+    val memoryModel: MemoryModel,
 ) : MutableExtendedExecution, MutableExecution<AtomicThreadEvent> by execution {
 
+    val causalOrder :  Relation<AtomicThreadEvent>
```
eupp: I would reserve `causalityOrder` terminology for `programOrder union readsFrom` to avoid ambiguity. 

Using `happensBefore` instead actually would make more sense here: under SeqCst `happensBefore == causalityOrder`. 

Alternatively, we can just ensure that under SeqCst, when we build `happensBefore` we treat every read as acquire and every write as release (regardless of actual `memoryOrder` passed --- see my other comment about this) --- so then we woud ensure `happensBefore == causalityOrder` property for SC by construction --- and we would no need this case analysis at all.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/SequentialConsistencyChecker.kt
```diff
@@ -53,8 +54,12 @@ class SequentialConsistencyChecker(
         check(execution.executionOrderComputable.computed)
         val executionOrder = execution.executionOrderComputable.value
             .ensure { it.isConsistent() }
-        SequentialConsistencyReplayer().ensure {
-            it.replay(executionOrder.ordering) != null
+
+        // TODO: The current coherence checker is just for RA, we need to seperate the SC coherence from RA coherence
+        if(memoryModel == MemoryModel.SequentialConsistency) {
```
eupp:
```suggestion
        if (memoryModel == MemoryModel.SequentialConsistency) {
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/MemoryModelConsistencyChecker.kt
```diff
@@ -53,8 +54,12 @@ class SequentialConsistencyChecker(
...
-            it.replay(executionOrder.ordering) != null
+
+        // TODO: The current coherence checker is just for RA, we need to seperate the SC coherence from RA coherence
+        if(memoryModel == MemoryModel.SequentialConsistency) {
+            SequentialConsistencyReplayer().ensure {
+                it.replay(executionOrder.ordering) != null
+            }
         }
         return null
     }
```
eupp, line 64: If we about to rename `SequentialConsistencyChecker` may I also ask you to remove dead commented code below? It was part of alternative SeqCst checker implementation --- and if we renaming the class it does not belong here anymore. Besides it's dead code anyway; and if we ever will need it in the future for some reason --- we can restore it via git history.

eupp (changes requested):

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: Actually this change was also added in the Locks PR, since it was a necessary bugfix there as well. 
I will do the fix there first, since it probably will get merged first as it is smaller.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/MemoryModelConsistencyChecker.kt
viliamDokov: Wait what should we rename SequentalConsistencyChecker to? Just ConsistencyChecker?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: Nevermind, CI is already running, I pushed the change here

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: There is no threadDescriptor passed as a parameter? Should I change it so there is one?

viliamDokov:
I changed it so that the memory order of every memory access is Volatile if we are on SC. 
Now happensBefore == causalOrder. Question is if we should change AtomicMemoryAccessEventIndex to not take happensBefore as a relation? For the not rc11 style definition of JAM happens before is a bit more complicated than that so maybe not? But it is also very likely that we can get away with it

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
eupp: Oh, sorry, my bad. Let's not change it then here, the problem is unrelated to this PR.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/consistency/MemoryModelConsistencyChecker.kt
eupp: > Wait what should we rename SequentalConsistencyChecker to? Just ConsistencyChecker?

Sorry, it looks like one of my comments was lost somehow...

It looks weird that now we have a class named `SequentalConsistencyChecker` which actually checks both SeqCst or RelAcq depending on what `memoryModel` argument is passed. 

I propose to rename it then. I have no good idea for the name though. `ConsistencyChecker` is already taken by the interface. `MemoryModelConsistencyChecker` is my best idea right now, although it's a bit generic...

If you have other suggestions, let's discuss.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/Event.kt
```diff
@@ -303,6 +305,19 @@ abstract class AbstractThreadEvent(
         }
     }
 
+    // TODO: In the future we need also resolve release-acquire fences
+    final override val happensBeforeClock: VectorClock = run {
+        dependencies.fold(parent?.happensBeforeClock?.copy() ?: MutableVectorClock()) { clock, event ->
+            if (this.label.isAcquire() && event.label.isRelease() ) {
```
eupp:
```suggestion
            if (this.label.isAcquire() && event.label.isRelease()) {
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventLabel.kt
```diff
@@ -642,6 +642,7 @@ fun ReadModifyWriteAccessLabel(read: ReadAccessLabel, write: WriteAccessLabel):
             writeValue = write.value,
             readModifyWriteDescriptor = read.readModifyWriteDescriptor!!,
             codeLocation = read.codeLocation,
+            // NOTE: Possible bug? We should include the memory order of the write somehow
```
eupp, line 645: True, perhaps `ReadModifyWriteAccessLabel` should have two modes for both read/write part. 
It's kinda independent problem though, I would leave as is for now.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventLabel.kt
```diff
@@ -698,6 +699,25 @@ fun EventLabel.isExclusiveWriteAccess(): Boolean =
 fun EventLabel.isInitializingWriteAccess(): Boolean =
     this is InitializationLabel || this is ObjectAllocationLabel
 
+fun EventLabel.isRelease(): Boolean =
+    (this is WriteAccessLabel && (
+    this.memoryOrdering == MemoryOrdering.RELEASE || this.memoryOrdering == MemoryOrdering.VOLATILE
```
eupp:
```suggestion
            this.memoryOrdering == MemoryOrdering.RELEASE || this.memoryOrdering == MemoryOrdering.VOLATILE
```

eupp:
> Question is if we should change AtomicMemoryAccessEventIndex to not take happensBefore as a relation? For the not rc11 style definition of JAM happens before is a bit more complicated than that so maybe not? But it is also very likely that we can get away with it.

I would say that if we don't need it in the context of this PR let's not change it and let's not make `happensBefore` a parameter of `AtomicMemoryAccessEventIndex`. If in the context of JAM PR we re-discover it is needed, we can always add it back.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -1271,4 +1271,14 @@ internal class EventStructure(
         readCodeLocationsCounter.keys.retainAll { (tid, _) -> tid != iThread }
     }
 
+    // NOTE: In the case of sequential consistency, to make sure that
+    // happensBeforeOrder === causalityOrder, we change the memory ordering of
+    // reads to VOLATILE
```
eupp:
```suggestion
    // reads to VOLATILE, effectively treating every read/write as sequentially-consistent (aka `volatile`)
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -1271,4 +1271,14 @@ internal class EventStructure(
...
+    private fun memoryOrderOverride(memoryOrder: MemoryOrdering) : MemoryOrdering {
+        if(memoryModel == MemoryModel.SequentialConsistency) {
```
eupp:
```suggestion
        if (memoryModel == MemoryModel.SequentialConsistency) {
```

eupp (approved):

PR #1026, opened 2026-06-12
Add proper support for locks in EventStructureStrategy

viliamDokov:
Notably we change how blocked events works. Now we do not create new blocked during the backward revisit itself, and instead just pass on whatever blocked events we encountered during normal execution. If we carefully pin events, we can guarantee termination and completeness.


lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1522,4 +1469,342 @@ class PrimitivesTest {
             return@litmusTest b1
         }
     }
+}
+
+
+class LocksTest {
```
viliamDokov: Probably move this to a different file? Primitives test is getting a bit big

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -738,33 +738,42 @@ private class EventStructureMonitorTracker(
     override fun registerThread(threadId: Int) {}
 
     override fun acquiringThreads(monitor: Any): List<Int> {
-        TODO("Not yet implemented")
+        val mutexID = eventStructureObjectTracker[monitor]!!.objectNumber.toLong()
+        return waitLockStack.mapNotNull { if (it.value?.mutexID == mutexID) it.key else null }
     }
 
     override fun interruptWait(threadId: Int) {
```
viliamDokov, line 745: I decided to not handle wait/notify properly as non of the current data structures use it.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -709,27 +708,28 @@ private class EventStructureMemoryTracker(
 
 }
 
-// NOTE: Some issues here as well , there are some missing members
 private class EventStructureMonitorTracker(
     private val eventStructure: EventStructure,
     private val eventStructureObjectTracker: EventStructureObjectTracker,
 ) : MonitorTracker {
 
+    data class LockWaitData(val mutexID: ValueID, val lockStack: MutableList<AtomicThreadEvent>)
```
viliamDokov: Not sure if this is the proper shape of the data class.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -738,33 +738,42 @@ private class EventStructureMonitorTracker(
...
     }
 
     // NOTE: This should be a bool?
     override fun releaseMonitor(threadId: Int, monitor: Any): Boolean {
-        return issueUnlock(threadId, monitor.opaque())
+        return issueUnlock(threadId, monitor)
     }
 
 
-    private fun issueLockRequest(iThread: Int, monitor: OpaqueValue): AtomicThreadEvent {
```
viliamDokov, line 771: Object tracker does not work with OpaqueValues directly now. Maybe we can completely get rid of them?

viliamDokov: Opaque value is currently only used to pass the object to `registerValueIfAbsent`, which takes is a two line method that takes in an optional `OpaqueValue`, handles the null and passes it on  `registerObjectIfAbsent`, by unwrapping the opauque value. We can skip this and use `registerObjectIfAbsent` directly.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -738,33 +738,42 @@ private class EventStructureMonitorTracker(
...
+    private fun issueLockRequest(iThread: Int, monitor: Any): AtomicThreadEvent {
         val mutexID = eventStructureObjectTracker[monitor]!!.objectNumber
```
viliamDokov: This is also a very common and annoying pattern. We can maybe use `ObjectTracker.getObjectNumber`, but annoyingly this method does not throw NPE if it it does not find the object in the object tracker and instead returns a default value, which is not something we really want. Maybe we can make two methods, one which throws NPE if object missing and one which does not (which is the current one).

viliamDokov: Also it is annoying that we need to constantly cast ObjectNumber to ValueID, which is a long. Maybe we can use ObjectNumber directly?

viliamDokov: Ok I went in and resoloved this

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
viliamDokov: Also resolved this. Leaving it in case you disagree

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -324,21 +324,7 @@ internal class EventStructure(
...
-                // We already compute the conflicting events when they are added.
-                // This way we do not have to compute them here every time
-                val conflicts = getConflictingEvents(
-                    it.threadId,
-                    nextEvent.label,
-                    it,
-                    it.dependencies
-                ).filter { it != nextEvent }
-                return@filter conflicts.isNotEmpty()
+                blockedEvents.values.any { blockedDesc -> blockedDesc.request == it && blockedDesc.response == null }
```
eupp: Are you sure `&& blockedDesc.response == null` part is correct?

By definition: 
```kotlin
val danglingRequests = frontier.getDanglingRequests()
```
So the frontier will only store request parts of blocked events, even if it happend that `blockedDesc.response != null` in the current execution. So after you'll restore execution graph after backtracking, you might end up having some events in the execution that are blocked requests, but were not added to  `blockedRequests`. Is it a desired behavior?

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -591,6 +577,26 @@ internal class EventStructure(
             label is ReadAccessLabel && label.isResponse ->
                 sequenceOf()
 
+            // Non-re-entrant lock-request synchronizes only with non-reentrant unlock event or allocation event that are not pinned
+            (label is LockLabel && !label.isReentry) -> {
+                // We can either sync with the allocation event of the mutex or with one of the unlock events
+                val unlockEventCandidates = sequenceOf(allocationEvent(label.mutexID)!!) +
+                        execution.filter { it.label.refine<UnlockLabel> { !isReentry && mutexID == label.mutexID } != null }
+
+                //But we cannot synch with unlock events that are already being read by a pinned lock response
```
eupp:
```suggestion
// But we cannot synch with unlock events that are already being read by a pinned lock response
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructureStrategy.kt
```diff
@@ -381,25 +376,29 @@ internal class EventStructureStrategy(
...
         super.beforeThreadStart(threadDescriptor, startingThread, startingThreadDescriptor)
-        val newThreadId = threadScheduler.getThreadId(startingThread)
-        //NOTE: Main thread is special cased, since it is forked from the "initial" thread
-        //      which is not tracked in the eventstrcuture. Therefore, we want to skip tracking it
-        if (newThreadId != -1 || newThreadId != eventStructure.mainThreadId) {
-            val currentThreadId = threadScheduler.getCurrentThreadId()
-            eventStructure.addThreadForkEvent(currentThreadId, setOf(newThreadId))
+        runInsideIgnoredSection {
+            val newThreadId = threadScheduler.getThreadId(startingThread)
+            //NOTE: Main thread is special cased, since it is forked from the "initial" thread
```
eupp:
```suggestion
// NOTE: Main thread is special cased, since it is forked from the "initial" thread
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ObjectTracker.kt
```diff
@@ -232,9 +232,12 @@ fun ObjectID.getObjectHashCode(): Int =
  * @return the unique object number if the object is registered in the tracker,
  *   or -1 if no entry is associated with the given object.
  */
-fun ObjectTracker.getObjectNumber(obj: Any): Int =
+fun ObjectTracker.getObjectNumberOrDefault(obj: Any): Int =
     get(obj)?.objectNumber ?: -1
 
+fun ObjectTracker.getObjectNumber(obj: Any): Int =
```
eupp, line 247: Please add KDoc comment (similar to the one for `getObjectNumberOrDefault`).

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1522,4 +1469,342 @@ class PrimitivesTest {
...
+            t1.join()
+            t2.join()
+
+            Triple(r1, r2, x)
+        }
+    }
+
+
+    @Test
+    fun testFooLock() {
```
eupp: What `Foo` stands for? Can we give a more descriptive name to this test ?)

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1522,4 +1469,342 @@ class PrimitivesTest {
...
+            t1.join()
+            t2.join()
+            t3.join()
+            t4.join()
+            listOf(r1,r2,r3,r4)
+        }
+    }
+
+    @Test
+    fun testWaitNotifySwap() {
```
eupp: Please move towards `testWaitNotify` test.

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1522,4 +1469,342 @@ class PrimitivesTest {
...
+        val writeAndNotify = SynchronizedVariable::writeAndNotify
+        val waitAndRead = SynchronizedVariable::waitAndRead
+        val testScenario = scenario {
+            parallel {
+                thread {
+                    actor(waitAndRead)
```
eupp: Do you need to use scenario/actors for this test? Can we rewrite it without them?

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
eupp: Ok, let's extract into separate file. But then let's also move there other lock-related tests from `PrimitivesTest`.

eupp (changes requested):
Left few minor comments.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
viliamDokov: I started typing out a comment but it was wayy too long. The summary is no it is not, but it would maybe not happen in practice because of the `addUnblockingResponses` part earlier.

eupp (approved):

PR #1028, opened 2026-06-17
Reduce bound for spin-wait loops in ActiveThreadPoolExecutor

viliamDokov:
TODO: Also make this configurable per strategy? Possibly only managed strategy would need this change.

PR #1030, opened 2026-06-17
Park and Unpark thread waiting for awaitTurn.

viliamDokov:
In order to clean up the profiling trace, we now park/unpark the thread in awaitTurn of threadScheduler. Maybe add an option to disable this like in the profiling case?

### DIFF
--- a/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
@@ -76,7 +76,7 @@ class Spinner private constructor(
      * the spin-loop should perform before yielding to other threads.
      */
     fun pollYieldLimit(): Int =
-        1 + if (isSpinning) SPIN_CYCLES_LIMIT else 0
+        1 + if (isSpinning) spinLimit else 0
 
     /**
      * Determines the limit for the number of iterations
@@ -235,4 +235,4 @@ fun SpinnerGroup(nThreads: Int, spinLimit: Int = SPIN_CYCLES_LIMIT): List<Spinne
 
 //NOTE: Should be powers of 2
 const val SPIN_CYCLES_LIMIT: Int = 1_048_576 // 2^20
-const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024
+const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024 // 2^10
--- a/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
@@ -9,6 +9,7 @@
  */
 
 package org.jetbrains.lincheck.util
+import java.util.concurrent.locks.LockSupport
 
 /**
  * A spinner implements utility functions for spinning in a loop.
@@ -75,6 +76,13 @@ class Spinner private constructor(
      * the spin-loop should perform before yielding to other threads.
      */
     fun pollYieldLimit(): Int =
+        1 + if (isSpinning) PARK_CYCLES_LIMIT else 0
+
+    /**
+     * Determines the limit for the number of iterations
+     * the spin-loop should perform before parking the current thread
+     */
+    fun pollParkLimit(): Int =
         1 + if (isSpinning) spinLimit else 0
 
     /**
@@ -110,6 +118,28 @@ class Spinner private constructor(
             }
             if (counter % pollCount == 0) {
                 limit = pollYieldLimit()
+            }
+        }
+    }
+
+    /**
+     * Waits in the spin-loop until the given condition is true
+     * with periodical yielding to other threads.
+     *
+     * @param condition A lambda function that determines the condition to wait for.
+     *   The function should return true when the condition is satisfied, and false otherwise.
+     */
+    inline fun spinWaitUntilOrPark(condition: () -> Boolean) {
+        var counter = 0
+        var limit = pollParkLimit()
+        val pollCount = PARK_CYCLES_LIMIT
+        while (!condition()) {
+            counter++
+            if (counter % limit == 0) {
+                LockSupport.park()
+            }
+            if (counter % pollCount == 0) {
+                limit = pollParkLimit()
             }
         }
     }
@@ -205,4 +235,5 @@ fun SpinnerGroup(nThreads: Int, spinLimit: Int = SPIN_CYCLES_LIMIT): List<Spinne
 
 //NOTE: Should be powers of 2
 const val SPIN_CYCLES_LIMIT: Int = 1_048_576 // 2^20
-const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024 // 2^10
+const val PARK_CYCLES_LIMIT: Int = 1024  // 2^10
+const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024
--- a/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
+++ b/common/src/main/org/jetbrains/lincheck/util/Spinner.kt
@@ -76,7 +76,7 @@ class Spinner private constructor(
      * the spin-loop should perform before yielding to other threads.
      */
     fun pollYieldLimit(): Int =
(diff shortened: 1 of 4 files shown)


lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/ThreadScheduler.kt
```diff
@@ -335,6 +336,8 @@ open class ThreadScheduler {
     fun abortThread(threadId: ThreadId) {
         threads[threadId].apply {
             state = ThreadState.ABORTED
+            val thread = descriptor.thread
+            LockSupport.unpark(thread)
         }
```
eupp:
```suggestion
            LockSupport.unpark(descriptor.thread)
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedThreadScheduler.kt
```diff
@@ -88,7 +91,13 @@ class ManagedThreadScheduler : ThreadScheduler() {
...
                 raiseThreadAbortError()
             }
-            scheduledThreadId == threadId
+
+            if(scheduledThreadId == threadId) {
+                return@spinWaitUntil true
+            }
+
+            LockSupport.park()
+            false
```
eupp: You always need to wrap `LockSupport.park()` into `while` loop that checks the condition you are waiting for, because of the spurious wake-ups:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/LockSupport.html#park--

eupp: To handle this whole thing a bit more elegantly, let's add new extension function for `Spinner`: `spinWaitUntilOrPark`.

It should be same as `spinWaitUntil` except after some spinning it should fallback to parking (and also run park in a `while` loop until `condition` becomes true).

eupp (changes requested):

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedThreadScheduler.kt
viliamDokov: I think this was already the case right? After park we returned false which meant that we need to continue spinning

viliamDokov: Anyway I added the spinWaitUntilOrPark method, you can have a look at that

common/src/main/org/jetbrains/lincheck/util/Spinner.kt
```diff
@@ -64,6 +65,13 @@ class Spinner private constructor(
      * the spin-loop should perform before yielding to other threads.
      */
     fun pollYieldLimit(): Int =
+        1 + if (isSpinning) PARK_CYCLES_LIMIT else 0
```
eupp:
```suggestion
        1 + if (isSpinning) SPIN_CYCLES_LIMIT else 0
```

common/src/main/org/jetbrains/lincheck/util/Spinner.kt
```diff
@@ -103,6 +111,28 @@ class Spinner private constructor(
...
+     * Waits in the spin-loop until the given condition is true
+     * with periodical yielding to other threads.
+     *
+     * @param condition A lambda function that determines the condition to wait for.
+     *   The function should return true when the condition is satisfied, and false otherwise.
+     */
+    inline fun spinWaitUntilOrPark(condition: () -> Boolean) {
+        var counter = 0
+        var limit = pollParkLimit()
+        val pollCount = PARK_CYCLES_LIMIT
```
eupp: Shouldn't it be just?:

```suggestion
        val pollCount = SPIN_CYCLES_LIMITS_POLL_COUNT
```

And after https://github.com/JetBrains/lincheck/pull/1028 the `pollParkLimit` will become configurable via same `spinLimit` parameter?

common/src/main/org/jetbrains/lincheck/util/Spinner.kt
```diff
@@ -103,6 +111,28 @@ class Spinner private constructor(
         }
     }
 
+    /**
+     * Waits in the spin-loop until the given condition is true
+     * with periodical yielding to other threads.
```
eupp:
```suggestion
     * and periodically parks thread after spinning for too long .
```

common/src/main/org/jetbrains/lincheck/util/Spinner.kt
viliamDokov:
```
    /**
     * Determines the limit for the number of iterations
     * the spin-loop should perform before parking the current thread
     */
    fun pollParkLimit(): Int =
        1 + if (isSpinning) spinLimit else 0
 ```
Poll park limit is now also determined by the spin limit. Maybe we can add a separate variable for spinner, but I do not think it it necessary right now.

File: common/src/main/org/jetbrains/lincheck/util/Spinner.kt [hunk 6]
```diff
@@ -75,6 +76,13 @@ class Spinner private constructor(
      * the spin-loop should perform before yielding to other threads.
      */
     fun pollYieldLimit(): Int =
+        1 + if (isSpinning) SPIN_CYCLES_LIMIT else 0
```
eupp:
```suggestion
        1 + if (isSpinning) spinLimit else 0
```

common/src/main/org/jetbrains/lincheck/util/Spinner.kt
```diff
@@ -205,4 +235,4 @@ fun SpinnerGroup(nThreads: Int, spinLimit: Int = SPIN_CYCLES_LIMIT): List<Spinne
 
 //NOTE: Should be powers of 2
 const val SPIN_CYCLES_LIMIT: Int = 1_048_576 // 2^20
-const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024 // 2^10
\ No newline at end of file
+const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024
```
eupp:
```suggestion
const val SPIN_CYCLES_LIMITS_POLL_COUNT = 1024 // 2^10
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/ManagedThreadScheduler.kt
```diff
@@ -84,7 +87,7 @@ class ManagedThreadScheduler : ThreadScheduler() {
     fun awaitTurn(threadId: ThreadId) {
         check(threadId == getCurrentThreadId())
         val threadData = threads[threadId]!!
-        threadData.spinner.spinWaitUntil {
+        threadData.spinner.spinWaitUntilOrPark {
```
eupp, line 92: @viliamDokov I think we also need to set `spinLimit` for `threadData.spinner` in case of `ManagedStrategy`, similarly as we did for other spinners in https://github.com/JetBrains/lincheck/pull/1028?

viliamDokov: Sorry I missed this comment. The reason I did not add spinLimit to this spinner is because it was only used for spinWaitUntil, which did not use `SPIN_CYCLES_LIMIT` and instead used `SPIN_CYCLES_LIMITS_POLL_COUNT=1024`, which is way lower. I can change it so we set the spin limit to something else in case someone uses this spinner with another method

viliamDokov: I pushed a change, tell me if you have any comments

eupp (approved):

PR #1032, opened 2026-06-19
Add docs for Lincheck in KMP projects

cmlimm:
New docs about setting up Lincheck in KMP projects


docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
+[//]: # (title: Lincheck in Kotlin Multiplatform projects)
+[//]: # (description: Learn how to set up Lincheck tests in multiplatform projects)
+
+Lincheck can be used with [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform/quickstart.html) for
```
zamulla:
```suggestion
Lincheck can be used with [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform/quickstart.html) 
```

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
+[//]: # (title: Lincheck in Kotlin Multiplatform projects)
+[//]: # (description: Learn how to set up Lincheck tests in multiplatform projects)
+
+Lincheck can be used with [Kotlin Multiplatform](https://kotlinlang.org/docs/multiplatform/quickstart.html) for 
+testing the code targeting the JVM.
```
zamulla:
```suggestion
to test the code targeting the JVM.
```

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+>
+> Lincheck relies heavily on JVM for [bytecode manipulation](lincheck-testing-strategies.md#model-checking) that is not 
+> supported on other platforms. 
+> 
+> Implementing model checking for multiple targets is not only challenging, but also provides 
+> little value. Model checking enables Lincheck to test the code in a deterministic 
+> sandbox environment, which means that the results of the tests will be the same regardless 
+> of the platform.
+> 
+{style="tip"}
```
zamulla: I think I get why this is highlighted the way it is, but I would still ask to consider whether the same message cannot be conveyed in the main body of the text.

Something like

> **Lincheck is unlikely to support other platforms in the future.** Model checking for multiple targets is both challenging to implement and provides little value. ...

zamulla: Another solution is to make this a section, to avoid breaking up the intro pattern — I feel like I got distracted from the purpose of the article before it even started.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+
+You will:
+* Set up your project to run Lincheck tests for targets using the JVM.
+* Create a `Counter` structure shared between all targets.
+* Create tests for `Counter` shared between all targets and tests specific to the JVM.
+* Run the tests.
+
+## Set up a project
+
+1. Create a [new Kotlin Multiplatform project](https://kotlinlang.org/docs/multiplatform/quickstart.html) in IntelliJ IDEA or open an existing one.
```
zamulla: I'd avoid talking about an existing project because it adds a lot of complexity and makes us constantly repeat "if your project <is unlike the default one> ..." which is hard to read and to follow. Having a clear working example should be easy to translate into an existing project, right?

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+
+   ```kotlin
+   kotlin {
+       // ...
+       
+       jvm()
+       
+       // ...
+       
+       // If you have an Android app, the use of JVM is
```
zamulla:
```suggestion
       // If you have an Android target, the use of JVM is
```

KMP necessarily produces an Android library, but there might not be an Android app.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+           // ...
+       }
+   }
+   ```
+   
+2. Set up the tests folder and Lincheck dependency according to the target platforms in your project:
+   * [**Desktop/Android + iOS**](#server-or-desktop-or-android-ios)
+   * [**Server + iOS, Server + Desktop**, **Server + Desktop + iOS**](#server-desktop-android-ios)
+   * **Server/Desktop + Android**, **Server/Desktop + Android + iOS** – Kotlin doesn't currently support [sharing a source 
+     set for JVM + Android targets](https://kotlinlang.org/docs/multiplatform/multiplatform-hierarchy.html#manual-configuration).
```
zamulla: This limitation should be clear on the first step, no? Otherwise the user might create a desktop + Android project, arrive and this step, and have to start again.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+2. Set up the tests folder and Lincheck dependency according to the target platforms in your project:
+   * [**Desktop/Android + iOS**](#server-or-desktop-or-android-ios)
+   * [**Server + iOS, Server + Desktop**, **Server + Desktop + iOS**](#server-desktop-android-ios)
+   * **Server/Desktop + Android**, **Server/Desktop + Android + iOS** – Kotlin doesn't currently support [sharing a source 
+     set for JVM + Android targets](https://kotlinlang.org/docs/multiplatform/multiplatform-hierarchy.html#manual-configuration). 
+
+### Desktop/Android + iOS {id="server-or-desktop-or-android-ios"}
+
+The directory for the platform-specific tests is created by the Kotlin Multiplatform plugin automatically.
+For example, for a desktop target it is `shared/src/jvmTest`.
```
zamulla: It is created by default, but this is probably worth noting at the start: "make sure the **Include tests** checkbox stays checked" or something.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+                   implementation("org.jetbrains.kotlinx:lincheck:3.5")
+               }
+           }
+       }
+   }
+   ```
+
+## Create shared structures
+
+Create structures shared between all targets regardless of the platform:
```
zamulla: The article is talking about structures, but there are no structure entities in Kotlin :) the following examples clearly declare classes.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+
+1. Create an `UnsafeCounter` structure in the `core/src/commonMain` or `shared/src/commonMain` directory:
```
zamulla: We generally mention which file the code is supposed to be in. Does it matter if `UnsafeCounter` and `SafeCounter` are in the same file?

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+   class SafeCounterTest {
+      @Test
+      fun testIncrement() {
+         val counter = SafeCounter()
+         assertEquals(0, counter.get(), "Initial value should be 0")
+         counter.inc()
+         assertEquals(1, counter.get(), "Value after one increment should be 1")
+      }
+   }
+   ```
```
zamulla, line 197: It would be nice to mention that the user is supposed to add the missing imports themselves, especially since there are two potential imports for `assertEquals`.

zamulla: Or we can simply add the imports to the example.

docs/topics/lincheck-kmp.md
zamulla: Also, I tried to follow the instructions in a project with Desktop + Android + iOS, and didn't really get what is not supposed to work in this configuration. I can add dependencies for both JVM and Android, I can write tests and I can run them.

The question is, what is the actual call to action here? "Don't use Lincheck in Desktop + Android + iOS projects" doesn't seem to be it.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+            }
+        }
+    }
+    </code-block>
+</tab>
+</tabs>
+
+### Server + iOS, Server + Desktop, Server + Desktop + iOS {id="server-desktop-android-ios"}
+
+1. Create a `core/src/jvmTest` directory. Gradle will run the tests
```
zamulla:
```suggestion
1. Create a `core/src/jvmTest/kotlin` directory. Gradle will run the tests
```

docs/topics/lincheck-kmp.md
zamulla:
```suggestion
1. Create an `UnsafeCounter` structure in the `core/src/commonMain/kotlin` or `shared/src/commonMain/kotlin` directory:
```

docs/topics/lincheck-kmp.md
zamulla:
```suggestion
Create classes to be tested in common code:
```

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+   class UnsafeCounterConcurrentTest {
+       private val c = UnsafeCounter()
+   
+       @Operation
+       fun inc() = c.inc()
+   
+       @Operation
+       fun get() = c.get()
+   
+       @Test
```
zamulla, line 223: Here, too, it's unclear which `@Test` I'm supposed to use.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+
+The directory for the platform-specific tests is created by the Kotlin Multiplatform plugin automatically.
+For example, for a desktop target it is `shared/src/jvmTest`.
+
+Add a Lincheck dependency to the [test source set](https://kotlinlang.org/docs/multiplatform/multiplatform-add-dependencies.html#kotlinx-libraries):
+
+<tabs>
+<tab id="desktop" title="Desktop">
+    <code-block lang="Kotlin">
+    // app/shared/build.gradle.kts
```
zamulla:
```suggestion
    // shared/build.gradle.kts
```
If I understand the pattern correctly — without `server` we don't get an `app` directory.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,278 @@
...
+
+   ```kotlin
+   // core/build.gradle.kts
+   kotlin {
+       // ...
+       sourceSets {
+           // ...
+           jvmTest {
+               dependencies {
+                   implementation("org.jetbrains.kotlinx:lincheck:3.5")
```
zamulla: Gradle says: `:core:jvmTest: Could not find org.jetbrains.kotlinx:lincheck:3.5.`
Should this be different from the desktop dependency? When I go to import `ModelCheckingOptions`, IDEA says that the kotlinx version is deprecated.

zamulla (changes requested):
Thanks! Left some comments. Mainly I'm confused about the Android / Desktop incompatibility and whether I'm allowed to use Lincheck at all in an "all targets" project.

docs/topics/lincheck-kmp.md
cmlimm: Thanks, I was unaware of this!

docs/topics/lincheck-kmp.md
cmlimm: Fair, I will clarify that these are classes implementing counter data structures. I want to keep the word "structures" in because a big part of Lincheck is testing specifically data structures.

docs/topics/lincheck-kmp.md
cmlimm: Added a warning in the first step with a clarification that the user can still test their code

docs/topics/lincheck-kmp.md
cmlimm: Changed the wording to be the same as in other section ("Create a `filename.kt` file...")

docs/topics/lincheck-kmp.md
cmlimm: Added a section in the end of the article

docs/topics/lincheck-kmp.md
```diff
@@ -197,9 +210,13 @@ regardless of the platform:
 
 Write tests that are [only run for platforms targeting the JVM](https://kotlinlang.org/docs/multiplatform/multiplatform-run-tests.html#add-platform-specific-tests):
 
-1. In the [test directory for JVM-specific tests](#set-up-a-project), create an `UnsafeCounterConcurrentTest.kt` file:
+1. Create an `UnsafeCounterConcurrentTest.kt` file in the [test directory for JVM-specific tests](#set-up-a-project):
```
zamulla, line 203: It's a bit of a drag to follow this link, figure out what I need to see there, and then find my way back again. Should we just list possible options here?

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,308 @@
...
+You will:
+* Set up your project to run Lincheck tests for targets using the JVM.
+* Create classes for counter data structures shared between all targets.
+* Create tests for the counters shared between all targets and tests specific to the JVM.
+* Run the tests.
+
+## Set up a project
+
+1. Create a [new Kotlin Multiplatform project](https://kotlinlang.org/docs/multiplatform/quickstart.html) in IntelliJ IDEA. When creating a project, make sure that:
+   * The **Include tests** checkbox is checked when creating a project.
```
zamulla:
```suggestion
   * The **Include tests** checkbox is checked.
```

Is already there in the intro.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,308 @@
...
+2. Set up the tests folder and Lincheck dependency according to the target platforms in your project:
+   * [**Desktop/Android + iOS**](#server-or-desktop-or-android-ios)
+   * [**Server + iOS, Server + Desktop**, **Server + Desktop + iOS**](#server-desktop-android-ios)
+
+### Desktop/Android + iOS {id="server-or-desktop-or-android-ios"}
+
+The directory for the platform-specific tests is created by the Kotlin Multiplatform plugin automatically.
+For example, for a desktop target it is `shared/src/jvmTest`.
+
+Add a Lincheck dependency to the platform-specific [test source set](https://kotlinlang.org/docs/multiplatform/multiplatform-add-dependencies.html#kotlinx-libraries) and a Kotlin test dependency to the
```
zamulla:
```suggestion
Add a Lincheck dependency to the platform-specific [test source set](https://kotlinlang.org/docs/multiplatform/multiplatform-add-dependencies.html#kotlinx-libraries) and a `kotlin.test` dependency to the 
```

"Kotlin test" is a generic enough phrase to be confusing in some situations, here and elsewhere I'd give the specific library name.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,308 @@
...
+           jvmTest.dependencies {
+               implementation("org.jetbrains.lincheck:lincheck:%lincheckVersion%")
+           }
+       }
+   }
+   ```
+
+## Create shared classes
+
+Create classes implementing counter data structures that are shared between all targets regardless of the platform:
```
zamulla:
```suggestion
Create classes implementing counter data structures that are shared between all targets:
```

A bit tautological.

zamulla (commented):
LGTM overall, a couple of polishing suggestions.

docs/topics/lincheck-kmp.md
```diff
@@ -14,31 +14,32 @@ You will:
...
-              // ...
-          }
-      }
-      ```
+         jvm()
+         
+         // ...
+         
+         // If you have an Android target, the use of JVM is
+         // stated in the Android configuration section
```
zamulla: stated = configured?
Also I'm not sure what this is actually referring to — the following `androidLibrary {}` block? If so, I would just say exactly that, "configured in the `androidLibrary {}` block". Or some other block if I guessed incorrectly :) but better to be precise.

zamulla (approved):
Thanks! LGTM, with one final nit.

docs/topics/lincheck-kmp.md
```diff
@@ -0,0 +1,312 @@
...
+
+   ```kotlin
+   import kotlin.concurrent.atomics.AtomicInt
+   import kotlin.concurrent.atomics.ExperimentalAtomicApi
+
+   @OptIn(ExperimentalAtomicApi::class)
+   class SafeCounter {
+       private val _value = AtomicInt(0)
+   
+       var value: Int
```
eupp: For the sake of this example, I think it would be better if `UnsafeCounter` and `SafeCounter` have the same interface. Thus `value` should either be public or private in both. I would vote for making and private and exposing publically only `get()` and `inc()`.

eupp (approved):
Left one nit, please address it before we merge.

PR #1036, opened 2026-06-26
Add docs for verification and validation

cmlimm:
* Add docs explaining the verification and validation processes.
* Incorporate the Sequential specification article into the verification docs.


eupp (approved):

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+
+       fun poll(): Int? = synchronized(this) {
+           if (elements.isEmpty()) return null
+           elements.shuffle()
+           elements.removeAt(0)
+       }
+   }
+   ```
+
+   This concurrent structure behaves incorrectly: it stores the elements like a typical queue but returns them randomly.
```
danil-pavlov:
```suggestion
   This concurrent structure behaves incorrectly: it stores elements like in a typical queue, but returns them randomly.
```

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+    // Throw an exception if the invariant is violated
+    check(size >= 0) { "Size must be non-negative, but was $size" }
+}
+```
+
+The validation function should:
+* Accept no arguments.
+* Throw an exception if the data structure is in an invalid state.
+
+## See also
```
danil-pavlov:
```suggestion
## What's next
```

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+
+class SequentialQueue {
+    private val s = LinkedList<Int>()
+
+    fun add(x: Int) = s.add(x)
+    fun poll(): Int? = s.poll()
+}
+```
+
+### Understand the difference between linearizability and serializability
```
danil-pavlov: The header is too long, 4 lines in the right-hand menu. what about
Serializability model? since the whole section is about trying out this model.

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+* `LinearizabilityVerifier` – the default option. Concurrent execution is valid if there is a sequential execution 
+  that preserves the ["happens-before"](https://en.wikipedia.org/wiki/Happened-before) relations among the operations 
+  in the concurrent execution.
+* `QuiescentConsistencyVerifier` – works similarly to the linearizability model, but the "happens-before" constraints 
+  are not applied to the operations annotated with `@QuiescentConsistent`:
+
+   ```kotlin
+   @Operation
+   @QuiescentConsistent
+   fun foo() = { ... }
```
danil-pavlov: We're trying not to use foo/bar example and have some meaningful names in code samples instead

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+    private val s = LinkedList<Int>()
+
+    fun add(x: Int) = s.add(x)
+    fun poll(): Int? = s.poll()
+}
+```
+
+### Understand the difference between linearizability and serializability
+
+With the serializability verification model, it is enough for a concurrent execution to have any corresponding
```
danil-pavlov: I'm a bit confused with the flow here. first, these description of the two models should relate to the properties section, but there's Sequential specification section in between. Second, the 1st paragraph here is about serializability, then Linearizability, and then serializability again

maybe start with: "Compared to the default linearizability model that requires..., serializability model..."

docs/topics/lincheck-results-validation.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Understand the difference between linearizability and serializability
+
+With the serializability verification model, it is enough for a concurrent execution to have any corresponding 
+sequential execution that leads to the same results.
+
+Linearizability is a stricter verification model – the corresponding sequential execution should also preserve 
+the "happens-before" relationships among the operations.
+
+In this example, you will use Lincheck to show that a structure can be serializable, but not linearizable:
```
danil-pavlov: I think you can go straight to the example instead of the future tense here

"Let's see how Lincheck verifies data structures using the serializable model:"
or just drop this sentense

danil-pavlov (changes requested):

docs/topics/lincheck-results-validation.md
cmlimm: I do agree, the header is a bit too long, but its contents are not only about serializability. It is specifically about how linearizability and serializability are different from each other (we directly apply both models and have explanations for both of them at step 6). At least this was my intention:) If it is not clear, I can reformulate the contents to make the intention more apparent

WDYT of "Difference between linearizability and serializability" or even "Compare linearizability and serializability"? Unfortunately, the names of the properties are already quite long, so I think we cannot get a header shorter than at least 3 lines

danil-pavlov: Yeah, Compare would probably be the best option here

docs/topics/lincheck-results-validation.md
cmlimm: Agreed. I've moved the Sequential specification section to be the first in section, since there is no real reason to mention verification models first. This also allowed to not explain the models in the intro paragraph once again, because they are mentioned quite close in the article. I've also changed the header to H4, since it only relates to the Verification models subsection (also mitigating the problem with the header being too long in the ToC).

danil-pavlov (approved):

PR #1037, opened 2026-07-01
Add model checking docs

cmlimm:
* Add a dedicated article about Model checking
* Update the links to Model checking to reference the new article


docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+  or `java.lang.System.currentTimeMillis()`.
+* **Global variables** – Lincheck resets the values of global variables between invocations in model checking tests:
+
+  ```kotlin
+  class GlobalVariableResetTest {
+      companion object {
+          private var atomicInt = AtomicInteger(1)
+      }
+ 
+      @Operation
```
eupp: I think this example can be rewritten with lambda-based API to make it more concise.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Uncontrolled sources of non-determinism
+
+Lincheck is not able to control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it 
+  does with global variables. See an example of an error caused by the usage of thread-local variables and a workaround 
+  in a [dedicated section](#thread-local-variables).
+* **Weak references** – Lincheck does not control when a garbage collector removes objects that are only referenced 
+  by weak references. Calling `get()` on such objects produces a non-deterministic result which is treated as invalid
```
eupp: Let's remove "which is treated as invalid by Lincheck". It's just a non-deterministic result, not necessarily invalid. The test might even pass without "Non-determinism found." error with some probability.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+
+Lincheck is not able to control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it 
+  does with global variables. See an example of an error caused by the usage of thread-local variables and a workaround 
+  in a [dedicated section](#thread-local-variables).
+* **Weak references** – Lincheck does not control when a garbage collector removes objects that are only referenced 
+  by weak references. Calling `get()` on such objects produces a non-deterministic result which is treated as invalid 
+  by Lincheck.
+* **I/O API calls** – Lincheck does not support calls to I/O APIs, including operations on files, sockets, and networks.
```
eupp:
```suggestion
* **I/O API calls** – Lincheck does not support calls to I/O APIs, including operations on files and sockets. 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+
+## Bounded exploration
+
+The number of possible execution schedules grows exponentially with the size of the program, which increases testing 
+times. Lincheck circumvents this problem by limiting the [number of explored execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
+If the number of possible execution schedules exceeds the specified limit, Lincheck stops the exploration. 
+In such cases, Lincheck tries to evenly analyze logically different schedules:
+
+* Lincheck applies context bounding by first exploring schedules with one preemptive context switch, then two, and so on.
+* When choosing the next schedule to explore, Lincheck prioritizes schedules with context switches in new locations.
```
eupp:
```suggestion
* When choosing the next schedule to explore, Lincheck prioritizes schedules with thread switches in new code locations.
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+class Example {
+   var x = 0 // Not @Volatile
+   var y = 0 // Not @Volatile
+
+   fun thread1() {
+       x = 1
+       y = 1
+   }
+
+   fun thread2() {
```
eupp: Let's rewrite this example as Lincheck's lambda test:
```kotlin
Lincheck.runConcurrentTest {
    thread {
       x = 1
       y = 1
    }
    thread {
        if (y == 1 && x == 0) {
           // Code in this block might be executed on real hardware because of
           // store buffer and compiler reordering.
           // Lincheck cannot model this behavior with model checking.
           error("Unreachable under sequential consistency")
       }
    }
}
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+
+```kotlin
+class ThreadLocalVariableTest {
+  @Operation
+  fun inc(): Int {
+      return getLocalCounter().getAndIncrement()
+  }
+
+  @Test
+  fun modelCheckingTest() = ModelCheckingOptions()
```
eupp: Let's rewrite this example as Lincheck's lambda test.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,373 @@
...
+
+When running a test with the model checking strategy, Lincheck controls the following sources of non-determinism:
+
+* **Thread-switching** – instead of relying on the JVM for thread-switching, Lincheck inserts explicit thread-switch 
+  instructions at points of shared-memory access (`read` and `write`) or at synchronization points, such as lock 
+  acquisition and release, `park`/`unpark`, `wait`/`notify`, and others.
+* **Random number generators** – Lincheck fixes the random seed.
+* **Identity hash codes** – Lincheck fixes the [identity hash codes](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object) 
+  of objects.
+* **Time API calls** – Lincheck returns a [predefined constant](#time-api-calls) when calling `java.lang.System.nanoTime()`
```
eupp: I think here you can just tell that Lincheck's intercepts time APIs, returning deterministic results, and refer to "Time API calls" sub-section for details. The fact that "predefined constant" is returned is an implementation detail, that would probably be changed in the future.

eupp (approved):
Left few comments, mostly about the code snippets.
Otherwise LGTM.

docs/topics/lincheck-model-checking.md
cmlimm: Is there a reason we should use "thread switch" here specifically? Because we often refer to "context switching" and it is better to use the same term for the action everywhere. If I'm missing some difference between "thread switching" and "context switching", I will, of course, update the section accordingly.

docs/topics/lincheck-model-checking.md
cmlimm: I've rewritten it to use lambda-based API, but found a better way to showcase that the value of the counter is reset. This has increased the line count significantly, but I think this makes the example much more clear. Let me know what you think :)

cmlimm: Discussed during the meeting, this example is okay

docs/topics/lincheck-model-checking.md
cmlimm: Discussed during the meeting, leaving "context switches" to support uniformity

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+* [Bounded](#bounded-exploration) – each test explores only a [limited number of execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
+  Otherwise, the number of possible execution schedules would exponentially grow with the size of the program, which significantly increases testing times.
+
+Compared to stress testing, model checking enables Lincheck to collect execution traces and guarantees bug reproduction 
+for failed tests. For a more detailed comparison, see the table in the [Testing strategies](lincheck-testing-strategies.md)
+article.
+
+## Deterministic exploration
+
+Model checking in Lincheck requires that, given the same execution schedule, the code under test produces the same
```
zamulla:
```suggestion
Model checking in Lincheck requires that, given the same input data and execution schedule, the code under test produces the same 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+  Otherwise, the number of possible execution schedules would exponentially grow with the size of the program, which significantly increases testing times.
+
+Compared to stress testing, model checking enables Lincheck to collect execution traces and guarantees bug reproduction 
+for failed tests. For a more detailed comparison, see the table in the [Testing strategies](lincheck-testing-strategies.md)
+article.
+
+## Deterministic exploration
+
+Model checking in Lincheck requires that, given the same execution schedule, the code under test produces the same 
+results if the input data has not changed. Deterministic execution in model checking is what enables Lincheck to
```
zamulla:
```suggestion
results. Deterministic execution in model checking is what enables Lincheck to 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+      @Test
+      @Order(4)
+      fun valuePersistsAfterRegularIncTest() {
+          // Verify `atomicInt` still holds 1 after `regularIncTest()`
+          check(atomicInt.get() == 1)
+      }
+  }
+  ```
+
+### Uncontrolled sources of non-determinism
```
zamulla, line 110: It feels like the intent behind this section is the same as for the "Known limitations", because uncontrolled non-determinism is the biggest limitation of the model, right? And we even link from each of the limitations down the page for details. Should we merge them?

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+Lincheck does not control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it
```
zamulla:
```suggestion
* **Thread-local variables**. Lincheck does not reset thread-local variables between the invocations of a test as it 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+  ```
+
+### Uncontrolled sources of non-determinism
+
+Lincheck does not control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it 
+  does with global variables. See an example of an error caused by the usage of thread-local variables and a workaround 
+  in a [dedicated section](#thread-local-variables).
+* **Weak references** – Lincheck does not control when a garbage collector removes objects that are only referenced
```
zamulla:
```suggestion
* **Weak references**. Lincheck does not control garbage collector removing objects that are only referenced 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+Lincheck does not control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it 
+  does with global variables. See an example of an error caused by the usage of thread-local variables and a workaround 
+  in a [dedicated section](#thread-local-variables).
+* **Weak references** – Lincheck does not control when a garbage collector removes objects that are only referenced 
+  by weak references. Calling `get()` on such objects produces non-deterministic results. The test might pass successfully 
+  despite the use of weak references, but if Lincheck encounters an [inconsistency between the runs of the same test](#deterministic-exploration),
+  it will raise a non-determinism error.
+* **I/O API calls** – Lincheck does not support calls to I/O APIs, including operations on files and sockets.
```
zamulla:
```suggestion
* **I/O API calls**. Lincheck does not support calls to I/O APIs, including operations on files and sockets. 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+  it will raise a non-determinism error.
+* **I/O API calls** – Lincheck does not support calls to I/O APIs, including operations on files and sockets. 
+  Calling I/O APIs leads to `java.lang.IllegalStateException`. See an example in a [dedicated section](#i-o-api-calls).
+
+## Bounded exploration
+
+The number of possible execution schedules grows exponentially with the size of the program, which increases testing 
+times. Lincheck circumvents this problem by limiting the [number of explored execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
+If the number of possible execution schedules exceeds the specified limit, Lincheck stops the exploration. 
+In such cases, Lincheck tries to evenly analyze logically different schedules:
```
zamulla: I'd make this a start of the next paragraph, to highlight new information. "When the number of execution schedules is limited, Lincheck tries..."

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+  by weak references. Calling `get()` on such objects produces non-deterministic results. The test might pass successfully 
+  despite the use of weak references, but if Lincheck encounters an [inconsistency between the runs of the same test](#deterministic-exploration),
+  it will raise a non-determinism error.
+* **I/O API calls** – Lincheck does not support calls to I/O APIs, including operations on files and sockets. 
+  Calling I/O APIs leads to `java.lang.IllegalStateException`. See an example in a [dedicated section](#i-o-api-calls).
+
+## Bounded exploration
+
+The number of possible execution schedules grows exponentially with the size of the program, which increases testing 
+times. Lincheck circumvents this problem by limiting the [number of explored execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
```
zamulla: The link text does not really prepare for what's on the other side: "execution schedule" is not mentioned there, so I'm not sure why I should have followed that link.

On the other hand, the parameter in the table mentions "invocations" and that reminds me that "invocation" has not been defined yet. Is there a difference between "invocation" and "exploration of an execution schedule"? If not, I vote for "invocation".

zamulla: Also let's not anthropomorphize Lincheck: it doesn't circumvent the problem, it has a useful option with a reasonable default. So as with most tools, let's say that it helps to circumvent the problem.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+If the number of possible execution schedules exceeds the specified limit, Lincheck stops the exploration. 
+In such cases, Lincheck tries to evenly analyze logically different schedules:
+
+* Lincheck applies context bounding by first exploring schedules with one preemptive context switch, then two, and so on.
```
zamulla: I might be out of my depth, but to me it's not clear what a context switch is ("context bounding" kind of continues "limiting the number of schedules", so less questions there).

As a consequence, I can't really read the diagram: are the less-colored windows context switches? Or are the blocks of the opposite color within those windows context switches? Pls ELI5.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+* When choosing the next schedule to explore, Lincheck prioritizes schedules with context switches in new locations.
+
+  For example, see how Lincheck models execution schedules with a single context switch in a two-thread scenario:
+
+  ![A diagram of execution schedules modeled by Lincheck for the same concurrent scenario.](model-checking.svg){width=700}
+
+  Because Lincheck starts by modeling a schedule with a context switch in the first thread, the next modeled schedule 
+  is more likely to have a context switch in the second thread. This continues until Lincheck either reaches the limit 
+  of explored schedules or exhausts all possible schedules.
```
zamulla: When a list item has four paragraphs (and is the bulk of the section), I wonder if it should stay a list item or become a subsection. Maybe the list can keep the "When choosing..." sentence, and the rest put in a section called "Prioritizing context switches" or something.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+## Known limitations and workarounds
+
+The model checking strategy has the following known limitations.
+
+### Relaxed Java memory model
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/370).
+>
+{style="tip"}
```
zamulla, line 161: Maybe let me read about the issue first :) I think after the first paragraph would be fine, to not let it get lost after the code block.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+Model checking requires Lincheck to assume a
+[sequentially consistent memory model](https://en.wikipedia.org/wiki/Sequential_consistency) of the execution.
+This means that Lincheck does not simulate and cannot catch bugs related to instruction reordering, memory cache 
+behavior, and other similar effects under the relaxed [Java memory model](https://en.wikipedia.org/wiki/Java_memory_model).
```
zamulla: A lot of new words which are not easy to intuit the grammar for (Linchek does not simulate under the model?). Can we break it down? "Relaxed Java memory model can introduce bugs related to X, Y, Z. With model checking, Lincheck can not simulate such behavior and catch bugs related to it."

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+#### Workaround
+
+If you want to test concurrent code without the assumption of a sequentially consistent memory model, Lincheck provides 
+a [stress testing strategy](lincheck-testing-strategies.md#stress-testing) for concurrent data structures.
+
+### Threads created outside the scenario (coroutines, `ForkJoinPool`)
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/388).
+>
+{style="tip"}
```
zamulla, line 203: Again, I'd move this link below the first paragraph. And for the rest of these sections.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+Lincheck can only track the threads created inside a concurrent scenario. It can miss bugs occurring in externally 
+created threads, such as when using [coroutines](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html) or
```
zamulla: The link is to a specific page rather than coroutines as a whole, should the link text be more specific, too? "[coroutine dispatchers]"?

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+Java's [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html).
+
+#### Workaround {id="workaround-externally-created-threads"}
+
+Local dispatchers/thread pools guarantee that Lincheck can track the lifecycle and activity of threads in a
```
zamulla:
```suggestion
Local dispatchers and thread pools guarantee that Lincheck can track the lifecycle and activity of threads in a 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+avoiding the accumulation problem.
+
+### Weak references
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/279).
+>
+{style="tip"}
+
+Lincheck does not control when a garbage collector removes objects that are only referenced by weak references.
+Calling `get()` on such objects produces non-deterministic results. The test might pass successfully despite the use
```
zamulla, line 324: I think this is the exact text from the section on "uncontrolled sources", which is another argument to merge it into here.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+of weak references, but if Lincheck encounters an [inconsistency between the runs of the same test](#deterministic-exploration), 
+it will raise a non-determinism error.
+
+### Time API calls
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/390).
+>
+{style="tip"}
+
+Lincheck models calls of `java.lang.System.nanoTime()` and `java.lang.System.currentTimeMillis()` by always returning
```
zamulla: Since "model" and "call" are also verbs, this can get confusing, I'd try to rephrase (after a couple of rereads I understood that it's not multiple models and a typo, it's Lincheck that models something, and that something is calls). Especially since we introduced "model" as a noun in the title of the page.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+### Virtual threads
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/261).
+>
+{style="tip"}
+
+Support for [virtual threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html) has not been 
+verified and is not guaranteed. Consider using traditional threads for scenarios requiring model checking until support
```
zamulla: Is "traditional" the best word here? The linked page contrasts virtual threads with platform threads, is that anything?

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+### I/O API calls
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/1027).
+>
+{style="tip"}
+
+Lincheck does not support calls to I/O APIs, including operations on files and sockets to [prevent inconsistencies 
+between the runs of the same test](#deterministic-exploration).
+
+Calling I/O APIs leads to `java.lang.IllegalStateException`:
```
zamulla, line 353: I don't know if it's intended, but apart from the exception calling I/O produces the following. If it's not a bug we probably don't want this reported?

```
org.jetbrains.lincheck.LincheckAssertionError:
Wow! You've caught a bug in Lincheck.
We kindly ask to provide an issue here: https://github.com/JetBrains/lincheck/issues,
attaching a stack trace printed below and the code that causes the error.
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+created threads, such as when using [coroutines](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html) or
+Java's [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html).
+
+#### Workaround {id="workaround-externally-created-threads"}
+
+Local dispatchers/thread pools guarantee that Lincheck can track the lifecycle and activity of threads in a 
+concurrent scenario.
+
+<tabs>
+ <tab id="coroutines" title="Coroutines">
```
zamulla: The tab name can be clearer: I understand that there are only two options to choose from, but these tabs could provide more value with more descriptive names. "Local coroutine dispatchers" instead of "Coroutines", for example.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+            val future1 = executorService.submit(task)
+            val future2 = executorService.submit(task)
+            future1.get()
+            future2.get()
+        } finally {
+            executorService.shutdown()
+        }
+    }
+}
+</code-block>
```
zamulla, line 249: Is the suggested workaround for `ForkJoinPool` not to use `ForkJoinPool`? I definitely did not expect that in the tab named "ForkJoinPool" :) so I'd name the tab clearer here, too.

Come to think of it the lead up to the tabs can also be clearer: "fixed thread pools can guarantee X, so use them instead of ForkJoinPool to allow Lincheck ..."

docs/topics/lincheck-model-checking.md
zamulla: Also, it makes sense in my head, but I'm not sure if that's actually valid: wouldn't Lincheck start exploring with schedules which have _zero_ context switches? Whatever they are.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+shared-memory access (`read` and `write`) or at synchronization points, such as lock acquisition and release, 
+`park`/`unpark`, `wait`/`notify`, and others. This approach enables Lincheck to controllably explore the execution 
+schedules of a program and find the ones that lead to incorrect results.
+
+When testing concurrent code using model checking, Lincheck makes sure that the exploration of the execution 
+schedules is:
+* [Deterministic](#deterministic-exploration) – each invocation of a model checking test returns the same result if 
+  the input data has not changed.
+* [Bounded](#bounded-exploration) – each test explores only a [limited number of execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
+  Otherwise, the number of possible execution schedules would exponentially grow with the size of the program, which significantly increases testing times.
```
zamulla:
```suggestion
  The number of possible execution schedules grows exponentially with the size of the program and always exploring all of them would significantly increase testing times.
```
If I understand the logic correctly. And maybe the link to configuration options makes sense here: "You can adjust the limit by tweaking X."

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+    // ...
+
+    private fun getLocalCounter() = threadLocalCounters.computeIfAbsent(Thread.currentThread().id) {
+        AtomicInteger(0)
+    }
+}
+```
+
+Because `threadLocalCounters` is an instance field of the test class, Lincheck resets it between invocations, 
+avoiding the accumulation problem.
```
zamulla: Just in case: is "resetting instance fields" something that is listed under "Controlled sources of non-determinism"? I only see resetting global variables, but not sure if they are equivalent.

docs/topics/lincheck-testing-strategies.md
```diff
@@ -34,11 +34,9 @@ This means that Lincheck does not simulate and cannot catch bugs related to inst
 cache behavior, and other similar effects under the relaxed 
 [Java memory model](https://en.wikipedia.org/wiki/Java_memory_model).
 
-<!-- TODO: uncomment after the article is published
 > For more information, see [Model checking](lincheck-model-checking.md).
 >
 {style=”tip”}
```
zamulla:
```suggestion
{style="tip"}
```
That's a weird one, I'd search the docs for other instances.

docs/images/model-checking.svg
zamulla, line 1: This is a nit, but on the other diagram we use `inc()` and `dec()`, and in the examples project it's `foo()` and `buzz()`. Doesn't make a lot of difference, but would be nice to align :)

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+times. Lincheck circumvents this problem by limiting the [number of explored execution schedules](lincheck-testing-strategies-options.md#scenario-generation).
+If the number of possible execution schedules exceeds the specified limit, Lincheck stops the exploration. 
+In such cases, Lincheck tries to evenly analyze logically different schedules:
+
+* Lincheck applies context bounding by first exploring schedules with one preemptive context switch, then two, and so on.
+* When choosing the next schedule to explore, Lincheck prioritizes schedules with context switches in new locations.
+
+  For example, see how Lincheck models execution schedules with a single context switch in a two-thread scenario:
+
+  ![A diagram of execution schedules modeled by Lincheck for the same concurrent scenario.](model-checking.svg){width=700}
```
zamulla:
```suggestion
  ![A diagram of four execution schedules modeled by Lincheck for the same two-thread scenario with a single context switch between threads.](model-checking.svg){width=700}
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/370).
+>
+{style="tip"}
+
+Model checking requires Lincheck to assume a
+[sequentially consistent memory model](https://en.wikipedia.org/wiki/Sequential_consistency) of the execution.
+This means that Lincheck does not simulate and cannot catch bugs related to instruction reordering, memory cache 
+behavior, and other similar effects under the relaxed [Java memory model](https://en.wikipedia.org/wiki/Java_memory_model).
+
+Most concurrency bugs can be found even under this assumption, however, Lincheck can miss some bugs caused by low-level
```
zamulla:
```suggestion
Most concurrency bugs can be found even under this assumption. However, Lincheck can miss some bugs caused by low-level 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+### Weak references
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/279).
+>
+{style="tip"}
+
+Lincheck does not control when a garbage collector removes objects that are only referenced by weak references.
+Calling `get()` on such objects produces non-deterministic results. The test might pass successfully despite the use
+of weak references, but if Lincheck encounters an [inconsistency between the runs of the same test](#deterministic-exploration), 
+it will raise a non-determinism error.
```
zamulla:
```suggestion
it raises a non-determinism error.
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+Lincheck does not control the following sources of non-determinism:
+
+* **Thread-local variables** – Lincheck does not reset thread-local variables between the invocations of a test as it 
+  does with global variables. See an example of an error caused by the usage of thread-local variables and a workaround 
+  in a [dedicated section](#thread-local-variables).
+* **Weak references** – Lincheck does not control when a garbage collector removes objects that are only referenced 
+  by weak references. Calling `get()` on such objects produces non-deterministic results. The test might pass successfully 
+  despite the use of weak references, but if Lincheck encounters an [inconsistency between the runs of the same test](#deterministic-exploration),
+  it will raise a non-determinism error.
```
zamulla:
```suggestion
  it raises a non-determinism error.
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+This approach might not correctly model timeouts, elapsed-time comparisons, rate-limiting, or other intended behavior.
+
+### I/O API calls
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/1027).
+>
+{style="tip"}
+
+Lincheck does not support calls to I/O APIs, including operations on files and sockets to [prevent inconsistencies
```
zamulla:
```suggestion
Lincheck does not support calls to I/O APIs, including operations on files and sockets, to [prevent inconsistencies 
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+|                                                 |   AtomicInteger#1.getAndIncrement(): 2 |
+| Thread#1.join()                                 |                                        |
+| counter.element ➜ AtomicInteger#1               |                                        |
+| AtomicInteger#1.get(): 3                        |                                        |
+| ---------------------------------------------------------------------------------------- |
+```
+
+#### Workaround {id="workaround-thread-local-variables"}
+
+Create thread-local variables manually by storing values in a `ConcurrentHashMap` with thread IDs as key values:
```
zamulla:
```suggestion
Create thread-local variables manually by storing values in a `ConcurrentHashMap` with thread IDs as keys:
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+Java's [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html).
+
+#### Workaround {id="workaround-externally-created-threads"}
+
+Local dispatchers/thread pools guarantee that Lincheck can track the lifecycle and activity of threads in a 
+concurrent scenario.
+
+<tabs>
+ <tab id="coroutines" title="Coroutines">
+     <code-block lang="Kotlin">
```
zamulla:
```suggestion
     <code-block lang="kotlin">
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+                // store buffer and compiler reordering.
+                // Lincheck cannot model this behavior with model checking.
+                error("Unreachable under sequential consistency")
+           }
+        }
+    }
+}
+```
+
+#### Workaround
```
zamulla:
```suggestion
#### Workaround {id="workaround-relaxed-java-memory-model"}
```

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+Support for [virtual threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html) has not been 
+verified and is not guaranteed. Consider using traditional threads for scenarios requiring model checking until support 
+is verified.
+
+## See also
+
+* Learn how to [test any concurrent code](lincheck-testing-arbitrary-code.md) with model checking.
+* Learn how to use model checking to [test a concurrent data structure](lincheck-how-to-test-data-structures.md).
+* Learn how to [configure a testing strategy](lincheck-testing-strategies-options.md) in Lincheck.
+* Learn how [model checking compares to stress testing](lincheck-testing-strategies.md).
```
zamulla: I'd drop "learn how" and try to make it a list of links rather than embed links in sentences which is harder to read.
* Test any concurrent code with model checking
* Use model checking to test a concurrent data structure
* Configure a Lincheck testing strategy
* Compare model checking to stress testing

docs/kl.tree
```diff
@@ -14,5 +14,6 @@
             <toc-element topic="lincheck-progress-guarantees.md" accepts-web-file-names="progress-guarantees.html"/>
             <toc-element topic="lincheck-results-validation.md" accepts-web-file-names="sequential-specification.html"/>
         </toc-element>
-	</snippet>
+        <toc-element topic="lincheck-model-checking.md"/>
```
zamulla, line 17: Should this be last in the group? I'd imagine it works better before "Testing data structures".

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+
+// Using ThreadLocal to create a variable leads to a failed test
+private val localCounter: ThreadLocal<AtomicInteger> = ThreadLocal.withInitial {
+    AtomicInteger(0)
+}
+```
+
+This test fails with an error because the value of the counter accumulates across scenario invocations:
+
+```text
```
zamulla, line 286:
```suggestion
```text
The following interleaving leads to the error:
```

Not important, but I thought for consistency, since it's a part of output.

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,371 @@
...
+### Time API calls
+
+> Vote for the related issue and track its progress on [GitHub](https://github.com/JetBrains/lincheck/issues/390).
+>
+{style="tip"}
+
+Lincheck models calls of `java.lang.System.nanoTime()` and `java.lang.System.currentTimeMillis()` by always returning
+a predefined constant to [prevent inconsistencies between the runs of the same test](#deterministic-exploration).
+
+This approach might not correctly model timeouts, elapsed-time comparisons, rate-limiting, or other intended behavior.
```
zamulla: "other intended behavior" can be be confusing because it's way too broad. "or other logic that depends on elapsed time"?

zamulla (changes requested):
That's a big one :)

None of my comments are critical, but there's some potential confusion with terminology that I'd clean up.

The biggest suggestion is to merge "Uncontrolled sources of non-determinism" into "Known limitations".

docs/topics/lincheck-model-checking.md
cmlimm: I intentionally omitted this part because to me it seems like mentioning this is an error two times (in the paragraph and in the code block) is excessive.
We already state that the test failed with an error and this code block expands on "the value of the counter accumulates across scenario invocations", not showcases how Lincheck reports errors. I don't think omitting the message would lead to any confusion :)

docs/kl.tree
cmlimm: My logic for adding this last is that it's less relevant to most people than "Testing data structures". You don't need to understand how model checking works to use Lincheck. It is only relevant to users who already familiarized themself with everything Lincheck has to offer and are looking into how Lincheck works on the inside.

docs/topics/lincheck-model-checking.md
cmlimm: Hm, I remember seeing this, yes. I will talk with the devs about this message

docs/topics/lincheck-model-checking.md
cmlimm: I've reworded this section as whole to be more specific :)

docs/topics/lincheck-model-checking.md
cmlimm: As I understand it, the word "platform" is used to differentiate "regular" threads from virtual, but in general you would not use the word "platform" when you describe them as they are the default ones.

Now that I write this, this is exactly the place to use the term platform threads to avoid ambiguity and to stick closely to a formal tone of voice :)

docs/topics/lincheck-model-checking.md
cmlimm: Context switch is the same as a thread switch when talking about concurrent programming. Bounding the context = limiting the number of thread switches in each invocation of a test.
Regarding the diagram: take the first column – `foo()` runs, pauses, execution switches to `buzz()`, `buzz()` finishes, `foo()` continues. This executions switch = thread switch = context switch.
In the next column, we run the test one more time, but switch the execution at a different point.
And so on.

Regarding zero context switches: if we don't add context switches, the test runs exactly as it is written – `foo()` is never interrupted by `buzz()` and there is no space for an error to occur. An error can only occur when we interrupt an operation (see the [Overview](https://kotlinlang.org/docs/lincheck-guide.html) article if you're interested in an example). So there Lincheck doesn't need to model this.

I agree, that "context switching" is a confusing term. I've initially decided to use it to relate to "context bounding", but mentioning "context bounding" doesn't really provide any value, especially if we just explain what it does. I've removed any mention of "context" overall, let me know if this reads more clearly now.

docs/topics/lincheck-model-checking.md
cmlimm: I would not say that non-determinism is the biggest limitation of model checking by itself, rather a source of a number of limitations placed on model checking :) Moreover, it only causes 3 (4 counting Time APIs) of the 7 described limitations, and the biggest one is still the relaxed memory model.
But I agree that repeating the content multiple times is ill advised (as per your next comment about weak references). I've removed most of the text from this section and replaced it with links to the corresponding sections of the limitations section.

cmlimm: Why I don't want to merge this section into Limitations completely: this would lead to a somewhat unbalanced Deterministic exploration chapter where we state what sources of non-determinism are controlled, but say nothing about which source are not.

zamulla: I think you got exactly what I meant — removing the section would make the page harder to navigate, but merging the content into "Limitations" + links from over here is what makes it all a bit easier to digest.

docs/topics/lincheck-model-checking.md
```diff
@@ -105,54 +107,57 @@ When running a test with the model checking strategy, Lincheck controls the foll
...
 ## Known limitations and workarounds
 
 The model checking strategy has the following known limitations.
 
 ### Relaxed Java memory model
 
+Model checking requires Lincheck to assume a
+[sequentially consistent memory model](https://en.wikipedia.org/wiki/Sequential_consistency) of the execution.
+
+The relaxed memory model used in Java can introduce bugs related to instruction reordering, memory cache
```
zamulla: Are we okay losing the link to https://en.wikipedia.org/wiki/Java_memory_model here?

docs/topics/lincheck-model-checking.md
```diff
@@ -300,40 +305,41 @@ class ThreadLocalVariableTest {
...
 {style="tip"}
 
-Lincheck models calls of `java.lang.System.nanoTime()` and `java.lang.System.currentTimeMillis()` by always returning
+### Time API calls
+
+Lincheck simulates calls of `java.lang.System.nanoTime()` and `java.lang.System.currentTimeMillis()` by always returning
 a predefined constant to [prevent inconsistencies between the runs of the same test](#deterministic-exploration).
 
-This approach might not correctly model timeouts, elapsed-time comparisons, rate-limiting, or other intended behavior.
+This approach might not correctly model timeouts, elapsed-time comparisons, rate-limiting, or other logic that depends
```
zamulla: If it's not hurting the intent, I'd get rid of "model" here too, like "might not correctly simulate..." or "represent..."

docs/topics/lincheck-model-checking.md
```diff
@@ -105,54 +107,57 @@ When running a test with the model checking strategy, Lincheck controls the foll
 
 ### Uncontrolled sources of non-determinism
 
-Lincheck does not control the following sources of non-determinism:
+Usage of the features related to the uncontrolled sources of non-determinism is either restricted or requires 
+workarounds. Each uncontrolled source of non-determinism is explained in detail in a dedicated section:
```
zamulla: The problem I still have is the term itself: implicitly we define "uncontrolled sources of non-determinism" against "controlled sources of non-determinism" above, but that could be clearer, and "usage of the features" is a bit too abstract.

The intro might be: "Lincheck controls [some sources of non-determinism](#controlled-sources-of-non-determinism), but not all. Using code that is non-deterministic in a way that Lincheck can't handle either prevents you from using Lincheck for this particular code or requires workarounds."

docs/topics/lincheck-model-checking.md
```diff
@@ -105,54 +107,57 @@ When running a test with the model checking strategy, Lincheck controls the foll
...
+### Example: schedules with one thread switch 
 
-  ![A diagram of execution schedules modeled by Lincheck for the same concurrent scenario.](model-checking.svg){width=700}
+See how Lincheck models execution schedules with a single thread switch in a two-thread scenario:
 
-  Because Lincheck starts by modeling a schedule with a context switch in the first thread, the next modeled schedule 
-  is more likely to have a context switch in the second thread. This continues until Lincheck either reaches the limit 
-  of explored schedules or exhausts all possible schedules.
+![A diagram of four execution schedules modeled by Lincheck for the same two-thread scenario with a single 
+context switch between threads.](model-checking.svg){width=700}
```
zamulla: One mention of "context switch" left.

docs/topics/lincheck-model-checking.md
```diff
@@ -105,54 +107,57 @@ When running a test with the model checking strategy, Lincheck controls the foll
...
+of a program is explored in each scenario invocation. Because the number of possible execution schedules grows exponentially 
+with the size of the program, [the number of invocations](lincheck-testing-strategies-options.md#scenario-generation) 
+for a single execution scenario test is limited to reduce the testing times. 
+If the number of the invocations required to explore all execution schedules exceeds the specified limit, Lincheck 
+stops the exploration.
 
-* Lincheck applies context bounding by first exploring schedules with one preemptive context switch, then two, and so on.
-* When choosing the next schedule to explore, Lincheck prioritizes schedules with context switches in new locations.
+When Lincheck is unable to analyze _all_ execution schedules, it tries to evenly analyze logically different ones:
+* Lincheck first explores all schedules with a single preemptive thread switch, then all schedules with two, and so on.
```
zamulla, line 132: I missed this the first time: what is a _preemptive_ switch here and how is it different from any other switch?

docs/topics/lincheck-model-checking.md
```diff
@@ -177,29 +182,29 @@ class RelaxedMemoryModelTest {
...
 
-Lincheck can only track the threads created inside a concurrent scenario. It can miss bugs occurring in externally 
-created threads, such as when using [coroutines](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html) or
-Java's [`ForkJoinPool`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html).
-
 #### Workaround {id="workaround-externally-created-threads"}
 
-Local dispatchers/thread pools guarantee that Lincheck can track the lifecycle and activity of threads in a 
-concurrent scenario.
+Using a fixed thread pool as a local coroutines dispatcher or instead of the common thread pool guarantees that Lincheck
```
zamulla: This may be hard to parse. Might be:

> Use a fixed thread pool:
> * as a local coroutine dispatcher (in the coroutines scenario)
> * instead of the common thread pool (such as `ForkJoinPool`)
> This guarantees that <...>

docs/topics/lincheck-model-checking.md
```diff
@@ -300,40 +305,41 @@ class ThreadLocalVariableTest {
 }
 ```
 
-Because `threadLocalCounters` is an instance field of the test class, Lincheck resets it between invocations, 
-avoiding the accumulation problem.
+Because `threadLocalCounters` is a [global variable](#controlled-sources-of-non-determinism), Lincheck resets it
```
zamulla: Now I wonder what is the actual definition of "global variable" we're working with here, because Kotlin docs don't define this directly, even though it is mentioned several times in different contexts. Ideally I would ground this in Kotlin fundamentals — as is, I have to assume that the examples given are exhaustive examples, and I suspect that they aren't.

docs/topics/lincheck-model-checking.md
```diff
@@ -10,19 +10,21 @@ When testing concurrent code using model checking, Lincheck makes sure that the
 schedules is:
 * [Deterministic](#deterministic-exploration) – each invocation of a model checking test returns the same result if
```
zamulla: If we're liking the period instead of the dash in lists (like in controlled sources on non-determinism), I'd change it here, too.

zamulla (commented):
Nice, thanks! The rest of my comments are a polishing pass, but I do get hung up on definitions of "global variable" and "uncontrolled sources", I think these are worth making crystal clear as long as we're trying to get under the hood properly.

docs/topics/lincheck-model-checking.md
cmlimm: This is a thread switch that happens before the operation finishes its execution. As far as I understand, this is a very common term in concurrency, so I've decided not to explain it.

docs/topics/lincheck-model-checking.md
cmlimm: I've clarified with the devs what was meant by global variables and adjusted the related sentences here and in the Controlled sources of non-determinism section :)

zamulla (approved):
Nice, thanks!

docs/topics/lincheck-model-checking.md
```diff
@@ -0,0 +1,387 @@
...
+* **Identity hash codes**. Lincheck fixes the [identity hash codes](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#identityHashCode-java.lang.Object) 
+  of objects.
+* **Time API calls**. Lincheck intercepts the time API calls and [returns deterministic results](#time-api-calls).
+* **Top-level and `companion object` properties**. Lincheck resets the values of top-level `var` properties
+  and `companion object` properties (the Kotlin equivalents of global variables) between invocations in 
+  model checking tests:
+
+  ```kotlin
+  @TestMethodOrder(MethodOrderer.OrderAnnotation::class)
+  class GlobalVariableResetTest {
```
eupp: Since the term "global variables" is no longer used in the doc, perhaps the test itself should be renamed too.

eupp (approved):
Left one nit.

PR #1040, opened 2026-07-29
Replace docs examples with snippets

cmlimm:


examples/src/test/kotlin/org/lincheck/docs/CounterTestWithIterations.kt
```diff
@@ -0,0 +1,24 @@
+package org.lincheck.docs
+
+import org.jetbrains.lincheck.Lincheck
+import kotlin.concurrent.thread
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class CounterTestWithIterations {
```
eupp: Ths parameter is called invocations, not iterations. "Iterations" is a name of the parameter that defines how many concurrent scenarios (i.e., small test programs) will be generated for a data structure test. Invocations is the number of times each concurrent program is run (for both data-structure and lambda-based syntax).

eupp (changes requested):
I think it makes sense to first wait for these 2 PRs to be merged:
- https://github.com/JetBrains/lincheck/pull/1037
- https://github.com/JetBrains/lincheck/pull/1032

and then fixes all snippets at once.

eupp (approved):

PR #1043, opened 2026-08-12
Fix Event Structure strategy read-modify-write operations handling

viliamDokov:
Now removing porf successors of conflicting rmws. Probably needs more tests


lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -297,9 +297,13 @@ internal class EventStructure(
         // [addBinarySynchronizedEvents] and [addBarrierSynchronizedEvents] methods
         // which always pass in a parent
         check((event.label !is InitializationLabel) implies (event.parent != null)) { "Backtracked event must have a parent: $event" }
+        val causalityFrontier = execution.calculateFrontier(event.causalityClock)
+        // Special case for rmw events. We need to skip if we have pinned eve
```
eupp: This sentence is not finished.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -297,9 +297,13 @@ internal class EventStructure(
...
+        for (event in conflicts) {
+            if (event in pinnedEvents) return
+        }
```
eupp: Could be done a little bit more elegantly:
```kotlin
if (conflicts.any { it in pinnedEvents }) return
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
eupp: Also, it probably worth re-phrasing this a bit. 
The comment tells that this is specific for RMW events only, but the check itself do not involve any is-this-an-rmw-event predicate. So it would be performed for any type of conflicts (e.g., locks).

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -297,9 +297,13 @@ internal class EventStructure(
         // [addBinarySynchronizedEvents] and [addBarrierSynchronizedEvents] methods
         // which always pass in a parent
         check((event.label !is InitializationLabel) implies (event.parent != null)) { "Backtracked event must have a parent: $event" }
+        val causalityFrontier = execution.calculateFrontier(event.causalityClock)
```
eupp: Looks like this can be moved after the conflicts/pinned events check. It probably wouldn't matter in terms of performance (early return due to conflicts/pinned events check would likely be a rare event). But just for code hygiene...

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+            fun two(): Pair<Int, Int> {
+                val r3 = x.getAndIncrement()
+                val r4 = x.getAndIncrement()
+                return r3 to r4
+            }
+        }
+
+        val scenario = scenario {
+            parallel {
+                thread{ actor(TestClass::one)}
```
eupp:
```suggestion
                thread { actor(TestClass::one) }
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+                thread{ actor(TestClass::two)}
```
eupp:
```suggestion
                thread { actor(TestClass::two) }
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+            listOf(1,0,0),
+            listOf(0,1,0),
+            listOf(0,0,1),
+        )
+        litmusTest(assertSame(outcomes)) {
+            val x = AtomicInteger(0)
+            val r = IntArray(3)
+
+            val t1 = thread {
+                r[0] = if(x.compareAndSet(0, 1)) 1 else 0
```
eupp:
```suggestion
                r[0] = if (x.compareAndSet(0, 1)) 1 else 0
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+            }
+            val t2 = thread {
+                r[1] = if(x.compareAndSet(0, 1)) 1 else 0
```
eupp:
```suggestion
                r[1] = if (x.compareAndSet(0, 1)) 1 else 0
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+            }
+            val t3 = thread {
+                r[2] = if(x.compareAndSet(0, 1)) 1 else 0
```
eupp:
```suggestion
                r[2] = if (x.compareAndSet(0, 1)) 1 else 0
```

lincheck/src/jvm/test/org/jetbrains/kotlinx/lincheck_test/strategy/eventstructure/PrimitivesTest.kt
```diff
@@ -1647,6 +1647,167 @@ class PrimitivesTest {
...
+        val outcomes = setOf<List<Int>>(
+            listOf(1,0),
+            listOf(0,1)
+        )
+        litmusTest(assertSame(outcomes)) {
+            val x = AtomicInteger(0)
+            val r = IntArray(2)
+
+            val t1 = thread {
+                r[0] = if(x.weakCompareAndSetPlain(0, 3)) 1 else 0
```
eupp:
```suggestion
                r[0] = if (x.weakCompareAndSetPlain(0, 3)) 1 else 0
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -311,11 +315,16 @@ internal class EventStructure(
             // or are observed by the event, a la GenMC
             cut(conflicts)
             cut { cutEvent ->
-                (
-                    // This is safe because of the check at the beginning of the function
-                    cutEvent.id <= event.parent!!.id  ||
-                    newPinnedEvents.contains(cutEvent)
+                val shouldDelete = (
```
eupp: Let's rename to `shouldCut`

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -311,11 +315,16 @@ internal class EventStructure(
...
+                    // Deleted events are with id > than the parent request event which are not in the causality frontier of
+                    event.parent!!.id < cutEvent.id && // This is safe because of the check at the beginning of the function
```
eupp: IMO, swapping the order makes it a bit easier to understand the intent:
```kotlin
cutEvent.id > event.parent!!.id && 
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -311,11 +315,16 @@ internal class EventStructure(
             // or are observed by the event, a la GenMC
             cut(conflicts)
             cut { cutEvent ->
-                (
-                    // This is safe because of the check at the beginning of the function
-                    cutEvent.id <= event.parent!!.id  ||
-                    newPinnedEvents.contains(cutEvent)
+                val shouldDelete = (
+                    // Deleted events are with id > than the parent request event which are not in the causality frontier of
```
eupp: Let's spell out "greater than" instead of `>` symbol. 
Also it looks like the sentence is not finished.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -311,11 +315,16 @@ internal class EventStructure(
...
+                    event.parent!!.id < cutEvent.id && // This is safe because of the check at the beginning of the function
+                    !causalityFrontier.contains(cutEvent)
                 )
+                // Bail out if one of the events we want to delete is pinned
+                if (shouldDelete && pinnedEvents.contains(cutEvent)) {
+                    return
```
eupp: Can inline `return` on the same line as `if` check to save few lines of code.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -311,11 +315,16 @@ internal class EventStructure(
...
-                (
-                    // This is safe because of the check at the beginning of the function
-                    cutEvent.id <= event.parent!!.id  ||
-                    newPinnedEvents.contains(cutEvent)
+                val shouldDelete = (
+                    // Deleted events are with id > than the parent request event which are not in the causality frontier of
+                    event.parent!!.id < cutEvent.id && // This is safe because of the check at the beginning of the function
+                    !causalityFrontier.contains(cutEvent)
                 )
+                // Bail out if one of the events we want to delete is pinned
```
eupp: Let's re-phrase the comment a bit. Currently, it's not obvious that `return` exits the whole `createBacktrackingPoint` function, thus skipping the bactracking point creation entirely. Let's highlight that.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -395,6 +404,22 @@ internal class EventStructure(
                     }
                 }
             }
+            label is ReadAccessLabel && label.isResponse && label.isExclusive -> run {
+                // TODO: Also handle the case where there are non-exclusive reads messing this up...
+                val write = dependencies.first()
```
eupp: It's a bit more robust to use `find { it. is WriteAccessLabel || ... }` instead of `first()`. 
Otherwise, you implicitly rely on the invariant that write event is always stored as a first element in the dependencies list --- this invariant may brake in the future.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -395,6 +404,22 @@ internal class EventStructure(
...
+                val writeLabel = write.label
+                check(writeLabel is WriteAccessLabel || writeLabel is ObjectAllocationLabel || writeLabel is InitializationLabel)
+                for (event in execution) {
```
eupp: We need to abstract and extract this functionality. 
Essentially, here you want to given a write event `write` obtain a list of read events that read from this `write`. 

Moreover, there is a problem that currently to find all reads reading from the write event, a full execution traversal is performed (`O(n)` complexity).

We can improve on that by maintaining an index `write -> [reads]` that gives for each write event a list of read events that read from it. 

We already have an infrastructure for this. The class `ExtendedExecution` enhances the `Execution` with some additional information, such as auxilarly relations and indices. In particular, there is `AtomicMemoryAccessEventIndex` class that maintains various indices for faster (`O(1)`) events lookup. 

So we can add to `AtomicMemoryAccessEventIndex` the following method:
```kotlin
fun getReadResponses(write: AtomicThreadEvent, location: MemoryLocation): SortedList<AtomicThreadEvent>
```

that for a given `write` event will return a list of reads that read-form it.
(Note that additional `location` argument is needed to disambiguate for the cases when `write` is `ObjectAllocation` or `Initialization` event). 

Inside `AtomicMemoryAccessEventIndex` implementation we can maintain a `Map<Pair<AtomicThreadEvent, MemoryLocation>,List<AtomicThreadEvent> >` . 
Each time a new read-response event is indexed, we will update this map.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
```diff
@@ -395,6 +404,22 @@ internal class EventStructure(
...
+                    val otherLabel = event.label
+                    val otherReadLabel = otherLabel.refine<ReadAccessLabel> { isResponse && isExclusive } ?: continue
+                    if (otherReadLabel.location != label.location) continue
+                    if (event.readsFrom != write) continue
+                    val execlusiveWriteLabel = execution[event.threadId, event.threadPosition + 1]?.label as? WriteAccessLabel ?: continue
```
eupp: We also need a helper extension function over `Execution` that given a read-exclusive event will return its corresponding write-exclusive event (if any).

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExecutionFrontier.kt
```diff
@@ -105,7 +105,7 @@ inline fun <reified E : ThreadEvent> MutableExecutionFrontier<E>.cut(events: Lis
     }
 }
 
-fun <E: ThreadEvent> MutableExecutionFrontier<E>.cut(pred: (ThreadEvent) -> Boolean) {
+inline fun <E: ThreadEvent> MutableExecutionFrontier<E>.cut(pred: (ThreadEvent) -> Boolean) {
```
eupp, line 109: I just realized that `cut(predicate)` has a bit confusing semantics. 
It cuts all events that DO NOT satisfy predicate, but intuitively one would assume it should cut events that DO satisfy the predicate. 

Can we "flip" the semantics? In terms of implementation it's simple, since we just need to insert negation along `lastEvent?.pred(...)` call. 

In any case, could we also please add a KDoc comment to all `cut` methods expaining their semantics. I was re-reading their code while reviewing this PR and realized without comments it's actually a bit hard to understand their semantics.

eupp (changes requested):

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
viliamDokov: To be honest this made the code a bit more annoying as now I have to do an explicit cast for the label of the event. I can make it so that we just return the label but this is not ideal as it is more useful in general to return the entire event. Anyway, tell me what you think

As a sidenote, I think it would be cool if the type of events was changed to be `ThreadEvent<l: EventLabel>`, so we can now encode the type of label that this event posses in the return type of functions, as it very often happens that we have some special utility functions which return events with a specific label, but lose that information in the return type.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExecutionFrontier.kt
viliamDokov: I have some options regarding the options I tried. Old and new indicates the order that I tried them in. You can see the commit history to check the options out.

OLD COMMENT: I went with a rename + comment. I was not suprised with the semantics, but I see how it is confusing with the `!shouldDelete`. Perhaps we should just have different functions with seperate names?

NEW COMMENT: I flipped the predicate inside the `cut` fuction. And flipped the conditions so the code works. It looks cleaner so maybe we keep it  

POSSIBLE VERSION: add an `invert` option to the `pred` function `cut` relies on. This is because we now need to wrap the `pred` with a lambda that negates it and this is not ideal? maybe? 

Tell me what you prefer.

viliamDokov:
Should have most the fixes now. I will run full test suite locally tomorrow.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventStructure.kt
eupp: > I think it would be cool if the type of events was changed to be `ThreadEvent<l: EventLabel>`

Yeah, this is a good idea. Also, as a side-note, I think it would be good to refactor events hierarchy in general. I initially added this complex hierarchy with intermediate classes like `ThreadEvent`, `AtomicThreadEvent`, etc. But, in the end, I think it turned out to be an over-complication. We can leave only single class parameterized by an label type `Event<L : EventLabel>`. 

But this refactoring is out-of-scope of this PR.

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/EventIndexing.kt
```diff
@@ -235,9 +246,23 @@ private class MutableAtomicMemoryAccessEventIndexImpl : MutableAtomicMemoryAcces
...
         }
         updateRaceStatus(label.location, event)
+        updateReadsFromIndex(event)
         index.index(event)
     }
 
+    private fun updateReadsFromIndex(event: AtomicThreadEvent) {
+        // Handle only read access labels
+        val label = event.label as? ReadAccessLabel ?: return
+        if(!label.isResponse) return
```
eupp:
```suggestion
        if (!label.isResponse) return
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExecutionFrontier.kt
```diff
@@ -99,15 +99,17 @@ inline fun <reified E : ThreadEvent> MutableExecutionFrontier<E>.cut(events: Lis
...
         }
     }
 }
 
-fun <E: ThreadEvent> MutableExecutionFrontier<E>.cut(pred: (ThreadEvent) -> Boolean) {
+// Pushes back the mutable frontier by removing events while they satisfy the given predicate
+inline fun <E: ThreadEvent> MutableExecutionFrontier<E>.cut(pred: (ThreadEvent) -> Boolean) {
     threadMap.forEach { (tid, lastEvent) ->
-        val pred = lastEvent?.pred(inclusive = true, pred)
+        //NOTE: we need to invert pred as it removes events UNTIL the current event is satisfied
```
eupp:
```suggestion
        // NOTE: we need to invert pred as it removes events UNTIL the current event is satisfied
```

lincheck/src/jvm/main/org/jetbrains/kotlinx/lincheck/strategy/managed/eventstructure/ExecutionFrontier.kt
eupp: I would prefer to keep "NEW COMMENT" `cut` version with inverted conditions (i.e., `cut` removes the events satisfying the predicate).  In addition, we can add another version - `cutUntilTrue` - that cuts events until the one that satisfies predicate (that is, old behavior).

eupp (approved):

